PR #3, opened 2019-10-25
Rework makefile

bertsky:


bertsky:
I still have to merge your latest changes into the PR.

bertsky:
@stweil you just cherrypicked a single commit from the above. This makes keeping this PR up much harder for me. Please consider merging the whole thing.

bertsky:
BTW, looking at the current problems with `pillow` (where some modules already moved to `>=6.2` whereas others still hold at `==5.4.1`) and `tensorflow` (where some modules will want to use `tensorflow-gpu` which is mutually exclusive with `tensorflow`): I don't think we can ever have piece of mind if we try to stuff everything into a single venv. But it sure helps at least trying!

I also added a `clean` target. I think one way to go would be modifying the generated venv for particular workflows and storing the results under different venvs.

bertsky:
Oh, and one more thing: would you mind adding my [workflow makefilization](https://github.com/bertsky/workflow-configuration)? I can think of a good way to incorporate this...

Makefile
```diff
@@ -4,137 +4,210 @@
...
+all: $(VENV) $(OCRD_MODULES)
 
-OCRD_COR_ASV_ANN := $(BIN)/ocrd-cor-asv-ann-evaluate
-OCRD_COR_ASV_ANN += $(BIN)/ocrd-cor-asv-ann-process
-OCRD_EXECUTABLES += $(OCRD_COR_ASV_ANN)
+clean:
+	$(RM) -r $(VENV)
 
-OCRD_EXECUTABLES += $(BIN)/ocrd-dinglehopper
+# update subrepos to the commited revisions:
```
stweil: s/commited/committed/

Makefile
```diff
@@ -4,137 +4,210 @@
...
-OCRD_CIS += $(BIN)/tqdm
-OCRD_EXECUTABLES += $(OCRD_CIS)
+.PHONY: all clean always-update
+all: $(VENV) $(OCRD_MODULES)
 
-OCRD_COR_ASV_ANN := $(BIN)/ocrd-cor-asv-ann-evaluate
-OCRD_COR_ASV_ANN += $(BIN)/ocrd-cor-asv-ann-process
-OCRD_EXECUTABLES += $(OCRD_COR_ASV_ANN)
+clean:
+	$(RM) -r $(VENV)
```
stweil: I just recently learned that recursive removing files with a variable path can be extremly dangerous: A user of Tesseract for Windows removed all installed programs ([now fixed](https://github.com/UB-Mannheim/tesseract/commit/5ffdd78e09776c9e43bda7ca610dd6a257c39769)).

It is easy for the user to run `rm -r venv` instead of running `make clean`. Therefore I'd ommit that rule.

Makefile
```diff
@@ -4,137 +4,210 @@
...
 
-OCRD_EXECUTABLES += $(BIN)/ocrd-dinglehopper
+# update subrepos to the commited revisions:
+# - depends on phony always-update,
```
stweil: Then add $(OCRD_MODULES) as `.PHONY` targets, and `always-update` is not needed and can be removed.

Makefile
```diff
@@ -4,137 +4,210 @@
...
-OCRD_CIS += $(BIN)/ocrd-cis-ocropy-recognize
-OCRD_CIS += $(BIN)/ocrd-cis-ocropy-resegment
-OCRD_CIS += $(BIN)/ocrd-cis-ocropy-segment
-OCRD_CIS += $(BIN)/ocrd-cis-ocropy-train
-OCRD_CIS += $(BIN)/ocrd-cis-profile
-OCRD_CIS += $(BIN)/ocrd-cis-stats
-OCRD_CIS += $(BIN)/tqdm
-OCRD_EXECUTABLES += $(OCRD_CIS)
+.PHONY: all clean always-update
+all: $(VENV) $(OCRD_MODULES)
```
stweil: Your new code for `OCRD_MODULES` is nice. But the default target `all` should not get all submodules because that requires much time and disk space. All submodules are only needed for special use cases, so either we could add a special phony target for that or simply run `git submodule update --init`.

Makefile
bertsky: No, that's a different thing. `$(OCRD_MODULES)` are not phony themselves: they are in the filesystem, they have timestamps, they need not always cause updates themselves. It's just the attempt to update them which is unconditiional!

Makefile
bertsky: I don't see why that could be problematic at all. The `$(VENV)` is under make control. The user does not necessarily know about it. It's the best way to encapsulate, because `make clean` can always be expected to undo everything `make` does.

Makefile
bertsky: If you think so, then yes, one could sub-differentiate between `all` and other targets (like only `modules`), and not make `all` the default. But having the user run `git submodule update --init` is dangerous: it will unconditionally change the module's file tree (without also setting the module directory's timestamp).

Makefile
```diff
@@ -4,137 +4,210 @@
...
+# Build by entering subdir (first dependent), then
+# install ignoring the existing version (to ensure
+# the binary updates):
+$(filter-out $(CUSTOM_INSTALL),$(OCRD_EXECUTABLES)) install-clstm install-tesserocr:
+	. $(ACTIVATE_VENV) && cd $< && pip install -I .
 
-tesserocr/setup.py:
-	git submodule update --init tesserocr
+# At last, add venv dependency (must not become first):
+$(OCRD_EXECUTABLES) install-clstm install-tesserocr $(BIN)/wheel: $(ACTIVATE_VENV)
```
stweil: This line is not needed because `$(OCRD_EXECUTABLES) install-clstm install-tesserocr` depends on `$(BIN)/wheel` (see next line) which already depends on `$(ACTIVATE_VENV)` (see line 67).

Makefile
```diff
@@ -11,6 +11,13 @@ ACTIVATE_VENV := $(VENV)/bin/activate
 
 export PKG_CONFIG_PATH := $(VENV)/lib/pkgconfig
 
+# Use curl for getting a URL, fall back to wget.
+ifeq ($(shell test -x /usr/bin/curl && echo true),)
```
bertsky: Why not just `ifeq ($(shell which curl),)`?

bertsky:
> The old code did not work on MacOS.

Could you please post the output of the failure?

Makefile
```diff
@@ -23,8 +23,12 @@ CUSTOM_INSTALL := $(BIN)/ocrd # add more non-pip installation targets below
...
 OCRD_MODULES := $(shell git submodule status | while read commit dir ref; do echo $$dir; done)
 
-.PHONY: all clean always-update
-all: $(VENV) $(OCRD_MODULES)
+.PHONY: all clean always-update modules
+
+all: $(VENV)
+
+# Get all submodules.
+modules: $(OCRD_MODULES)
```
bertsky: That's not going to work, because `$(OCRD_EXECUTABLES)` still depends on all the modules.

My idea was to offer `all` as it was (unchanged), but not make it default anymore. The new `.DEFAULT_GOAL` could be `ocrd` for example.

Makefile
```diff
@@ -81,7 +81,8 @@ $(BIN)/wheel: $(ACTIVATE_VENV)
 # Install Python modules from local code.
 
 .PHONY: install-clstm
-install-clstm: clstm install-numpy
+install-clstm: clstm
+install-clstm: install-numpy
```
bertsky: That's odd. The problem here was that the `$(ACTIVATE_VENV)` dependency got interpreted as the _first prerequisite_, causing the recipe based on `$<` to fail. But it escapes me why a prerequisite added at the very bottom of the file would receive that status, and why splitting a prerequisite list way up above would change that. Could you please state the version of your GNU make to ensure we really solved that one?

Makefile
bertsky: You're right!

bertsky: But on the other hand: If anyone later-on feels they should remove `wheel`, this essential dependency would be lost. It does not hurt to have it explicit here.

Makefile
stweil: That's fine and shorter, too.

stweil:
> Why is that more portable? The find syntax used in `printf` is perfectly POSIX.

Maybe, but BSD `find` which is provided on MacOS does not support it. See the [manual](https://www.freebsd.org/cgi/man.cgi?find(1)).

Makefile
stweil: Apple uses a rather old version for MacOS: GNU Make 3.81.

Makefile
stweil: Wouldn't users expect that `make all` is the default? I would because that is quite common.

stweil:
> > The old code did not work on MacOS.
> 
> Could you please post the output of the failure?

The old code used `find -printf` which is unsupported by BSD `find`.

stweil:
> On the contrary: I specifically avoided `xargs` because submodules **very likely** will have too many files to fit on one command line (due to the usual Unix kernel limit). This simply does not work. Besides, it is also slower: the files have to be `stat`ed twice to sort them.

The current submodules don't have too many files to cause a too long command line.
Yes, the new code is slower, but I don't think that it is time critical. It takes less than 4 seconds to touch all submodules.

Makefile
stweil: Only those executables which are built locally depend on a subset of all modules.

bertsky: > Wouldn't users expect that `make all` is the default? I would because that is quite common.

But as you said: this is a huge metarepo with _lots_ of stuff in `all`. And as outlined above, `all` might not at all times be guaranteed to converge due to conflicting requirements. And it goes much more against expectancies to have `all` be less then _all executables_. Anyway, this is just convention and can be documented.

Makefile
bertsky: I see. Well, make never seizes to amaze me. (And I think the _order of prerequisites_ topic is a particularly messy one across versions.) We're lucky the solution is so simple.

bertsky:
> > Why is that more portable? The find syntax used in `printf` is perfectly POSIX.
> 
> Maybe, but BSD `find` which is provided on MacOS does not support it. See the [manual](https://www.freebsd.org/cgi/man.cgi?find(1)).
> > > The old code did not work on MacOS.
> > 
> > 
> > Could you please post the output of the failure?
> 
> The old code used `find -printf` which is unsupported by BSD `find`.

I see. This is really bad news. 

> > On the contrary: I specifically avoided `xargs` because submodules **very likely** will have too many files to fit on one command line (due to the usual Unix kernel limit). This simply does not work. Besides, it is also slower: the files have to be `stat`ed twice to sort them.
> 
> The current submodules don't have too many files to cause a too long command line.

You bet! On my Ubuntu system, the limit (`getconf ARG_MAX`) is 8388608 bytes, but it can be as low as 131072 bytes. We already have `ocrd_olena` with filename entries totaling 1893996 bytes.

> Yes, the new code is slower, but I don't think that it is time critical. It takes less than 4 seconds to touch all submodules

Granted. But we need an approach that does not just fail any day anywhere.

I will come up with something based on bash arrays. MacOS/BSD _does_ have bash?

Makefile
bertsky: I have an even better version which does not need `cd` before and removes the targets if wget/curl failed to download fully.

bertsky:
Maybe `touch -r $$(find $@ -type f -newer $@)` would already suffice? We only need to ensure that the update will trigger (and it's unlikely files in the module directory will have been updated at multiple different times all newer than the module directory itself, given that we makefile-manage them like that).

stweil:
> MacOS/BSD does have bash?

The latest MacOS update switched from bash (5.0.11) to zsh, but bash is still there.

Makefile
stweil: It would help if you could send that as a separate pull request. Usually it is much easier for me to handle many small pull requests than a single large one with lots of commits.

stweil:
> Maybe touch -r $$(find $@ -type f -newer $@) would already suffice?

That's a good idea. But it must be extended a little bit to find at least one file or directory.

bertsky:
> > MacOS/BSD does have bash?
> 
> The latest MacOS update switched from bash (5.0.11) to zsh, but bash is still there.

That reminds me that we forgot to set `SHELL = /bin/bash` since we are already using `$(...)` for command substitution. Or rather keep POSIX and use backticks?

bertsky:
> > Maybe touch -r $$(find $@ -type f -newer $@) would already suffice?
> 
> That's a good idea. But it must be extended a little bit to find at least one file or directory.

True. Exactly one.

Makefile
bertsky: Well the problem is that you had
- already started cherrypicking from the PR
- already started pushing commits here (some of which were incorrect)
- already requested further changes
So I have already lots of (now incompatible) improvements stacked up. I don't think this is how we are supposed to be using PRs. You gather stuff that belongs together, then it gets discussed and touched up until it works. Only then does it make sense to merge. And for everything that comes up in the discussion, the OP should get the first chance at doing the touching up.

I'd rather open a new PR altogether with all my incremental changes since yesterday. Or force-push here, if you prefer that. (I made sure to include every feature you asked for.)

But to separate this out of a still emerging PR does not make sense to me.

Makefile
bertsky: (Self-documentation is among the many things I have here waiting for a push.)

stweil:
Feel free to rebase this PR on my latest Git master, remove or squash commits as needed and force-push it again.

stweil:
> Or rather keep POSIX and use backticks?

No. Shell code checkers suggest to replace all backtick shell code.

bertsky:
> Feel free to rebase this PR on my latest Git master, remove or squash commits as needed and force-push it again.

I decided to force-push here (to have a linear PR) but then merge. IMO this PR is ready for adoption.

Makefile
```diff
@@ -2,153 +2,289 @@
...
 # Special rule for equ.traineddata which is only available from tesseract-ocr/tessdata.
 $(TESSDATA)/equ.traineddata:
 	@mkdir -p $(dir $@)
-	cd $(dir $@) && wget https://github.com/tesseract-ocr/tessdata/raw/master/$(notdir $@)
+	$(WGET) $@ https://github.com/tesseract-ocr/tessdata/raw/master/$(notdir $@) || \
+		{ $(RM) $@; false; }
 
 # Default rule for all other traineddata models.
-%.traineddata:
+$(TESSDATA)/%.traineddata:
```
stweil: What was the reason for this modification? With the old code it was possible to install a model anywhere by just saying `make PATH/MODEL.traineddata`.

Makefile
```diff
@@ -162,4 +298,13 @@ $(BIN)/tesseract: tesseract/configure
 	cd tesseract/build && ../configure --disable-openmp --disable-shared --prefix="$(VENV)" CXXFLAGS="-g -O2 -fPIC"
 	cd tesseract/build && make install
 
+# do not delete intermediate targets:
+.SECONDARY:
```
stweil: Are there such targets which should not be deleted?

README.md
```diff
@@ -28,14 +30,47 @@ Otherwise or for the latest Tesseract code it can also be built locally.
...
+
+### Variables
+
+<dl>
+  <dt>PYTHON</dt>
+  <dd>name of the Python binary to use (at least python3 required)</dd>
+  <dt>VENV</dt>
+  <dd>Directory to use for venv</dd>
+  <dt>VIRTUAL_ENV</dt>
+  <dd>Override for VENV (so an already activated venv gets re-used).</dd>
```
stweil: Your modified Makefile automatically uses an already activated `VIRTUAL_ENV`. To make this even cleared, I suggest to drop `VENV` in favour of `VIRTUAL_ENV`.

README.md
```diff
@@ -47,27 +82,40 @@ To build the latest Tesseract locally, run this command first:
...
 Optionally install additional Tesseract models.
 
-    # Make sure to use the correct tessdata directory.
-    make venv/share/tessdata/script/Latin.traineddata
-    make venv/share/tessdata/script/Fraktur.traineddata
+    # Download models from tessdata_fast into the venv's tessdata directory.
+    make frk.traineddata
+    make script/Latin.traineddata
+    make script/Fraktur.traineddata
+
```
stweil: See my comment above. If users have installed Tesseract from their Linux distribution (or `brew` on MacOS), the `tessdata` directory is not part of the virtual Python environment.

README.md
```diff
@@ -47,27 +82,40 @@ To build the latest Tesseract locally, run this command first:
...
+- ocrd_tesserocr (too old in PyPI; needs `ocrd>=2.0.0` / core branch `edge`)
+- tesserocr (too old in PyPI)
+
+### Conflicting requirements
+
+Merging all packages into one venv does not always work.
+Modules may require mutually exclusive sets of dependent packages.
+
+- `Pillow` (`==5.4.1` vs `>=6.2.0`)
+- `tensorflow` (`-gpu` vs CPU, `<2.0` vs `>=2.0`)
```
stweil: Idea: would it be possible to install the real `tensorflow` and a dummy `tensorflow-gpu` in parallel (or vice versa if GPU support is available)?

Makefile
bertsky: > What was the reason for this modification?

To have the user install models without knowing the details of venv and tessdata path – just the way they see it in the API/CLI.

> With the old code it was possible to install a model anywhere by just saying `make PATH/MODEL.traineddata`

That's still possible:
```bash
sudo make TESSDATA=/usr/local/share/tessdata frk.traineddata
```

Makefile
bertsky: Yes, all the abbreviated forms, e.g. `frk.traineddata` (not removing `venv/share/tessdata/frk.traineddata`) or `ocrd-tesserocr-recognize` (not removing `venv/bin/ocrd-tesserocr-recognize`).

README.md
bertsky: Yes, thinking it over, it does not make sense anymore to discern the two.

README.md
bertsky: Yes, but in that case, users had to know to put their tessdata path in front, and use `sudo` anyway. And there's no harm in using `venv/share/tessdata` by default.

I can separate this whole thing out of the PR if you don't like it.

README.md
bertsky: You _can_ have both `tensorflow` and `tensorflow_gpu` in parallel (even on a different version). But some modules depend on either the one or the other being installed to decide which models to load (roughly, `CuDNNLSTM` vs `LSTM` layers). But I would argue that this strategy is not reliable either (even in a single venv) because `tensorflow` can be built with GPU support and libcudnn as well. The better way is to check GPU/cudnn availability from within Python with Tensorflow code, and then configure the model accordingly – as I have done in `ocrd_keraslm` and `cor-asv-ann`.

stweil: See https://github.com/stweil/tensorflow_gpu_to_tensorflow for a possible solution.

bertsky: > See https://github.com/stweil/tensorflow_gpu_to_tensorflow for a possible solution.

That's not going to work if what the package really requires is CuDNN support, since there's no guarantee the available `tensorflow` has it. (It's usually only the case via custom GPU-enabled bazel builds of Tensorflow.)

stweil: The dummy package helps on hosts for which there is no `tensorflow-gpu`. And it allows at least running

    venv/bin/ocrd-anybaseocr-binarize --help

bertsky: Yes, this is no doubt useful under certain circumstances. But not for the problem just discussed: if the program really needs CuDNN then a dummy would only shift the error towards runtime. And on the other hand, if the `tensorflow` wheel already includes it, then the dummy is not needed anyway.

In my modules (`cor-asv-ann` and `ocrd_keraslm`) I have made sure to not rely on CuDNN being available, but still make use of it if it is (and that both implementations can share their weights). That extra effort is not needed after Tensorflow 2.0 (see below), but up until recenlty, this was the only way IMHO. Otherwise you would have to either refrain from relying on CuDNN implementations, or make them a requirement during installation (by depending on the `tensorflow-gpu` package, among other things). But even this can get messy.

For example, I ran into [problems](https://github.com/OCR-D/ocrd_calamari/issues/6) because my modules just require `tensorflow` but ocrd_calamari requries `tensorflow-gpu`. You can install both in parallel, but apparently then your venv is unusable. Equally, [segmentation-runner](https://github.com/ocr-d-modul-2-segmentierung/segmentation-runner) uses a `page-segmentation` subrepo which they install together with either `tensorflow` or `tensorflow-gpu` by defining respective setuptools extras `tf_cpu` and `tf_gpu`, depending on a setting in their Makefile (which they just now simplified by introducing an environment variable `TENSORFLOW_GPU`). But what if another package uses other rules?

Then again, that story very recently changed completely: since Tensorflow 1.15, there is no separate `tensorflow-gpu` package anymore, and since Tensorflow 2.0, even the RNN model APIs have been unified. So modules really should not concern themselves with these differences and leave the question of GPU support entirely to the runtime wisdom of Tensorflow and to the non-Python part of the installer. And since it is quite difficult to properly install or even detect whether CUDA toolkit, libCuDNN etc are installed, or even available on your platform, it's probably best to use the Tensorflow docker distribution.

To sum up: CuDNN/GPU support in Tensorflow should very soon become a non-issue, and we should encourage modules to update to Tensorflow 2.0. But all that does not change the underlying problem for the unified venv approach: you can have conflicting version requirements (eg. Tensorflow 1.15 vs Tensorflow 2.0, Pillow, etc).

mikegerber: FYI ocrd_calamari depends on `tensorflow-gpu` and this falls back to CPU if no GPU is available, so I do not see an issue with just installing `tensorflow-gpu`. (Docker is more difficult, because you need to use a nvidia/cuda base image.)

mikegerber: @bertsky 
> For example, I ran into problems because my modules just require tensorflow but ocrd_calamari requries tensorflow-gpu. You can install both in parallel, but apparently then your venv is unusable

I have a venv here with both installed, and no issues, I don't think that was your problem. I almost exclusively use fresh venvs or containers, I might not run into issues with downgrading or conflicting versions etc.

README.md
bertsky: done

README.md
bertsky: @mikegerber You'll notice the issue if you try `make all` here and run Calamari. This whole discussion is whether this approach can (ever) work. I'd say let's merge it as it is, but try to come up with better encapsulation as the next step. That'll likely be separate venvs for different modules, and re-using docker images where available. But that's just my opinion.

bertsky: > That'll likely be separate venvs for different modules, and re-using docker images where available.

@stweil, I will make a separate PR along those lines. Briefly, we will have:
- a custom venv directory for each module (perhaps named `venv.MODULE`) that does not offer a docker image
- a single `scripts/` directory where we put executable scripts for all modules generated by the makefile; each script points to either the module's venv or uses its docker image

Thus, it suffices to add `scripts/` into the `PATH` and you can run everything you want.

README.md
bertsky: Resolve or separate this change?

README.md
stweil: I'd prefer a differenz solution because one venv per module wastes build time and disk space. As far as I see most modules can be put in a single venv. Then only one or a small number of extra venv directories are needed for the special cases. Hook scripts in the main venv/bin could forward to the others.

Makefile
```diff
@@ -2,153 +2,289 @@
...
-# Install Python modules from local code.
+$(OCRD_SEGMENTATION_RUNNER): segmentation-runner
 
-clstm/setup.py:
-	git submodule update --init clstm
+OCRD_EXECUTABLES += $(OCRD_ANYBASEOCR)
 
-clstm: $(ACTIVATE_VENV) clstm/setup.py numpy $(BIN)/wheel
-	. $(ACTIVATE_VENV) && cd $@ && pip install .
+OCRD_ANYBASEOCR := ocrd-anybaseocr-crop
```
stweil: This pattern (without leading `$(BIN)/`) does not work currently, although I'd prefer it.

Makefile
```diff
@@ -2,153 +2,289 @@
...
+#   so this will not only work on first checkout
+# - updates the module to the revision registered here
+# - then changes the time stamp of the directory to that
+#   of the latest modified file contained in it,
+#   so the directory can be used as a dependency
+modules: $(OCRD_MODULES)
+$(OCRD_MODULES): always-update
+	git submodule status $@ | grep -q '^ ' || { \
+		git submodule update --init $@ && \
+		touch -r $$(find $@ -type f -newer $@ | tail -1 || echo $@) $@; }
```
stweil: That failed for submodule directories which were newer than their content. Fixed now in master.

Makefile
bertsky: I don't see this. 

```bash
make ocrd-anybaseocr-crop
git submodule status LAYoutERkennung | grep -q '^ ' || { \
	git submodule update --init LAYoutERkennung && \
	touch -r $(find LAYoutERkennung -type f -newer LAYoutERkennung | tail -1 || echo LAYoutERkennung) LAYoutERkennung; }
. venv/bin/activate && cd LAYoutERkennung && pip install  .
Processing LAYoutERkennung
  Requirement already satisfied (use --upgrade to upgrade): ocrd-anybaseocr==0.0.1 from LAYoutERkennung in venv/lib/python3.6/site-packages
Requirement already satisfied: numpy>=1.15.4 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: opencv-python>=3.4 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd-fork-ocropy>=1.4.0a3 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd-fork-pylsd>=0.0.3 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: click in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd>=1.0.0b15 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: torch>=1.1.0 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: torchvision in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: dominate>=2.3.5 in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: pandas in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: keras in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Collecting tensorflow-gpu==1.14 (from ocrd-anybaseocr==0.0.1)
  Cache entry deserialization failed, entry ignored
  Using cached https://files.pythonhosted.org/packages/76/04/43153bfdfcf6c9a4c38ecdb971ca9a75b9a791bb69a764d652c359aca504/tensorflow_gpu-1.14.0-cp36-cp36m-manylinux1_x86_64.whl
Requirement already satisfied: scikit-image in venv/lib/python3.6/site-packages (from ocrd-anybaseocr==0.0.1)
Requirement already satisfied: scipy>=1.1.0 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: imageio>=2.4.1 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: matplotlib>=2.2.3 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: six>=1.11.0 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: Pillow>=2.7.0 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: lxml>=3.5.0 in venv/lib/python3.6/site-packages (from ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: bagit>=1.7.0 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: bagit_profile>=1.3.0 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: requests in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: opencv-python-headless in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: Flask in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: jsonschema in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: pyyaml in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: atomicwrites>=1.3.0 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: Deprecated==1.2.0 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd_utils==2.0.0a1 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd_models==2.0.0a1 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd_modelfactory==2.0.0a1 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: ocrd_validators==2.0.0a1 in venv/lib/python3.6/site-packages (from ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: pytz>=2017.2 in venv/lib/python3.6/site-packages (from pandas->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: python-dateutil>=2.6.1 in venv/lib/python3.6/site-packages (from pandas->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: keras-preprocessing>=1.0.5 in venv/lib/python3.6/site-packages (from keras->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: h5py in venv/lib/python3.6/site-packages (from keras->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: keras-applications>=1.0.6 in venv/lib/python3.6/site-packages (from keras->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: wrapt>=1.11.1 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: wheel>=0.26 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: protobuf>=3.6.1 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: gast>=0.2.0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: absl-py>=0.7.0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: tensorboard<1.15.0,>=1.14.0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: tensorflow-estimator<1.15.0rc0,>=1.14.0rc0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: grpcio>=1.8.6 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: astor>=0.6.0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: google-pasta>=0.1.6 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: termcolor>=1.1.0 in venv/lib/python3.6/site-packages (from tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: networkx>=2.0 in venv/lib/python3.6/site-packages (from scikit-image->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: PyWavelets>=0.4.0 in venv/lib/python3.6/site-packages (from scikit-image->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: cycler>=0.10 in venv/lib/python3.6/site-packages (from matplotlib>=2.2.3->ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: kiwisolver>=1.0.1 in venv/lib/python3.6/site-packages (from matplotlib>=2.2.3->ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: pyparsing!=2.0.4,!=2.1.2,!=2.1.6,>=2.0.1 in venv/lib/python3.6/site-packages (from matplotlib>=2.2.3->ocrd-fork-ocropy>=1.4.0a3->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: certifi>=2017.4.17 in venv/lib/python3.6/site-packages (from requests->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in venv/lib/python3.6/site-packages (from requests->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: idna<2.9,>=2.5 in venv/lib/python3.6/site-packages (from requests->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1 in venv/lib/python3.6/site-packages (from requests->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: Jinja2>=2.10.1 in venv/lib/python3.6/site-packages (from Flask->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: itsdangerous>=0.24 in venv/lib/python3.6/site-packages (from Flask->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: Werkzeug>=0.15 in venv/lib/python3.6/site-packages (from Flask->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: importlib-metadata in venv/lib/python3.6/site-packages (from jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: setuptools in venv/lib/python3.6/site-packages (from jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: pyrsistent>=0.14.0 in venv/lib/python3.6/site-packages (from jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: attrs>=17.4.0 in venv/lib/python3.6/site-packages (from jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: markdown>=2.6.8 in venv/lib/python3.6/site-packages (from tensorboard<1.15.0,>=1.14.0->tensorflow-gpu==1.14->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: decorator>=4.3.0 in venv/lib/python3.6/site-packages (from networkx>=2.0->scikit-image->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: MarkupSafe>=0.23 in venv/lib/python3.6/site-packages (from Jinja2>=2.10.1->Flask->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: zipp>=0.5 in venv/lib/python3.6/site-packages (from importlib-metadata->jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Requirement already satisfied: more-itertools in venv/lib/python3.6/site-packages (from zipp>=0.5->importlib-metadata->jsonschema->ocrd>=1.0.0b15->ocrd-anybaseocr==0.0.1)
Installing collected packages: tensorflow-gpu
Successfully installed tensorflow-gpu-1.14.0
. venv/bin/activate && cd LAYoutERkennung && pip install --no-deps --force-reinstall  .
Processing LAYoutERkennung
  Requirement already satisfied (use --upgrade to upgrade): ocrd-anybaseocr==0.0.1 from venv/lib/python3.6/site-packages
(venv) $ ocrd-anybaseocr-crop --help
Using TensorFlow backend.
Usage: ocrd-anybaseocr-crop [OPTIONS]

Options:
  -V, --version                   Show version
  -l, --log-level [OFF|ERROR|WARN|INFO|DEBUG|TRACE]
                                  Log level
  -J, --dump-json                 Dump tool description as JSON and exit
  -p, --parameter TEXT            Parameters, either JSON string or path to
                                  JSON file
  -g, --page-id TEXT              ID(s) of the pages to process
  -O, --output-file-grp TEXT      File group(s) used as output.
  -I, --input-file-grp TEXT       File group(s) used as input.
  -w, --working-dir TEXT          Working Directory
  -m, --mets TEXT                 METS URL to validate
  --help                          Show this message and exit.
```

What exactly does not work for you currently?

Makefile
bertsky: Oh, right – I hadn't thought of that.

README.md
bertsky: > I'd prefer a differenz solution because one venv per module wastes build time and disk space. As far as I see most modules can be put in a single venv. Then only one or a small number of extra venv directories are needed for the special cases. Hook scripts in the main venv/bin could forward to the others.

But you'd have to manually find out which modules need to be separated at any given time, because we don't have a means to find out conflicting requirements automatically. The separate venvs would be a fallback solution for modules that don't provide docker images. The extra time (spent once during setup) for installing separate venvs and the extra disk space are a small price to pay for automated consistency IMO. (And that price can be negotiated down by dockerizing modules.)

Makefile
stweil: `make all` fails.

bertsky: After the merge (and your clstm update), I can see this:
```
git submodule status clstm | grep -q '^ ' || { \
	git submodule update --init clstm && \
	touch -r $(find clstm -type f -newer clstm -o -type d -name clstm | tail -1) clstm; }
error: Server does not allow request for unadvertised object e479b7a87b457be44dbcde9e92c5d06f2dfd5df6
Fetched in submodule path 'clstm', but it did not contain e479b7a87b457be44dbcde9e92c5d06f2dfd5df6. Direct fetching of that commit failed.
```

Maybe you forgot to push your last commits to `clstm`?

stweil: No, but your local clstm might still use stweil instead of tmbdev as remote repository.

bertsky: > No, but your local clstm might still use stweil instead of tmbdev as remote repository.

It does!

I am surprised `git submodule update` does not change that. Is this a bug or a feature? (In the latter case, we need a workaround here.)

stweil: The problem was only for `clstm` because my fork had an update for Python3 which was missing in the official repository. I don't think a workaround is needed, as only a few others like @kba and @wrznr might be affected.

bertsky: My concern is with the general problem of submodules changing their upstream and clones/installations not being able to update automatically. What we need is provided by `git submodule sync`. 

I also wonder whether we should make `git submodule update --recursive`, because we already have modules which have submodules (e.g. `segmentation-runner/page-segmentation`).

PR #4, opened 2019-11-01
update submodules

kba:


kba:
Updated clstm and ocrd_cis to point to upstream master

stweil:
I used `dev` instead of `master` for `ocrd_cis`...

kba:
> I used dev instead of master for ocrd_cis...

My bad, fixed.

PR #6, opened 2019-11-18
update, add workflow-configuration, deps-ubuntu and docker

bertsky:
@stweil, this is my attempt at a "fat" docker setup. Having difficulties getting versioning right for my [slim docker proposal](https://github.com/stweil/ocrd_all/pull/3#discussion_r341503872), I think this is at least a good stopgap. The result can be obtained under dockerhub as `bertsky/ocrd_all` for now (but I hope someone will publish it under `ocrd/all` soon).

@mikegerber @kba, I also included an updated version of bertsky/workflow-configuration. (For those afraid of calling `make` I included a system-wide installation as `ocrd-make`, simply wrapping the make calls.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ ENV PIP_INSTALL="pip install --timeout=3000 -q"
 # start a shell script (so we can comment individual steps here)
 RUN echo "set -ex" > docker.sh
 # get packages for build
-RUN echo "apt-get -y install build-essential automake autoconf libtool pkg-config git make" >> docker.sh
+RUN echo "apt-get -y install automake autoconf libtool pkg-config g++ git make" >> docker.sh
 # create git repo just so the (unconditional) submodule update recipes don't fail
 RUN echo "git init" >> docker.sh
 # we want to use PREFIX as venv
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,10 @@ RUN echo "rm -fr /build" >> docker.sh
 # run the script in one layer/step (to minimise image size)
 RUN bash docker.sh
 
+# remove (dated) security workaround preventing use of
+# ImageMagick's convert on PDF/PS/EPS/XPS:
+RUN rm /etc/ImageMagick-6/policy.xml
+
 ENV DEBIAN_FRONTEND teletype
 
 WORKDIR /data
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ ENV BUILDDEPS="build-essential automake autoconf libtool pkg-config git"
 ENV PIP_INSTALL="pip install --timeout=300 -q"
 
 # start a shell script (so we can comment individual steps here)
-RUN echo "set -x" > docker.sh
+RUN echo "set -ex" > docker.sh
 # get packages for build
 RUN echo "apt-get -y install $BUILDDEPS make" >> docker.sh
 # create git repo just so the (unconditional) submodule update recipes don't fail
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,13 +42,12 @@ WORKDIR /build
 #  so we must rely on .dockerignore here)
 COPY . .
 
-ENV BUILDDEPS="build-essential automake autoconf libtool pkg-config git"
-ENV PIP_INSTALL="pip install --timeout=300 -q"
+ENV PIP_INSTALL="pip install --timeout=3000 -q"
 
 # start a shell script (so we can comment individual steps here)
 RUN echo "set -ex" > docker.sh
 # get packages for build
-RUN echo "apt-get -y install $BUILDDEPS make" >> docker.sh
+RUN echo "apt-get -y install build-essential automake autoconf libtool pkg-config git make" >> docker.sh
 # create git repo just so the (unconditional) submodule update recipes don't fail
 RUN echo "git init" >> docker.sh
 # we want to use PREFIX as venv
@@ -61,8 +60,6 @@ RUN echo "make -j install-tesseract" >> docker.sh
 RUN echo "make ${OCRD_EXECUTABLES}" >> docker.sh
 # post-install fixup against conflicting requirements
 RUN echo "make fix-pip" >> docker.sh
-# remove build pkgs, but keep `make` for makefile-based workflow processing
-RUN echo "apt-get -y remove $BUILDDEPS" >> docker.sh
 # remove unneeded automatic deps and clear pkg cache
 RUN echo "apt-get -y autoremove && apt-get clean" >> docker.sh
 # remove source directories from image
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 # Python version (python3 required).
 PYTHON := python3
-PIP_INSTALL := pip3 install
+PIP_INSTALL ?= pip3 install
 
 # directory for virtual Python environment
 # (but re-use if already active):
@@ -75,13 +75,13 @@ $(OCRD_MODULES): always-update
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
 	$(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && pip install --upgrade pip
+	. $(ACTIVATE_VENV) && $(PIP_INSTALL) --upgrade pip
 
 # Get Python modules.
 
 # avoid making this .PHONY so it does not have to be repeated
 $(SHARE)/numpy: | $(ACTIVATE_VENV)
-	. $(ACTIVATE_VENV) && pip install numpy
+	. $(ACTIVATE_VENV) && $(PIP_INSTALL) numpy
 	@touch $@
 
 OCRD_EXECUTABLES += $(OCRD_KRAKEN)
--- a/Makefile
+++ b/Makefile
@@ -356,7 +356,7 @@ endef
 deps-ubuntu: CLSTM_DEPS = scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig
 deps-ubuntu: $(CUSTOM_DEPS)
 	set -e; for dir in $^; do make -C $$dir deps-ubuntu; done
-	apt-get -y install wget build-essential automake python3-venv $(CLSTM_DEPS)
+	apt-get -y install wget build-essential automake $(CLSTM_DEPS)
 
 .PHONY: docker
 docker: DOCKER_TAG ?= ocrd/all
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,8 @@ $(OCRD_OCROPY): ocrd_ocropy
 ocrd: $(BIN)/ocrd
 $(BIN)/ocrd: core
 	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="$(PIP_INSTALL) --force-reinstall"
+	# workaround for core#351:
+	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="$(PIP_INSTALL) --no-deps"
 
 .PHONY: wheel
 wheel: $(BIN)/wheel
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ PKG_CONFIG_PATH := $(VIRTUAL_ENV)/lib/pkgconfig
 export PKG_CONFIG_PATH
 
 OCRD_EXECUTABLES = $(BIN)/ocrd # add more CLIs below
-CUSTOM_INSTALL := $(BIN)/ocrd # add more non-pip installation targets below
-CUSTOM_DEPS := core # add more modules which need deps-ubuntu below
+CUSTOM_INSTALL = $(BIN)/ocrd # add more non-pip installation targets below
+CUSTOM_DEPS = core # add more modules which need deps-ubuntu below
 
 OCRD_MODULES := $(shell git submodule status | while read commit dir ref; do echo $$dir; done)
 
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,7 @@ $(OCRD_EXECUTABLES): | $(BIN)/wheel
 # - tensorflow>=2.0, tensorflow_gpu in another version
 # - pillow==5.4.1 instead of >=6.2
 fix-pip:
-	$(PIP_INSTALL) --force-reinstall \
+	. $(ACTIVATE_VENV) && $(PIP_INSTALL) --force-reinstall \
 		opencv-python-headless \
 		pillow>=6.2.0 \
 		$(pip list | grep tensorflow-gpu | sed -E 's/-gpu +/==/')
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ PIP_OPTIONS :=
 # directory for virtual Python environment
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
+export VIRTUAL_ENV
 
 BIN := $(VIRTUAL_ENV)/bin
 SHARE := $(VIRTUAL_ENV)/share
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 # Python version (python3 required).
 PYTHON := python3
-PIP_OPTIONS :=
+PIP_INSTALL := pip3 install
 
 # directory for virtual Python environment
 # (but re-use if already active):
@@ -53,7 +53,7 @@ Targets:
 Variables:
 	VIRTUAL_ENV: path to (re-)use for the virtual environment
 	PYTHON: name of the Python binary
-	PIP_OPTIONS: extra options to pass pip install like -q or -v
+	PIP_INSTALL: pass extra options to "pip install" like -q or -v
 EOF
 endef
 export HELP
@@ -100,12 +100,12 @@ $(OCRD_OCROPY): ocrd_ocropy
 .PHONY: ocrd
 ocrd: $(BIN)/ocrd
 $(BIN)/ocrd: core
-	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="pip install --force-reinstall $(PIP_OPTIONS)"
+	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="$(PIP_INSTALL) --force-reinstall"
 
 .PHONY: wheel
 wheel: $(BIN)/wheel
 $(BIN)/wheel: | $(ACTIVATE_VENV)
-	. $(ACTIVATE_VENV) && pip install --force-reinstall $(PIP_OPTIONS) wheel
+	. $(ACTIVATE_VENV) && $(PIP_INSTALL) --force-reinstall wheel
 
 # Install Python modules from local code.
 
@@ -250,13 +250,13 @@ $(BIN)/ocrd-make: workflow-configuration
 # install again forcefully without depds (to ensure
 # the binary itself updates):
 $(filter-out $(CUSTOM_INSTALL),$(OCRD_EXECUTABLES)):
-	. $(ACTIVATE_VENV) && cd $< && pip install $(PIP_OPTIONS) .
-	. $(ACTIVATE_VENV) && cd $< && pip install --no-deps --force-reinstall $(PIP_OPTIONS) .
+	. $(ACTIVATE_VENV) && cd $< && $(PIP_INSTALL) .
+	. $(ACTIVATE_VENV) && cd $< && $(PIP_INSTALL) --no-deps --force-reinstall .
 
 # avoid making these .PHONY so they do not have to be repeated:
 # clstm tesserocr
 $(SHARE)/%: % | $(ACTIVATE_VENV)
-	. $(ACTIVATE_VENV) && cd $< && pip install $(PIP_OPTIONS) .
+	. $(ACTIVATE_VENV) && cd $< && $(PIP_INSTALL) .
 	@touch $@
 
 # At last, add venv dependency (must not become first):
@@ -270,7 +270,7 @@ $(OCRD_EXECUTABLES): | $(BIN)/wheel
 # - tensorflow>=2.0, tensorflow_gpu in another version
 # - pillow==5.4.1 instead of >=6.2
 fix-pip:
-	pip install $(PIP_OPTIONS) --force-reinstall \
+	$(PIP_INSTALL) --force-reinstall \
 		opencv-python-headless \
 		pillow>=6.2.0 \
 		$(pip list | grep tensorflow-gpu | sed -E 's/-gpu +/==/')
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ PIP_INSTALL ?= pip3 install
 # directory for virtual Python environment
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
-export VIRTUAL_ENV
 
 BIN := $(VIRTUAL_ENV)/bin
 SHARE := $(VIRTUAL_ENV)/share
--- a/Makefile
+++ b/Makefile
@@ -356,7 +356,7 @@ endef
 deps-ubuntu: CLSTM_DEPS = scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig
 deps-ubuntu: $(CUSTOM_DEPS)
 	set -e; for dir in $^; do make -C $$dir deps-ubuntu; done
-	apt-get -y install wget python3-venv $(CLSTM_DEPS)
+	apt-get -y install wget build-essential automake python3-venv $(CLSTM_DEPS)
 
 .PHONY: docker
 docker: DOCKER_TAG ?= ocrd/all
--- a/Makefile
+++ b/Makefile
@@ -142,14 +142,14 @@ CUSTOM_INSTALL += $(BIN)/ocrd-im6convert
 CUSTOM_DEPS += ocrd_im6convert
 
 $(BIN)/ocrd-im6convert: ocrd_im6convert
-	. $(ACTIVATE_VENV) && cd $< && make install
+	. $(ACTIVATE_VENV) && cd $< && $(MAKE) install
 
 OCRD_EXECUTABLES += $(BIN)/ocrd-olena-binarize
 CUSTOM_INSTALL += $(BIN)/ocrd-olena-binarize
 CUSTOM_DEPS += ocrd_olena
 
 $(BIN)/ocrd-olena-binarize: ocrd_olena
-	. $(ACTIVATE_VENV) && cd $< && make install
+	. $(ACTIVATE_VENV) && cd $< && $(MAKE) install
 
 OCRD_EXECUTABLES += $(BIN)/ocrd-dinglehopper
 
@@ -238,7 +238,7 @@ WORKFLOW_CONFIGURATION := $(BIN)/ocrd-make
 WORKFLOW_CONFIGURATION += $(BIN)/ocrd-import
 
 $(BIN)/ocrd-make: workflow-configuration
-	cd $< && make install
+	cd $< && $(MAKE) install
 
 # Most recipes install more than one tool at once,
 # which make does not know; To avoid races, these
@@ -354,9 +354,10 @@ endef
 # (mainly intended for docker, not recommended for live systems)
 # FIXME: we should find a way to filter based on the actual executables required
 deps-ubuntu: CLSTM_DEPS = scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig
+deps-ubuntu: TESSERACT_DEPS = g++ make automake
 deps-ubuntu: $(CUSTOM_DEPS)
-	set -e; for dir in $^; do make -C $$dir deps-ubuntu; done
-	apt-get -y install wget build-essential automake $(CLSTM_DEPS)
+	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
+	apt-get -y install wget python3-venv $(TESSERACT_DEPS) $(CLSTM_DEPS)
 
 .PHONY: docker
 docker: DOCKER_TAG ?= ocrd/all
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ Targets:
 Variables:
 	VIRTUAL_ENV: path to (re-)use for the virtual environment
 	PYTHON: name of the Python binary
-	PIP_INSTALL: pass extra options to "pip install" like -q or -v
+	PIP_INSTALL: `pip install` command, optionally with extra options like `-q` or `-v`
 EOF
 endef
 export HELP
--- a/README.md
+++ b/README.md
@@ -11,22 +11,22 @@ Make sure that there is enough free disk space. 3 GiB or more is recommended for
 the required submodules, build data, temporary data, installed virtual environment
 and pip cache.
 
-Install GNU git and make:
+Install GNU make and git, and wget if you want to download Tesseract models.
 
-    # Debian / Ubuntu packages.
-    sudo apt install git make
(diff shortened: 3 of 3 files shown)


stweil:
@kba, should we move the repository to OCR-D? I suggest to keep the name then because I think that just `all` is too generic.

README.md
```diff
@@ -105,20 +115,32 @@ The following Python modules need an installation from code for different reason
...
+
+- `Pillow`:
+   * `==5.4.1` (required by ocrd_typegroups_classifier)
+   * `>=6.2.0` (required by all others)
+- Tensorflow:
+   * `tensorflow-gpu==1.14.0` (required by ocrd_calamari and OCR-D-LAYoutERkennung)
+   * `tensorflow` (which pulls in `>=2.0` which is incompatible; required by cor-asv-ann and ocrd_keraslm)
+- OpenCV:
+   * `opencv-python-headlesss` (required by core and others, avoids pulling in X11 libraries)
+   * `opencv-python` (required by OCR-D-LAYoutERkennung and segmentation-runner)
```
stweil: @mjenckel, @crater2150, is there a chance to change that?

README.md
```diff
@@ -105,20 +115,32 @@ The following Python modules need an installation from code for different reason
...
 
-- `Pillow` (`==5.4.1` vs `>=6.2.0`)
-- `tensorflow` (`-gpu` vs CPU, `<2.0` vs `>=2.0`)
+_pip does not stop or resolve conflicts – it merely warns._
+
+- `Pillow`:
+   * `==5.4.1` (required by ocrd_typegroups_classifier)
+   * `>=6.2.0` (required by all others)
+- Tensorflow:
+   * `tensorflow-gpu==1.14.0` (required by ocrd_calamari and OCR-D-LAYoutERkennung)
```
stweil: @mjenckel, would `tensorflow-gpu` 2.0.0 also work or is a code update to support that possible for `OCR-D-LAYoutERkennung`?

bertsky: In this instance, it will be easy for me to hold `ocrd_keraslm` and `cor-asv-ann` below 2.0. But the problem will likely re-surface when the first module project wants to go 2.0. (Calamari is already, so ocrd_calamari might follow soon.) The problem then is not only changes in the API (that are rather easy to adapt to) but also that models must be retrained.

BTW, as I explained in #3, there is no `tensorflow-gpu` for 2.0 (or even 1.15) anymore.

bertsky:
@stweil, regarding the pip inconsistencies, as you can see at the bottom of the Dockerfile, I alreaddy added a workaround there (that can be updated as long as no actual incompatibilities are involved).

Do you want me to move that to the Makefile so it is available for the local installation as well?

README.md
bertsky: > In this instance, it will be easy for me to hold `ocrd_keraslm` and `cor-asv-ann` below 2.0.

Done.

bertsky:
> Do you want me to move that to the Makefile so it is available for the local installation as well?

Done.

@stweil Can you merge?

README.md
crater2150: I didn't know about the headless version of opencv, I'm glad to reduce the amount of pulled in dependencies. I just updated our module on github and PyPI (version is now 0.1.2) to use opencv-python-headless.

bertsky: @crater2150 splendid, thanks! I just updated to include your new version.

README.md
```diff
@@ -11,26 +11,34 @@ Make sure that there is enough free disk space. 3 GiB or more is recommended for
...
+    # on Debian / Ubuntu:
+    sudo apt-get install make git wget
+
 Install the packages for Python3 development and for Python3 virtual environments
 for your operating system / distribution.
 
-    # Debian / Ubuntu packages.
-    sudo apt install python3-dev python3-venv
+    # on Debian / Ubuntu:
+    sudo apt-get install python3-dev python3-venv
```
stweil: Why did you change those two lines?

Makefile
```diff
@@ -7,17 +7,20 @@ PIP_OPTIONS :=
 # directory for virtual Python environment
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
+export VIRTUAL_ENV
```
stweil: Why is this needed?

stweil:
@bertsky, I started with cherry-picking changes from this pull request, so you'll have to rebase your branch.

README.md
bertsky: Because `apt` itself says:

> WARNING : apt does not have a stable CLI interface. Use with caution in scripts.

In contrast, `apt-get` has been stable for long and is already well received.

bertsky:
> I could not find the updated revision for ocrd_im6convert upstream.

Because [it is not merged yet](https://github.com/OCR-D/ocrd_im6convert/pulls).

> It would be much easier for me to review and merge changes presented in small pull requests which address a single feature and only contain one or two commits.

I see. But then those PRs would heavily depend on each other. And some things just belong in one PR. It's been quite difficult making this docker thing work. And now the automatic dependencies for local installation.

> @bertsky, I started with cherry-picking changes from this pull request, so you'll have to rebase your branch.

Please don't do this to me.

Makefile
bertsky: To give it to the shell in recipes. Make only exports variables that it already gets from the environment by default (not those defined within the file).

It is needed in various modules' `make install` (those modules that also re-use that variable).

bertsky:
@stweil When you take this PR apart, are you sure you want to test _every single change here_ with the huge long docker build, which for performance reasons is non-incremental in the commited version? It was already extremely time-consuming having this repeated layer by layer when I made changes. And testing is a must here, you cannot rely on anything...

README.md
stweil: So use `apt-get` in scripts and `Dockerfile`. `apt` was designed to be the better tool for users (progress bar, user friendly messages). That's why I'd stick to it in the README.de.

Makefile
stweil: I try to avoid environment variables whereever possible because I often had made bad experiences with them. Shell commands in recipes which need it can do that explicitely by activating the virtual environment.

If we set `VIRTUAL_ENV` in the environment of `make`, we claim that the process is running in a virtual Python environment. That claim is false because `make` can be started without such environment, and at least the `PATH` settings would be wrong.

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# create git repo just so the (unconditional) submodule update recipes don't fail
+RUN echo "git init" >> docker.sh
+# we want to use PREFIX as venv
+RUN echo "> $PREFIX/bin/activate" >> docker.sh
+# try to fetch all modules system requirements
+RUN echo "make -i deps-ubuntu" >> docker.sh
+# instead of the PPA version:
+RUN echo "make -j install-tesseract" >> docker.sh
+# build/install all requested tools:
+RUN echo "make ${OCRD_EXECUTABLES}" >> docker.sh
```
kba: Ah, nly this will be interpolated before the line is executed, right?

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# remove build pkgs, but keep `make` for makefile-based workflow processing
+RUN echo "apt-get -y remove $BUILDDEPS" >> docker.sh
+# remove unneeded automatic deps and clear pkg cache
+RUN echo "apt-get -y autoremove && apt-get clean" >> docker.sh
+# remove source directories from image
+RUN echo "rm -fr /build" >> docker.sh
+# run the script in one layer/step (to minimise image size)
+RUN bash docker.sh
+
+ENV DEBIAN_FRONTEND teletype
```
kba: Curious. Why?

Makefile
kba: I personally love environment variables and use a blank `export` statement in most makefiles because I can then pass arbitrary variables to `make` and they get passed on to shells and scripts.

That being said, I also would not recommend to *set* `VIRTUAL_ENV` since that should be set by the `bin/activate` script. As @stweil says, if you just set the variable, `PATH` won't be set.

Makefile
```diff
@@ -2,22 +2,27 @@
...
 
 BIN := $(VIRTUAL_ENV)/bin
 SHARE := $(VIRTUAL_ENV)/share
 ACTIVATE_VENV := $(VIRTUAL_ENV)/bin/activate
 
-WGET := $(if $(shell which wget),wget -O,$(if $(shell which curl),curl -o,$(error "found no cmdline downloader (wget/curl)")))
+define WGET
```
kba: It's a sweet solution with the macro but why not require one of wget or curl?

Makefile
```diff
@@ -304,6 +334,37 @@ $(BIN)/tesseract: tesseract/configure
 # suppress all built-in suffix rules:
 .SUFFIXES:
 
+# get the modules required by the given (1) executables
```
kba: "by parsing make debug output"?

Makefile
```diff
@@ -304,6 +334,37 @@ $(BIN)/tesseract: tesseract/configure
...
+define requires-modules
+$(shell LC_MESSAGES=C $(MAKE) -R -nd $(1) 2>&1 | \
+	fgrep -e 'Considering target file' -e 'Trying rule prerequisite' | \
+	cut -d\' -f2 | tr ' ' '\n' | fgrep -x $(OCRD_MODULES:%=-e %))
+endef
+# get only those (1) executables which depend on the given (2) set of allowed modules
+define filter-executables
+$(foreach tool, $(1), $(if $(filter-out $(2),$(call requires-modules, $(tool))),,$(tool)))
```
kba: That is some advanced makefile magic, a few more words what happens here would help maintaining.

README.md
```diff
@@ -105,20 +125,32 @@ The following Python modules need an installation from code for different reason
 - cor-asv-ann (not available in PyPI)
 - cor-asv-fst (not available in PyPI)
 - dinglehopper (not available in PyPI)
-- ocrd_cis (not available in PyPI; needs `ocrd>=2.0.0` / core branch `edge`)
-- ocrd_tesserocr (too old in PyPI; needs `ocrd>=2.0.0` / core branch `edge`)
+- ocrd_cis (not available in PyPI; needs `ocrd>=2.0.0`)
+- ocrd_tesserocr (too old in PyPI; needs `ocrd>=2.0.0`)
```
kba: Should be up-to-date in PyPI now?

README.md
kba: https://github.com/mjenckel/OCR-D-LAYoutERkennung/pull/20

kba:
> I could not find the updated revision for ocrd_im6convert upstream.
> 
> It would be much easier for me to review and merge changes presented in small pull requests which address a single feature and only contain one or two commits.

Yeah, maybe keep the submodule updates out of this PR.

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# use OCR-D base container (from ubuntu:18.04)
+FROM ocrd/core
+
+MAINTAINER OCR-D
+
+ENV PREFIX=/usr/local
+ENV VIRTUAL_ENV $PREFIX
+
+# make apt run non-interactive during build
+ENV DEBIAN_FRONTEND noninteractive
```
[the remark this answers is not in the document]
bertsky: Now that you are asking, I am not so sure anymore. 

ocrd/core is probably not going to be used for interactive stuff, but for module containers. 

But ocrd/all will quite likely benefit from the teletype default IMO.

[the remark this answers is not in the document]
bertsky: But it does not hurt to make the desired initial state robust and explicit here.

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# copy (sub)module directories to build
+# (ADD/COPY cannot copy a list of directories,
+#  so we must rely on .dockerignore here)
+COPY . .
+
+ENV BUILDDEPS="build-essential automake autoconf libtool pkg-config git"
+ENV PIP_INSTALL="pip install --timeout=300 -q"
+
+# start a shell script (so we can comment individual steps here)
```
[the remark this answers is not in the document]
bertsky: env vars are expanded. The point here was to have both:
- everything belonging to the docker setup in one file, with reasons (in comments) why this or that is done as it is
- getting as small as possible a resulting image, so putting all steps that add temporary data in one layer

People on Docker support have asked for various features to get that (here document syntax, better multi-line syntax etc). But this is a simple and reliable solution IMO.

Dockerfile
bertsky: Yes, execute is at the very bottom. All those RUNs do is create small layers of `docker.sh` increments. So changes to these lines will invalidate the cache just fine. But all cleanup happens in the same layer as the bloatup.

Dockerfile
bertsky: See above (to make interactive sessions smoother – when the user wants to install more packages). (I think this is the default in `ubuntu:18.04` as well.)

Makefile
bertsky: > That being said, I also would not recommend to set VIRTUAL_ENV since that should be set by the bin/activate script. As @stweil says, if you just set the variable, PATH won't be set.

(Cannot answer in context above for some reason.)

Well but this is not what this is about at all. It already __re-uses__ the setting from the environment. So if the user did `bin/activate`, then nothing would change here. (Remember, make always passes on its invocation environment to shells, which includes `PATH` anyway.)

The reason for this is what should happen when the user does _not_ call from a venv context. In that case, the Makefile _creates_ one in various rules. That was the idea Stefan based this repo on. But we have the same behaviour in other modules already, e.g. ocrd_im6convert or ocrd_olena. Now, we wouldn't want _those_ to create venvs or other installation prefixes of their own, would we?

Makefile
bertsky: > That being said, I also would not recommend to _set_ `VIRTUAL_ENV` since that should be set by the `bin/activate` script. As @stweil says, if you just set the variable, `PATH` won't be set.

Well but this is not what this is about at all. It already re-uses the setting from the environment. So if the user did bin/activate, then nothing would change here. (Remember, make always passes on its invocation environment to shells, which includes PATH anyway.)

The reason for this is what should happen when the user does not call from a venv context. In that case, the Makefile creates one in various rules. That was the idea Stefan based this repo on. But we have the same behaviour in other modules already, e.g. ocrd_im6convert or ocrd_olena. Now, we wouldn't want those to create venvs or other installation prefixes of their own, would we?

bertsky: > If we set `VIRTUAL_ENV` in the environment of `make`, we claim that the process is running in a virtual Python environment.

No, we don't make any such claim. In fact, all the rules that create a new venv and activate it as needed are still active.

And if hooking onto `VIRTUAL_ENV` was forbidden, then we should not do it in the first place.

Makefile
bertsky: Yes, I am afraid there is no other way. I know GNU make quite well, and I have specifically searched the mailing list for this. It's what the mainainers would have you do instead of adding features for this into the syntax themselves (like functions for dumping dependencies).

Makefile
bertsky: You mean like actual words in the comment, or merely an explanation here?

This is straightforward use of makefile functions:

For each CLI `tool` you pass in under `$1`, get all its prerequisites, and among those filter only the module names. Now filter out those module prerequisites which are already passed in under `$2` (i.e. the allowed modules), and check if you get any results at all. If so, then the tool obviously depends on more than the allowed modules, so we skip it. Otherwise, we echo it.

README.md
bertsky: Not `ocrd_cis`, but `ocrd_tesserocr` I can remove from the list now completely, yes.

bertsky:
> Yeah, maybe keep the submodule updates out of this PR.

I don't understand you guys. Is no-one seeing the connection? Almost all those updates _were necessary_ to get the new functionality (automatic system dependencies, Docker integration, module filtering, workarounds for conflicts) working. In many cases, I conjured up the module updates just for this PR!

With this attitude (keep it small, split it up and rebase), you are making it unnecessarily hard (and frustrating) to bring in solutions. It was already costly to fight through this (waiting for docker build to fail over and over again, wasting tons of network bandwidth, making all this work automatically and maintainably). 

I don't want go over all those steps and test them again for no good reason. 

You can easily `make all` / `make docker` on the PR as a whole – if that fails, we should talk.

README.md
bertsky: Yes you are right. The above commands are inteded for input by a human, not in a script. I was biased by my recent work in Docker. But people will have to make changes when copying instructions into scripts anyway (like removing `sudo`). I'll use `sudo apt` throughout the paragraph.

stweil:
@bertsky, I now get an ocrd/all docker of 5.02 GB when I run `make docker` with the latest code which includes most of your modifications. Which tests would you suggest to verify that everything is fine?

Makefile
```diff
@@ -304,6 +334,37 @@ $(BIN)/tesseract: tesseract/configure
...
+endef
+
+# allow installing system dependencies for all modules
+# (mainly intended for docker, not recommended for live systems)
+# FIXME: we should find a way to filter based on the actual executables required
+deps-ubuntu: CLSTM_DEPS = scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig
+deps-ubuntu: $(CUSTOM_DEPS)
```
stweil: Why should `deps-ubuntu` depend on several submodules?

bertsky: Precisely because we want to delegate to their own knowledge of their system dependencies, as encapsulated by their own `deps-ubuntu`. It's in the recipe!

bertsky:
> @bertsky, I now get an ocrd/all docker of 5.02 GB when I run make docker with the latest code which includes most of your modifications. Which tests would you suggest to verify that everything is fine?

Well that's already a (partial) success. (It should be around 7 GB though. Maybe intermediate steps have failed? I recently added `set -e` to docker.sh to avoid overlooking problems. I'll add that to the PR as soon as the [new problem in core](https://github.com/OCR-D/core/issues/351) is fixed.)

Next steps would be to run some workflow and validate. I am afraid there are no smoke tests for that in core or assets. But you can use `workflow-configuration` – this is what I do. For example, run `docker run -it -u $(id -u) -v /path/to/workspaces:/data ocrd/all ocrd-make -f gt-binarize-page-olena-sauvola-deskew-page-ocropy-clip-deskew-region-tesseract-resegment-dewarp-ocr-ocropy-tesseract.mk`.

Makefile
stweil: Hm. Those recipes seem to be very generous and install much more than needed. Especially they install Python modules which are completely useless in a virtual Python environment ...

bertsky:
@stweil Note, there seems to be connectivity problems with GH these days – you might have to repeat steps.

Makefile
stweil: As a typical user will run `sudo make deps-ubuntu`, she will end with files with owner root in her working directory. That's not desired.

bertsky: > Hm. Those recipes seem to be very generous and install much more than needed. Especially they install Python modules which are completely useless in a virtual Python environment ...

Which ones, where? Apart from `python3-pip python3-venv` I can only see non-Python packages.

And if an OCR-D module tries to install its Python dependencies via `deps-ubuntu` then that would be an error of that module (to be fixed independently). But I don't see that ATM.

bertsky: > As a typical user will run `sudo make deps-ubuntu`, she will end with files with owner root in her working directory. That's not desired.

There's no need to paint the devil. For now we don't have such errors in submodules, and if there are, these revisions must not be commited here until they are fixed, or if they are commited, then they need workarounds in the recursive `deps-ubuntu` recipe.

Furthermore, as a user, I would _never_ `sudo make something` without running `make -n something` first.

stweil:
@bertsky, I rebased the pull request now. The result only differs from the previous one in some submodule revisions, partly because they are now newer, partly because the old pull branch referred to non existing revisions.

Dockerfile
```diff
@@ -42,13 +42,12 @@ WORKDIR /build
...
-ENV BUILDDEPS="build-essential automake autoconf libtool pkg-config git"
-ENV PIP_INSTALL="pip install --timeout=300 -q"
+ENV PIP_INSTALL="pip install --timeout=3000 -q"
 
 # start a shell script (so we can comment individual steps here)
-RUN echo "set -x" > docker.sh
+RUN echo "set -ex" > docker.sh
 # get packages for build
-RUN echo "apt-get -y install $BUILDDEPS make" >> docker.sh
+RUN echo "apt-get -y install build-essential automake autoconf libtool pkg-config git make" >> docker.sh
```
stweil: "If you do not plan to build Debian packages, you don't need this package." (cited from the documentation for `build-essential`. Do you plan to build Debian packages?

Makefile
stweil: Well, that shows one more problem with the current code: `make deps-ubuntu -n` does not tell me all commands which would be executed. I still have to follow the `for` loop which runs `make deps-ubuntu` for some modules manually. And if I do that, I see that it installs unwanted Tesseract binaries from an untrusted third party source, for example.

Makefile
```diff
@@ -353,7 +356,7 @@ endef
 deps-ubuntu: CLSTM_DEPS = scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig
 deps-ubuntu: $(CUSTOM_DEPS)
 	set -e; for dir in $^; do make -C $$dir deps-ubuntu; done
-	apt-get -y install wget python3-venv $(CLSTM_DEPS)
+	apt-get -y install wget build-essential automake $(CLSTM_DEPS)
```
stweil: See my other comment on `build-essential`.

bertsky:
> @bertsky, I rebased the pull request now. The result only differs from the previous one in some submodule revisions, partly because they are now newer, partly because the old pull branch referred to non existing revisions.

@stweil, I cannot compare this, due to the rebase (that's why I opted for a merge commit instead). 

Which submodules are different, what revisions? (This is important, because it might break things again.)

You still say non-existing revisions, but as I [answered you already](https://github.com/stweil/ocrd_all/pull/6#issuecomment-555766105), these were perfectly valid revisions, just not on the module's master (yet).

Dockerfile
bertsky: That documentation is non-sense. It should read: `to build packages`, not _Debian_ packages specifically.

What are you implying, that I should remove the compilers in this build?

Makefile
bertsky: You cannot build Tesseract without a compiler. See changelog.

Dockerfile
stweil: That's the official Debian documentation, and it is correct. `build-essentials` depends on packages like `dpkg-dev` which are not relevant for normal builds like those required for `ocrd_all`. Just replace it by `g++` and `make`.

stweil:
>You still say non-existing revisions, but as I answered you already, these were perfectly valid revisions, just not on the module's master (yet).

A new revision from a pull request only exists in the repository from which that pull request originated. Such a revision is unavailable for most people.

The rule for a valid submodule revision is simple: the revision must exist in the repository of the submodule (not necessarily in the master branch). If it only exists in your fork, it is not valid.

stweil:
> @stweil, I cannot compare this, due to the rebase (that's why I opted for a merge commit instead).

These commands might work for your environment:

    git fetch
    git diff origin/fat-docker

Makefile
stweil: > You cannot build Tesseract without a compiler.

That's surprising news. 😃

Makefile
bertsky: > Well, that shows one more problem with the current code: `make deps-ubuntu -n` does not tell me all commands which would be executed. I still have to follow the `for` loop which runs `make deps-ubuntu` for some modules manually.

That's right, I forgot to use the `MAKE` variable instead of just using `make`.

I [fixed](https://github.com/bertsky/ocrd_all/commit/2c0016810aeb74662a9ffee053794377661b1acb) that on my master now. Your rebasing (and the uncertainty whether you preserved all my changes undamaged) increased the effort way too much for me to bring the forks in sync again. This is all _very_ frustrating.

bertsky:
> A new revision from a pull request only exists in the repository from which that pull request originated. Such a revision is unavailable for most people.

That's simply not true: A PR's commits are automatically added to a new branch `pull/NUM/head` on the receiving end. Did you even try to just `git submodule update`?

bertsky:
> These commands might work for your environment:
> 
> ```
> git fetch
> git diff origin/fat-docker
> ```

The updates seem to be ok (but I have not rebuilt and tested the docker images yet). Except for ocrd_im6convert, which definitely _needs_ https://github.com/OCR-D/ocrd_im6convert/pull/9/commits/1b082f072158bc2d044deea9bd4fc90e02321af8 (because `deps-ubuntu` must be run non-interactively), as I said.

Dockerfile
bertsky: Ok, fine. But you would have to replace it by `libc6-dev gcc g++ make`. (I used `build-essential` because it is used in core.)

stweil: `g++` directly depends on `gcc` and indirectly also on `libc6-dev`, so those are redundant.

bertsky: Right, `g++ make` it is.

Makefile
bertsky: So, I added `python3-venv` again (although it's already in core's `deps-ubuntu`, because it does not hurt to have it here as well), and replaced `build-essential` with `g++ make` as discussed above.

bertsky:
> The updates seem to be ok (but I have not rebuilt and tested the docker images yet). Except for ocrd_im6convert, which definitely _needs_ [OCR-D/ocrd_im6convert@1b082f0](https://github.com/OCR-D/ocrd_im6convert/commit/1b082f072158bc2d044deea9bd4fc90e02321af8) (because `deps-ubuntu` must be run non-interactively), as I said.

I reverted to the needed version, which is now also on ocrd_im6convert:master.

bertsky:
> > A new revision from a pull request only exists in the repository from which that pull request originated. Such a revision is unavailable for most people.
> 
> That's simply not true: A PR's commits are automatically added to a new branch `pull/NUM/head` on the receiving end. Did you even try to just `git submodule update`?

With even that out of the way, can you finally merge?

Makefile
bertsky: So you have [copied](https://github.com/stweil/ocrd_all/commit/9509584ecdf9d3b31729cd179fa2baa6e7ec1e3a) my change, fine.

Dockerfile
stweil:
```suggestion
RUN echo "apt-get -y install automake autoconf libtool pkg-config g++ git make" >> docker.sh
```

Makefile
```diff
@@ -2,11 +2,12 @@
 
 # Python version (python3 required).
 PYTHON := python3
-PIP_OPTIONS :=
+PIP_INSTALL ?= pip3 install
```
```suggestion
PIP_INSTALL ?= pip install
```

stweil: The command must always run in a virtual environment, so `pip` should be sufficient.

kba: I would commit if I could. LGTM

Dockerfile
bertsky: I'm not allowed to commit the suggestion for some reason, but please do.

Makefile
bertsky: > The command must always run in a virtual environment, so `pip` should be sufficient.

I'm not so sure. For example, the Docker build does _not_ run in a virtual environment (but uses the same rules). What if the image has both a Python 2 and Python 3 installation?

It does not _harm_ to use `pip3` here, but it ensures the above use case still works.

Makefile
```diff
@@ -269,7 +272,7 @@ $(OCRD_EXECUTABLES): | $(BIN)/wheel
 # - tensorflow>=2.0, tensorflow_gpu in another version
 # - pillow==5.4.1 instead of >=6.2
 fix-pip:
-	pip install $(PIP_OPTIONS) --force-reinstall \
+	$(PIP_INSTALL) --force-reinstall \
```
```suggestion
	. $(ACTIVATE_VENV) && $(PIP_INSTALL) --force-reinstall \
```

Makefile
stweil: > It is needed in various modules' make install

If there still remain such cases (I did not found any), they should be handled by an explicit `. $(ACTIVATE_VENV)`.

Makefile
```diff
@@ -2,11 +2,12 @@
 
 # Python version (python3 required).
 PYTHON := python3
-PIP_OPTIONS :=
+PIP_INSTALL ?= pip3 install
 
 # directory for virtual Python environment
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
+export VIRTUAL_ENV
```
```suggestion
```

Makefile
```diff
@@ -52,7 +53,7 @@ Targets:
 Variables:
 	VIRTUAL_ENV: path to (re-)use for the virtual environment
 	PYTHON: name of the Python binary
-	PIP_OPTIONS: extra options to pass pip install like -q or -v
+	PIP_INSTALL: pass extra options to "pip install" like -q or -v
```
stweil: That description would fit for `PIP_OPTIONS`, but does not for `PIP_INSTALL`

```suggestion
	PIP_INSTALL: `pip install` command, optionally with extra options like `-q` or `-v`
```

Makefile
bertsky: Ouch, how could I miss that? Thanks!

Makefile
bertsky: > If there still remain such cases (I did not found any), they should be handled by an explicit `. $(ACTIVATE_VENV)`.

That would probably work, too.

Makefile
kba: Then keep it as pip3. Better explicit than sorry.

Makefile
bertsky: True.

bertsky:
@stweil do you want me to update core again here (significantly, we now have processor parameter validation), or are you going to apply that yourself after the PR?

Makefile
stweil: > I would commit if I could. LGTM

@kba and @bertsky, both of you should have the necessary rights (push access) for this repository if you accepted the invitation which I had sent some days ago.

Dockerfile
stweil: The conversation was marked as resolved which prohibits to commit a suggestion. Please try again.

kba: No, still doesn't work :( "You do not have permission to push to this repository. Refresh and try again" (the latter doesn't help)

stweil:
I have updated `core` now.

Dockerfile
stweil: Hm. Any suggestion what could be wrong? I see your status as "the user has been invited to collaborate ...", so it looks like you still have not accepted my invitation.

bertsky: > The conversation was marked as resolved which prohibits to commit a suggestion. Please try again.

Oh, didn't see that. Works now!

stweil: Ah, I also cannot push. It's @bertsky's repository. So @bertsky, it's up to you to try the commit again. :-)

README.md
```diff
@@ -125,32 +186,41 @@ The following Python modules need an installation from code for different reason
...
 
+_(Solved temporarily by post-installation `fix-pip`.)_
+
 ### System requirements
 
 Not all modules advertise their system package requirements via `make deps-ubuntu`.
 
 - clstm: depends on `scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig`
+
+_(Solved by maintaining these requriements under `deps-ubuntu` here.)_
```
stweil:
```suggestion
_(Solved by maintaining these requirements under `deps-ubuntu` here.)_
```

Dockerfile
bertsky: @stweil but you have pushed (and rebased) before. Strange.

Makefile
kba: I'm certain I accepted the invitation but now I don't even see the "Settings" tab. Visiting https://github.com/stweil/ocrd_all/invitations just redirects back to the repo home. Strange things  afoot in GitHub today...

PR #8, opened 2019-11-21
remove pixelwise_segmentation_SBB

cneud:
Please remove the ``pixelwise_segmentation_SBB`` repo, it 
* was never planned for OCR-D use
* does not have an OCR-D interface
* only constitutes an early research prototype
* is no longer mainted by SBB
* is only increasing the Docker container size

We are working on a proper segmentation tool that will be OCR-D compliant and released before the end of 2019.


bertsky:
I was already exluding it in my own [docker builds](https://hub.docker.com/r/bertsky/ocrd_all:nosbb) (`DOCKER_MODULES`).

>     * is no longer mainted by SBB

@cneud Does that mean https://github.com/qurator-spk/pixelwise_segmentation_SBB is no longer maintained at all, or no longer maintained by SBB?

cneud:
> @cneud Does that mean https://github.com/qurator-spk/pixelwise_segmentation_SBB is no longer maintained at all, or no longer maintained by SBB?

IMHO effectively it means "no longer at all". We had to publish this experimental code for formal reasons, but have never intended it for real use, and AFAICT no one is using it. All our efforts in recent months have been directed at producing a first working release with OCR-D interface and documentation. Once that is released, this repository will be unpublished.

PR #9, opened 2019-12-03
Add command for git clone to README.md.

VolkerHartmann:
Makefile will not work if project is cloned without explicit --recursive flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,16 @@ in the current shell environment via PATH and PYTHONHOME.)
 
 ## Preconditions
 
-### Space
+Clone this project recursively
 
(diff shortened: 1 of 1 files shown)


stweil:
Which failure did you get without `--recursive`?

I expected that `Makefile` should do all necessary recursion automatically. The user interface should be as simple as possible, and requiring flags for `git clone` would contradict that purpose.

VolkerHartmann:
I can't reproduce it today!? Yesterday there was a message like 'Berechtigung fehlt'. 
I tried it twice without success. When I used the --recursive flag it worked. I don't no what went wrong yesterday.

PR #10, opened 2019-12-04
Write temporary files to local directory by default

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,13 @@ endef
 PKG_CONFIG_PATH := $(VIRTUAL_ENV)/lib/pkgconfig
 export PKG_CONFIG_PATH
 
+# Set TMPDIR (optionally by a value passed by the user),
+# so temporary files for building won't be written to /tmp.
+# Some builds require more than 2 GB free disk space for
+# such temporary files, and /tmp might be too small for that.
+TMPDIR ?= $(shell mkdir -p $(CURDIR)/tmp && echo $(CURDIR)/tmp)
+export TMPDIR
+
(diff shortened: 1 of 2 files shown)


stweil:
This fixes a build problem on one of my virtual machines which has only 1.9 GB free in the root filesystem (and so in `/tmp`, too), but provides a large volume for `/HOME` with my user data and 25 GB free disk space.

kba:
I concur with @bertsky, let's make this optional and default to `/tmp` so you CAN do

```sh
d=$(mktemp -d -p $PWD) && \
make ocrd-anybaseocr-crop TMPDIR="$d" && \
rm -rf -- "$d"
```

but do not HAVE to set TMPDIR explicitly

stweil:
I used `TMPDIR=$PWD/tmp make all` successfully without this patch, so this modifications just makes life a little bit easier for me.

bertsky:
> I used `TMPDIR=$PWD/tmp make all` successfully without this patch, so this modifications just makes life a little bit easier for me.

Then I'd say just put `export TMPDIR=path/to/ocrd_all/tmp` in your `~/.bashrc` or in `venv/bin/activate` and you're good. (We should still add this remark to `make help` and README, though.)

kba:
I added a paragraph to the README on this.

kba:
Can you cherry pick https://github.com/UB-Mannheim/ocrd_all/commit/984863617fb23c834aeeb18d55b8573a040273cb to master though?

stweil:
Already done. I also fixed a typo, so you will get a local conflict.

PR #12, opened 2019-12-04
Ocrd forks

kba:
Use OCR-D forks of LAYoutERkennung and typegroups_classifier. We publish PyPI packages from them and they allow us more rapid development.

(this can wait until repo transfer)

PR #17, opened 2019-12-06
re-add anybaseocr, got lost in #12

kba:


stweil:
Sorry, I was faster and already fixed that.

PR #18, opened 2019-12-06
Add new submodule sbb_textline_detector

stweil:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,3 +67,6 @@
 [submodule "workflow-configuration"]
 	path = workflow-configuration
 	url = https://github.com/bertsky/workflow-configuration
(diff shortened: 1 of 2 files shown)


stweil:
That module still needs some updates (naming conventions, requirements, maybe more) to fit into the OCR-D world, so this PR is currently just for discussions.

cneud:
Yep, see also https://github.com/qurator-spk/sbb_textline_detector/pulls that I would like to get merged asap before we dive deeper into code.

kba:
Sry @stweil did not see this when I opened #19 which already contains the updated executable name. Let's continue discussion there and you can make changes to your master branch without adding to this PR.

PR #19, opened 2019-12-06
add sbb_textline_detector

kba:


cneud:
@kba @stweil If you want me to, I can also "force-merge" https://github.com/qurator-spk/sbb_textline_detector/pull/3 now and we can omit ``opencv-python`` from ``ocrd_all``?

cneud:
IIUC the current version of ``ocrd_anybaseocr`` also no longer depends on it.

stweil:
> we can omit `opencv-python` from `ocrd_all`?

The submodule is still needed for platforms like ARM or Power8 were no pre-compiled binaries exist. :-(

PR #20, opened 2019-12-06
update core, dinglehopper, ocrd_calamari, ocrd_tesserocr

kba:
core is now at 2.0.2

I didn't update tesseract, @stweil should we keep tesseract at a certain commit/tag?


stweil:
> should we keep tesseract at a certain commit/tag

There are frequent updates and fixes on master and nearly no tags. Therefore no, I'd follow master for the moment.

PR #22, opened 2019-12-06
replace pixelwise_segmentation_SBB with sbb_textline_detector

cneud:


stweil:
See also pull request #19.

Commit 3a8ce54f92e44543233429d73679a56b01fa6f31 obviously only removed part of pixelwise_segmentation_SBB, so that part is indeed still missing.

stweil:
I now removed the remaining parts of pixelwise_segmentation_SBB in commit 19fcc801ef946218128a31d6b2e9b32eaef47a68. So it is sufficient to merge #19, and this pull request can be closed.

Issue #15, opened 2019-12-05
Tesseract must not be built with disable-shared


bertsky:
@stweil, what reason did you have for `--disable-shared` in https://github.com/UB-Mannheim/ocrd_all/blob/7d2306f94a1054bd051b46cc8c864fd9fec53f88/Makefile#L334?

This breaks both tesserocr install and scribo (which also links to -ltesseract) for me when no other Tesseract (from PPA) gets installed.

stweil:
It works on a fresh Ubuntu installation. Maybe you are using cached files which still require the shared library.

stweil:
One advantage is that it does not interfere with another Tesseract installation, so there is no danger of getting the wrong shared library.

Another advantage is faster code and smaller memory footprint.

bertsky:
> It works on a fresh Ubuntu installation. Maybe you are using cached files which still require the shared library.

It does not work in the Docker build when no libtesseract.so is installed via PPA. 

The failure of `pip install tesserocr` I observed was due to another experiment (which used `-fsanitize=address` in Tesseract build). But scribo build fails like this:

```
make[6]: Entering directory '/build/ocrd_olena/olena-git/build/scribo/src/text'
  CXX      cleantxt.o
  CXXLD    cleantxt
  CXX      pbm_lines_recognition-pbm_lines_recognition.o
  CXXLD    pbm_lines_recognition
//usr/local/lib/libtesseract.a(libtesseract_api_la-baseapi.o): In function `tesseract::TessBaseAPI::Threshold(Pix**)':
/build/tesseract/build/src/api/../../../src/api/baseapi.cpp:2012: undefined reference to `pixDestroy'
```

So the `-ltesseract` is static now, and this necessitates enumerating all of its dependencies in the linker command for 3rd party components (like `-llept` in this case). (And tesserocr is special here, because its setup.py build rules already encapsulate that knowledge.)

bertsky:
So maybe this _is_ an error on the part of Tesseract, in that it fails to properly set the libs for pkg-config when installing with `--disable-shared`?

stweil:
Yes, `pkg-config` fails to give the right linker flags.

bertsky:
Ok, then I guess Tesseract's `tesseract.pc.in` is the culprit. It says:
```
Libs: -L${libdir} -ltesseract @libarchive_LIBS@ @libcurl_LIBS@ @TENSORFLOW_LIBS@
[...]

PR #24, opened 2019-12-08
Conditional modules and editable mode installation

bertsky:
This is another major leap forward. It encapsulates knowledge about submodule specifics even better (esp. system dependencies) and allows selectively factoring modules out of all targets (including `deps-ubuntu` and `docker`). It also allows installing pip packages in editable mode (to save space and enable faster update regimes, both especially helpful in Docker images). Lastly, it adds support for a (non-git controlled) local/site customization makefile.

@stweil, this also avoids installing Tesseract both from source and PPA at the same time.

@kba, this also contains preconfigured Docker builds of different extent (which you could offer on an automatic basis via Dockerhub).

I did 2dfe06a in light of #15, but we probably should find a better way to fix the issue before merging.

Please don't ask me to split 009a042 into multiple commits – it's hard and would not make the changes more comprehensible at all (due to interdependencies). I'll gladly elaborate on any individual change here, though.


Makefile
```diff
@@ -68,52 +86,63 @@ help: ;	@eval "$$HELP"
...
 	git submodule sync --recursive $@
 	git submodule status $@ | grep -q '^ ' || { \
 		git submodule update --init --recursive $@ && \
 		touch -r $$(find $@ -type f -newer $@ -o -type d -name $@ | tail -1) $@; }
+endif
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
 	$(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && $(PIP_INSTALL) --upgrade pip
+	. $(ACTIVATE_VENV) && $(PIP) install --upgrade $(PIP_OPTIONS_E) pip
```
kba: `pip>=19.0.0` ?

Makefile
```diff
@@ -1,8 +1,16 @@
...
+-include local.mk
+
 # Python version (python3 required).
 PYTHON := python3
-PIP_INSTALL ?= pip3 install
+# Python packaging
+PIP ?= pip3
+# Derived variables to allow filtering -e, or inserting other options
+# (the option --editable must always be last and only applies to src install)
+PIP_INSTALL = $(PIP) install $(PIP_OPTIONS)
```
kba: This is not used anymore, right? Probably better not to define it then and not to refer to it in README

Makefile
bertsky: I'm still not convinced we need both `>=19` and `-U`. (And I'd say we can trust pip to be evolving in a non-breaking way for now.)

Makefile
bertsky: Right!

.gitignore
```diff
@@ -1 +1,4 @@
 /venv
+local.mk
+**/build/
+*~
```
stweil: Why is `*~` needed here? If it is only used to ignore backup files made by a local editor, it should be ignored in the local gitignore settings.

Makefile
```diff
@@ -45,16 +57,22 @@ Targets:
 	ocrd: installs only the virtual environment and OCR-D/core packages
 	modules: download all submodules to the managed revision
 	all: installs all executables of all modules
+	install-tesseract: download, build and install Tesseract (with some models)
```
```suggestion
	install-tesseract: download, build and install Tesseract (with required models)
```

stweil: The three models eng, osd and equ are required, so this is a minimum set of models. Therefore those models should always be installed, and the user should only be allowed to specify additional models. That requires changes below which I don't comment separately.

Makefile
```diff
@@ -45,16 +57,22 @@ Targets:
...
+	dockers: (re)build docker images for some pre-selected subsets of modules
 
 Variables:
 	VIRTUAL_ENV: path to (re-)use for the virtual environment
 	TMPDIR: path to use for temporary storage instead of the system default
 	PYTHON: name of the Python binary
-	PIP_INSTALL: `pip install` command, optionally with extra options like `-q` or `-v`
+	PIP: name of the Python packaging binary
+	PIP_OPTIONS: extra options for the `pip install` command like `-q` or `-v` or `-e`
+	OCRD_MODULES: list of submodules to include (defaults to all git modules, see `show`)
```
```suggestion
	OCRD_MODULES: list of submodules to include (defaults to all git submodules, see `show`)
```

stweil: In the context of git the terminus is `submodule`, so I'd use that to avoid confusions.

Makefile
```diff
@@ -68,52 +86,63 @@ help: ;	@eval "$$HELP"
 #   of the latest modified file contained in it,
 #   so the directory can be used as a dependency
 modules: $(OCRD_MODULES)
+# but bypass updates if we have no repo here (e.g. Docker build)
+ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
 	git submodule sync --recursive $@
 	git submodule status $@ | grep -q '^ ' || { \
 		git submodule update --init --recursive $@ && \
 		touch -r $$(find $@ -type f -newer $@ -o -type d -name $@ | tail -1) $@; }
```
stweil: I wonder whether this should be simplified to just `touch $@` as the current code does not work as it is expected to.

Makefile
```diff
@@ -68,52 +86,63 @@ help: ;	@eval "$$HELP"
...
 OCRD_OCROPY := $(BIN)/ocrd-ocropy-segment
 
 $(OCRD_OCROPY): ocrd_ocropy
+endif
 
 .PHONY: ocrd
 ocrd: $(BIN)/ocrd
 $(BIN)/ocrd: core
-	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="$(PIP_INSTALL) --force-reinstall"
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< install PIP_INSTALL="$(PIP) install --force-reinstall $(PIP_OPTIONS)"
```
stweil: Those `--force-reinstall` options are sometimes nagging. I'd like an option to disable them in `local.mk` (after this PR was merged).

Makefile
```diff
@@ -68,52 +86,63 @@ help: ;	@eval "$$HELP"
...
 	# workaround for core#351:
-	. $(ACTIVATE_VENV) && cd $< && make install PIP_INSTALL="$(PIP_INSTALL) --no-deps"
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< install PIP_INSTALL="$(PIP) install --no-deps $(PIP_OPTIONS)"
+
+deps-ubuntu: core
```
stweil: I still think that `deps-ubuntu` which must be run with root rights should not trigger Makefile targets which should be run with user rights. We just recently had a build problem which was probably caused by that.

Makefile
```diff
@@ -221,27 +283,37 @@ OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-layout-analysis
...
 
 OCRD_TYPECLASS := $(BIN)/ocrd-typegroups-classifier
 OCRD_TYPECLASS += $(BIN)/typegroups-classifier
 
 $(OCRD_TYPECLASS): ocrd_typegroups_classifier
+endif
 
+ifneq ($(findstring sbb_textline_detector, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(SBB_LINE_DETECTOR)
+
```
stweil: Personally I prefer the more compact form without empty lines, but that is a matter of taste.

Makefile
```diff
@@ -294,20 +366,27 @@ show:
 .PHONY: $(OCRD_EXECUTABLES:$(BIN)/%=%)
 $(OCRD_EXECUTABLES:$(BIN)/%=%): %: $(BIN)/%
 
+ifneq ($(findstring tesseract, $(OCRD_MODULES)),)
 # Tesseract.
 
+# when not installing via PPA, we must cope without ocrd_tesserocr's deps-ubuntu
+CUSTOM_DEPS += g++ make automake libleptonica-dev
+# but since we are building statically, we need more (static) libs at build time
+CUSTOM_DEPS += libgif-dev libtiff-dev libpng-dev libjpeg-dev
```
```suggestion
CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
```

README.md
```diff
@@ -228,6 +278,7 @@ _(Solved temporarily by post-installation `fix-pip`.)_
 
 Not all modules advertise their system package requirements via `make deps-ubuntu`.
 
-- clstm: depends on `scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig`
+- `clstm`: depends on `scons libprotobuf-dev protobuf-compiler libpng-dev libeigen3-dev swig`
+- `tesseract` (when installing from source not PPA): depends on `libleptonica-dev`
```
stweil:
```suggestion
- `tesseract` (when installing from source not PPA): depends on `libarchive-dev libcurl4-nss-dev libleptonica-dev`
```

.gitignore
bertsky: > Why is `*~` needed here? If it is only used to ignore backup files made by a local editor,

It is.

> it should be ignored in the local gitignore settings.

Why? If I edit Makefile or local.mk, I don't want to see `Makefile~` or `local.mk~` as unknown.

Makefile
bertsky: I don't understand – where does it not work? The point in using a reference file for the time stamp is that we don't want to cause arbitrary updates, only ones backed by actual changed files.

.gitignore
stweil: So it's a setting for you and you can add it to your `~/.gitignore` or to `.git/info/exclude`. Maybe other users want to see those `*~`files.

The general rule is to add only files which are produced by regular usage to `.gitignore`.

Makefile
bertsky: Sure! We could wrap them in a variable with `install --force-reinstall` that you can then override.

But I'm curious: what problems do they give you? I introduced this to ensure the executable files we use as targets always get updated along with new module versions (even if the new version did not directly affect them). So if you remove this option, make will attempt re-building them every time.

Makefile
stweil: The `touch` command is only executed when there is also an update, so just using the current time would be adequate. Currently it takes the time of the last of all newer files. That is not necessarily the newest file.

Makefile
bertsky: Oh, right – I forgot about that!

Well, that's too bad. Then we'll have to use variables again. I'll add a commit when I'm done with your suggestions.

Makefile
bertsky: Yes, let's do that in an extra commit / PR which also re-orders the rules (core before kraken, tesseract near ocrd_tesserocr).

.gitignore
bertsky: But editing is pretty regular usage here, isn't it? And you'll see an entry for vim's `.swp` in nearly every OCR-D repo. Are you saying these should all be removed, too?

stweil: Ideally yes, practically no. They should never have been added, but I would only remove them when touching `.gitignore` anyway.

stweil: You already have shown two examples of such local files. We could also add `.bak`, files and directories from the different IDEs and much more and would still be incomplete. So simply don't start adding any of those.

Makefile
bertsky: > Currently it takes the time of the last of all newer files. That is not necessarily the newest file.

That's true. You removed sorting from #6 to get it working on Mac OS IIRC.

But that can only happen when the submodule was last updated without the above rule. And it does no harm either, because the directory is not as new as it could be, but still newer than before, so triggering updates for executables is ensured.

> so just using the current time would be adequate

I would have to agree if the first command was correct itself, but the picture is complicated by the fact that you can have empty status prefixes yet dirty status messages – `modified content` or `untracked content`. The latter could be ignored, but the former happens when submodules of the submodules are at a wrong version, and since we want recursive consistency, this should actually trigger the update as well. (This happened to me recently with `opencv-python`, which had different `multibuild`, `opencv` and `opencv_contrib`, and so I only could get it clean by doing the `submodule update` manually.)

So I think we actually should remove the first line (`status`) and try the second (`update`) unconditionally, but then really look at the time stamps again.

Or does anyone have a better idea?

Makefile
bertsky: But the problem remains that you have to checkout the modules before you can `make deps-ubuntu` in them.

bertsky: So maybe we should in fact enter the git submodule update rules when running `make deps-ubuntu`, but within the update rules, chowning the submodules to the owner of the toplevel (ocrd_all) directory?

Makefile
stweil: I had a similar idea, but only run the `touch` line unconditionally. Then each `make modules` would update the timestamp until no newer file exists.

Makefile
stweil: Or we abort with a message if submodules are missing and tell the user how to get the required submodules before running `make deps-ubuntu`.

Makefile
bertsky: > I had a similar idea, but only run the `touch` line unconditionally. Then each `make modules` would update the timestamp until no newer file exists.

Like I said:  The point in using a reference file for the time stamp is that we don't want to cause arbitrary updates (of the executables!), only ones backed by actual changed files.

bertsky: Okay, I just found `git submodule status --recursive`! This is enough to address the `modified content` problem (without giving up the first line; this is important because it does not waste network bandwidth).

So I committed that solution. But then again, I don't think we even need `--recursive` anymore. So I added another commit that makes the new behaviour non-recursive, but allows setting `GIT_MODULES=--recursive` to get the old behaviour back.

Regarding simplification of `touch -r` to just `touch`: it does not harm to do it this way either.

Makefile
bertsky: > Or we abort with a message if submodules are missing and tell the user how to get the required submodules before running `make deps-ubuntu`.

You want to encapsulate this, or it will be too complicated to use.

I added a solution using `chown --reference=$(CURDIR)` on the prerequisite submodules which might have been updated with sudo.

README.md
bertsky: You'd have to mention all the others, too. But this is illustrative, so mentioning libleptonica should be enough.

bertsky:
@kba, if you want I can add 2 commits that remove ocrd_kraken and ocrd_ocropy now. Oh, and please tag this repo with the topic `ocr-d`!

bertsky:
So, any more suggestions/objections, or can I merge?

bertsky:
> ~/src/github/OCR-D/ocrd_all/ocrd_cis$ git log b79e1c07b9b1ed695999a529b5e91ba1d9986f95
> fatal: bad object b79e1c07b9b1ed695999a529b5e91ba1d9986f95

Thanks @stweil! I was sloppy with the module updates (ocrd_cis uses/needs the `dev` branch). I have fixed this and force-pushed to not clutter the history.

stweil:
@bertsky, I rebased your commits and added some more commits. Like that, I think it is ready for merging. Please review my additions.

stweil:
`make install-tesseract` fails now on macOS because of many missing static libraries.

stweil:
`make install-tesseract` fails now on macOS because of many missing static libraries.

bertsky:
The last 2 changes are ok of course. (Strange that the autoconf manual is incorrect and you still need `--disable-shared` on macOS...)

bertsky:
> These changes break my uniform solution for both wget (for local build) and apt (for PPA install).

Fixed that now.

@stweil, your review status still says `1 change requested`?

stweil:
> @stweil, your review status still says 1 change requested?

That was because of the broken default help goal (which I fixed later in a commit without updating the status).

PR #25, opened 2019-12-14
convert multi-output recipes to pattern rules to avoid repetition

bertsky:
This converts rules that specify multiple targets with a recipe that builds them all at once – as most multi-executable modules have – into pattern rules, so `make` "knows" that all the targets are built together. Thus, the rule only fires once per module (instead of once per executable), and parallel building (`make -j`) becomes possible again. 

We could have waited for GNU make 4.2 for this, which finally added new syntax to support such rule semantics (via `:&` instead of `:`). But pattern rules can do this, too. They are just more difficult to read and write.

(I encapsulated this in a parameterized function `multirule` which takes the original list of executables and replaces the `$(BIN)/` part – which is _assumed_ to always be present – with `%`, so the rule becomes a pattern rule, but no other target is likely to match it. This is only necessary for the actual recipe, the dependencies still must be specified normally.)


stweil:
Thank you.

PR #28, opened 2019-12-20
Update core to 2.1.0, ocrd_olena to 1.0.0

kba:


stweil:
I suggest to make such updates for submodules without pull request, just push to OCR-D:master.

bertsky:
> I suggest to make such updates for submodules without pull request, just push to OCR-D:master.

My vote goes against that. Adding a new version can introduce incompatibilities, especially with the quick pace of various repos these days. Not everyone has a full perspective. There should be a possibility to say wait, that version does not fit (yet), or needs more along with it.

PR #30, opened 2019-12-20
docker image based on ocrd/core, no need to build core from repo

kba:
More of a question: If we derive the docker images from `ocrd/core`, should we install `core` from ocrd_all?


bertsky:
I would prefer to keep this independent. Docker builds come with a delay in delivery time. And we have an extra degree of freedom here: we can use a branch of core, or some older version that we don't know or don't have to Docker image ID of.

PR #38, opened 2020-01-08
fix f8af582: avoid shared build rule (only share recipe)

bertsky:
Fixes #36

PR #39, opened 2020-01-08
Don't add opencv-python to OCRD_MODULES by default

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,9 @@ CUSTOM_DEPS = wget python3-venv # add more packages for deps-ubuntu below (or mo
 # Default to all submodules, but allow overriding by user
 # (and treat the empty value as if it was unset)
 ifeq ($(strip $(OCRD_MODULES)),)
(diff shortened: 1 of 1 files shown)


kba:
Why do we need opencv-python as part of this at all?

stweil:
Ideally `make all` works on any recent Linux, not only x86_64, but also ARM or Power64 based hosts which are interesting for OCR because of their included GPU or high memory bandwidth.

Currently there exist no pre-compiled Python packages `opencv-python` or `opencv-python-headless` for such non x86_64 hosts. So having the source at hand helps for a manual installation.

kba:
> Currently there exist no pre-compiled Python packages opencv-python or opencv-python-headless for such non x86_64 hosts. So having the source at hand helps for a manual installation.

Just as I was asking, @wrznr complained that there are no debug builds for opencv-headless py36 on pypi. I retract my question :)

kba:
I messed up merging this locally into #38 so your commit doesn't appear there. Sorry. But its contents are part of #38.

PR #40, opened 2020-01-08
LABEL the image with git rev, build date, maintainer, microbadger bad…

kba:
Will add labels to the container with build date and git revision that can be inspected with `docker inspect` - might be useful for debugging. `MAINTAINER` is deprecated. Additional benefit is that this is compatible with microbadger which displays docker image info in a shields.io-like badge in the README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -473,6 +473,8 @@ docker-maximum docker-maximum-git: DOCKER_MODULES = $(OCRD_MODULES)
 # (maybe we should add --network=host here for smoother build-time?)
 docker%: Dockerfile $(DOCKER_MODULES)
 	docker build \
+	--build-arg VCS_REF=$$(git rev-parse --short HEAD) \
+	--build-arg BUILD_DATE=$$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
 	--build-arg OCRD_MODULES="$(DOCKER_MODULES)" \
 	--build-arg PIP_OPTIONS="$(PIP_OPTIONS)" \
 	-t $(DOCKER_TAG):$(or $(*:-%=%),latest) .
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -473,6 +473,8 @@ docker-maximum docker-maximum-git: DOCKER_MODULES = $(OCRD_MODULES)
 # (maybe we should add --network=host here for smoother build-time?)
 docker%: Dockerfile $(DOCKER_MODULES)
 	docker build \
+	--build-arg VCS_REF=$$(git rev-parse --short HEAD) \
+	--build-arg BUILD_DATE=$$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
```
```suggestion
	--build-arg BUILD_DATE=$$(date -u +"%Y-%m-%dT%H:%M:%S%Z") \
```

stweil: ... or `%z` for the timezone?

bertsky:
> Will add labels to the container with build date and git revision that can be inspected with `docker inspect` - might be useful for debugging

Looks great. Isn't that also something to at least recommend in docker.md for all modules? (Otherwise you never know which version to pull...)

Makefile
kba: No an actual `Z` as in `2020-01-08T17:37:07Z` (ISO 8601)

kba:
> Isn't that also something to at least recommend in docker.md for all modules? (Otherwise you never know which version to pull...)

Absolutely. I'll open an issue to remember.

Makefile
stweil: Sorry, you are right. Local date would be `date +"%Y-%m-%dT%H:%M:%S%z` in ISO 8601.

stweil:
Is this PR ready for merging?

kba:
> Is this PR ready for merging?

Yes, I cancelled the build yesterday so it didn't hog the queue.

PR #41, opened 2020-01-08
cor-asv-fst: fix build

bertsky:
That should suffice to at least install cor-asv-fst properly. (It still does not make it [usable](https://github.com/ASVLeipzig/cor-asv-fst/blob/a0bc5f9f5196050c4c53ca58d0588aa42aaa1422/ocrd_cor_asv_fst/wrapper/decode.py#L237) though... so let's keep it deactivated here.)


bertsky:
> Maybe we should add circle ci command to build this? That way it would not be in the docker images but still be tested to build.

Good idea. Will look into that (also for cor-asv-ann and for ocrd_keraslm, which at least has tests).

PR #43, opened 2020-01-10
circleci: run every docker push as separate step for clarity

kba:
should make build failures like https://circleci.com/gh/OCR-D/ocrd_all/40 easier to understand.

Not comitting to master until the next PR merge because no functional changes and no need to rebuild (tesseract update earlier fixed the build).


stweil:
I think this was also part of PR #44, so it can be closed now.

PR #55, opened 2020-01-25
Move tesseract/build to $(VIRTUAL_ENV)/build/tesseract

stweil:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 /venv
 local.mk
-**/build/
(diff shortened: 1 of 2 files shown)


stweil:
The old location had an inherent problem: `make install-tesseract` had always built a new tesseract, because a new tesseract executable also changed the timestamp of the tesseract directory which caused an update of the configure script which caused a new build [...].

stweil:
In a 2nd step a similar change should be done for the olena build. This requires also changes for `ocrd_olena`.

PR #58, opened 2020-02-04
update Tue Feb  4 11:04:11 CET 2020

kba:
Mostly interesting for @mikegeber's glyph markup and the continuous updates to tesseract and roberts worfkflow-configuration


bertsky:
regarding ocrd_calamari, maybe we should wait for https://github.com/OCR-D/ocrd_calamari/issues/30 or even https://github.com/OCR-D/ocrd_calamari/issues/31 to be resolved?

regarding CI failure, is it possible you accidentally used the `master` instead of the `dev` branch of ocrd_cis?

mikegerber:
> regarding ocrd_calamari, maybe we should wait for [OCR-D/ocrd_calamari#30](https://github.com/OCR-D/ocrd_calamari/issues/30) or even [OCR-D/ocrd_calamari#31](https://github.com/OCR-D/ocrd_calamari/issues/31) to be resolved?

I plan to release ocrd_calamari 0.0.4 this week which will include fixes for these (https://github.com/OCR-D/ocrd_calamari/milestone/1).

stweil:
> regarding CI failure, is it possible you accidentally used the master instead of the dev branch of ocrd_cis?

@kba, I think the answer is yes, therefore I removed that part from your PR.

> regarding ocrd_calamari, maybe we should wait [...]

Do we have to wait? Or can we merge the PR now and update ocrd_calamari later again?

kba:
> regarding CI failure, is it possible you accidentally used the master instead of the dev branch of ocrd_cis?

Yes, sorry I though to have swiched to amster for good. thx for spotting&fixing @bertsky @stweil 

> Do we have to wait? Or can we merge the PR now and update ocrd_calamari later again?

The fav-favoirite feature of per-glyph confidences merits an early merge IMHO.

stweil:
So I'll merge this as soon as CircleCI has finished successfully if nobody objects.

bertsky:
> Do we have to wait? Or can we merge the PR now and update ocrd_calamari later again?

I am simply afraid if we merge too early, people will suddenly see their workspace validation fail. Plus glyph alternatives likely take longer, so starting out with a default on the word level would be better.

stweil:
@bertsky, so what do you suggest? Remove the update of ocrd_calamari from the PR before merging?

bertsky:
> so what do you suggest? Remove the update of ocrd_calamari from the PR before merging?

Yes, if this PR cannot wait for further ocrd_calamari improvements, I recommend keeping at the previous version for now. (Except if someone has already run the new version on a realistic dataset and can falsify the validation concerns.)

kba:
> falsify the validation concerns.

But the https://github.com/tesseract-ocr/tesseract/pull/2871 was the blocking PR. So users upgrading to the latest ans greatest tesseract will but confused the issues. Let's wait for @mikegerber finishing up https://github.com/OCR-D/ocrd_calamari/milestone/1 and then release ok @stweil @bertsky ?

mikegerber:
* The validating word segmentation is already in ocrd_calamari git master, I just want to add a test for this.
* Glyph level segmentation in ocrd_calamari does not add any significant computation because we're just using already computed results, it just makes the PAGE output bigger (that's why it's sensible to add an option to not include it)

bertsky:
> * The validating word segmentation is already in ocrd_calamari git master, I just want to add a test for this.
> 
> * Glyph level segmentation in ocrd_calamari does not add any significant computation because we're just using already computed results, it just makes the PAGE output bigger (that's why it's sensible to add an option to not include it)

@mikegerber thanks for clarifying!

Then I am in favour of merging as is and include the rest in another PR as proposed by @stweil.

stweil:
Thanks to everybody!

PR #62, opened 2020-02-13
Fix ownership for files in .git after `sudo make deps-ubuntu`

stweil:


stweil:
This fixes a permission problem caused by files and subdirectories under `.git` being owned by root after `sudo make deps-ubuntu`. I got this today when I tried to follow the documentation (obviously for the 1st time).

PR #66, opened 2020-02-20
module updates

bertsky:
among the changes:

- core 2.4.2
- ocrd_tesserocr (fixes in recognize and segment-region)
- ocrd_segment (extract with mimetype option, extract lines from tables)
- bash env and version checks
- assets updates


stweil:
@kba, can we merge this PR?

PR #71, opened 2020-02-27
⬆ Update sbb_textline_detector (to fix multi threading issue)

mikegerber:
See also https://github.com/qurator-spk/sbb_textline_detection/issues/27.


mikegerber:
(This is relatively urgent as it breaks the functionality predictably for some machines.)

Issue #69, opened 2020-02-25
Resolve the conflicting requirements issues


mikegerber:
In #68 @bertsky :
> But the real problem is that TF2 dependencies are lurking everywhere, so we will very soon have the unacceptable state that no catch-all venv (satisfying both TF1 and TF2 modules) is possible anymore. By then, a new solution needs to be in place, which (at least partially) isolates venvs from each other again.

mikegerber:
I think OCR-D has to decide if you want a "have it all" environment. Then either separate venvs per processor or some kind of Docker setup could be the solution.

I'd go for a Docker solution because that also solves dependency issues outside of the Python world. And it is actually possible to do without over-engineering too hard. I.e. I use a `run` one-liner that just calls a container with the right UID and working directory in place.

mikegerber:
```
docker run --rm -t  --user `id -u`:`id -g` --mount type=bind,src="$(pwd)",target=/data ocrd_example_processor
```

bertsky:
> I think OCR-D has to decide if you want a "have it all" environment. Then either separate venvs per processor or some kind of Docker setup could be the solution.

As long as we don't have REST integration, all-in-one is all we have. Isolation within that approach will always be up to demand and interdependencies. Whether we then delegate to local venvs or thin Docker containers is indeed another freedom of choice (which we ideally should also make a user choice here).

> I'd go for a Docker solution because that also solves dependency issues outside of the Python world. 

Yes, but it may also increase space and time cost unnecessarily under circumstances (depending on which modules are enabled and what platform the host is). So I'd really like that to be for the user to decide ultimately.

> And it is actually possible to do without over-engineering too hard. I.e. I use a `run` one-liner that just calls a container with the right UID and working directory in place.

Yes of course, once we wrap the processor CLIs in another shell script layer, we can make that local venv or Docker run again. And for the latter we only need to ensure we pass on all the arguments. (Perhaps we could even avoid spinning up a new container with each invocation by using `--attach` somehow.)

The only thing that troubles me with delegating to thin module Docker containers is that we more or less surrender version control. It's really difficult to do that with Docker images solely based on digest numbers. But we could of course petition module providers to use the `docker tag` mechanism in a certain way.

kba:
[...]

PR #76, opened 2020-04-08
update submodules (avoiding TF 1 vs 2 conflict)

bertsky:
This brings updates to ocrd_keraslm and cor-asv-ann which will completely avoid conflict with TF2 ocrd_anybaseocr – as a temporary solution to #69.


stweil:
@bertsky, recent Linux distributions with Python 3.8 might all fail with `make all`:

    ERROR: No matching distribution found for tensorflow-gpu==1.15.* (from ocrd-cor-asv-ann==0.1.2)

I use Debian Bullseye (=testing) with Python 3.8.2.

bertsky:
> all recent Linux distributions with Python 3.8 might fail with `make all`:
> 
> ```
> ERROR: No matching distribution found for tensorflow-gpu==1.15.* (from ocrd-cor-asv-ann==0.1.2)
> ```
> 
> I use Debian Bullseye (=testing) with Python 3.8.2.

Gosh, these TF people drive me crazy. First they just redefine their naming scheme **twice** for the same versions (`tensorflow` vs `tensorflow-gpu` on pre-2), then they don't even provide up-to-date builds anymore.

But you have to concede, they say so directly on their website:

> TensorFlow is tested and supported on the following 64-bit systems:
> 
> Python 3.5–3.7

And the [way they say it](https://github.com/tensorflow/tensorflow/issues/33838#issuecomment-566342226), it sounds they *will* provide 3.8 support...

stweil:
We have several ways how to handle this:
- Do nothing. It will be solved as soon as there is a TF binary for Python 3.8.
- Do nothing. It only affects people who use the latest distributions. Those people should know how to handle problems.
- Add documentation to the README and hope that people read it.
- Enhance the Makefile to check the Python version. Fail with an error message if it is 3.8, or look whether there is also a Python 3.6 or 3.7 available and use that.
- ...

kba:
I'm for a combination of "Do nothing" and a warning for python 3.8 users in the README. It's imperative for TF to provide up-to-date versions, supporting 3.8 must be high on their priorities and users that are on bleeding edge distros like Debian 11 or Ubuntu 20.04 can be expected to check the documentation or even install an older python in case of problems.

bertsky:
On the other hand, we don't even look for Python 3.8 support in core ourselves yet (CircleCI and Travis use 3.5-3.7). That's not "3.5 or higher"...

PR #77, opened 2020-04-08
CircleCI: run docker pushs in parallel but with steady stdout

bertsky:
Since we don't run `deploy` on PR branches, do we have a chance to test whether this actually works _before_ merging?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,12 +22,22 @@ jobs:
       - run:
           name: Login to Docker Hub
           command: echo "$DOCKERHUB_PASS" | docker login --username "$DOCKERHUB_USERNAME" --password-stdin
-      - run: docker push ocrd/all:minimum
-      - run: docker push ocrd/all:minimum-git
-      - run: docker push ocrd/all:medium
-      - run: docker push ocrd/all:medium-git
-      - run: docker push ocrd/all:maximum
-      - run: docker push ocrd/all:maximum-git
+      - run:
+          name: Push images to Docker Hub in parallel
+          command: |
+            docker push ocrd/all:minimum &
+            jobs=($!)
+            docker push ocrd/all:minimum-git &
+            jobs+=($!)
+            docker push ocrd/all:medium &
+            jobs+=($!)
+            docker push ocrd/all:medium-git &
+            jobs+=($!)
+            docker push ocrd/all:maximum &
+            jobs+=($!)
+            docker push ocrd/all:maximum-git &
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -22,12 +22,22 @@ jobs:
...
-      - run: docker push ocrd/all:medium-git
-      - run: docker push ocrd/all:maximum
-      - run: docker push ocrd/all:maximum-git
+      - run:
+          name: Push images to Docker Hub in parallel
+          command: |
+            docker push ocrd/all:minimum &
+            jobs=($!)
+            docker push ocrd/all:minimum-git &
+            jobs+=($!)
```
stweil, line 31: I tried the following sequence in a shell, and it did not work as expected:

    sleep 100 &
    jobs=($!)
    sleep 100 &
    jobs+=($!)

stweil: Sorry, I was mistaken. It works.

.circleci/config.yml
```diff
@@ -22,12 +22,22 @@ jobs:
...
+            docker push ocrd/all:medium &
+            jobs+=($!)
+            docker push ocrd/all:medium-git &
+            jobs+=($!)
+            docker push ocrd/all:maximum &
+            jobs+=($!)
+            docker push ocrd/all:maximum-git &
+            jobs+=($!)
+            while sleep 60; ps -p "${jobs[*]}"; do :; done
```
stweil, line 40:
```suggestion
            wait ${jobs[*]}```

bertsky: No, doing some command which actually _prints_ something on stdout was the whole point. See discussion on Gitter Lobby.

kba: The problem this PR adresses isn't parallelization but that CircleCI fails workflows if a job within it runs for more than 10 minutes without output to STDOUT. The parallelization is just a bonus.

kba:
> do we have a chance to test whether this actually works before merging?

Not without changing the configuration file. Let's wait until https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/111/workflows/3a12fd02-de90-477c-a8ad-17e3c63eb40c/jobs/117 finishes and merge then. If it doesn't work we can fix it later.

.circleci/config.yml
stweil: Would GitHub actions be a better alternative? See for example https://github.com/UB-Mannheim/tesseract/actions.

PR #85, opened 2020-04-30
update ocrd_olena submodule

bertsky:
@kba I did not include the most recent master of core, because OCR-D/core#474 is not merged yet and so AFAIK 2.5.1 is the last correct version.


kba:
OCR-D/core#474 is now merged and released as 2.5.3.

Issue #35, opened 2020-01-06
README: Update "conflicting requirements"


kba:
e.g. the pillow one should be gone.

ocrd_calamari is pushing for tf2 support.

When scrutinizing ocrd_anybaseocr this week, keep in mind to update if applicable.

bertsky:
> ocrd_calamari is pushing for tf2 support.
> 
> When scrutinizing ocrd_anybaseocr this week, keep in mind to update if applicable.

That makes it urgent to isolate modules' venv from each other (but still integrate all executables into one PATH). I still have something half-ready, somewhere...

stweil:
Can this be closed?

bertsky:
Definitely!

PR #86, opened 2020-04-30
Update min/med/max table, ht @bertsky, OCR-D/ocrd-website#92

kba:


bertsky:
now also fixes #35

README.md
```diff
@@ -355,16 +355,21 @@ Modules may require mutually exclusive sets of dependent packages.
 `pip` does not even stop or resolve conflicts – it merely warns!
 
 - Tensorflow:
-   * `tensorflow-gpu==1.14.0` (required by ocrd_calamari and ocrd_anybaseocr)
-   * `tensorflow` (required by cor-asv-ann and ocrd_keraslm)
+   * version 2 (required by ocrd_anybaseocr for dewarping and page segmentation)
+   * version 1 (required by cor-asv-ann, ocrd_keraslm, ocrd_calamari and sbb_textline_detector)
```
VolkerHartmann, line 359: tf 2 is also required by ocrc_pc_segmentation via ocr4all_pixel_classifier

README.md
```diff
@@ -355,16 +355,21 @@ Modules may require mutually exclusive sets of dependent packages.
...
+   
+   The temporary solution is to require different package names:
+   - `tensorflow>=2`
+   - `tensorflow-gpu==1.15.*`
    
-   Both can be installed in parallel in different versions, but may depend on a mutually exclusive set of `tensorboard` and `tensorflow_estimator`.
+   Both can be installed in parallel in different versions, but may depend on a mutually exclusive set of `tensorboard` and `tensorflow_estimator`. (However, these latter conflicts are only cosmetic, as they are only needed for development.)
```
VolkerHartmann, line 366: I created two independent venv (tf 1.15 & tf 2.0) and linked the binaries to the 'main' venv via a script:
#!/bin/bash
# Get absolute path of the both virtual environments.
SOURCE=$(readlink -f $1)
TARGET=$(readlink -f $2)
# Test for virtual environment
if [ ! -f "$SOURCE/bin/activate" ]; then
    echo "$SOURCE is not a virtual environment!"
    exit 1
fi
if [ ! -f "$TARGET/bin/activate" ]; then
    echo "$TARGET is not a virtual environment!"
    exit 1
fi
echo link $SOURCE to $TARGET
for processors in $SOURCE/bin/ocrd-*
do 
  filename=$(basename -- "$processors")
  echo $filename
  if grep -q import $processors; then
    TARGET_FILE=$TARGET/bin/$filename
    # Write script to target virtual environment.
    echo "#!/bin/bash"> $TARGET_FILE 
    echo " " >> $TARGET_FILE
    echo "# Get command" >> $TARGET_FILE
    echo "command=\$(basename -- \"\$0\")" >> $TARGET_FILE 
    echo " " >> $TARGET_FILE
    echo ">&2 echo Switch virtual environment from $TARGET to $SOURCE" >> $TARGET_FILE 
    echo source $SOURCE/bin/activate  >> $TARGET_FILE 
    echo "\$command \"\$@\""  >> $TARGET_FILE 
    chmod 755 $TARGET_FILE 
   else
    echo "Skip - Not a valid virtual environment script."
  fi
done

bertsky: Sure, we have been planning to go module-isolated venvs for some time. (But you wouldn't do it that way. You would create a wrapper shell script as entry point to the respective venv for each tool you install into the overall target prefix. It's then also possible to have an arbitrary mix of overall venv, subset venv or even run thin Docker images.)

But there are still some conceptual problems with that:
1. completely isolating venvs would also bloat the filesystem size a lot, because most module do in fact have many similar, large dependencies
2. where modules depend _on each other_ (either in Python world or via libraries), they must share their venv; e.g. ocrd_tesserocr → tesserocr → tesseract
3. how do you approach updating/synchronizing Docker images; you cannot just `docker pull` the most recent version, but it's hard to keep a book of image IDs like revisions in .gitmodules

README.md
bertsky: Are you sure about that?

`ocrd_pc_segmentation/requirements.txt` only has `ocr4all_pixel_classifier` from PyPI, which really is https://github.com/ocr-d-modul-2-segmentierung/page-segmentation. But that only needs:
```
Pillow
numpy
scipy
scikit-image
tqdm
opencv-python-headless
dataclasses
efficientnet
lxml
```

And of those, only `efficientnet` needs TF, although it does not correctly advertise that in its setup.py. But the [README says](https://github.com/qubvel/efficientnet#requirements) it wants TF1.

crater2150: Sorry for the late reply.
ocr4all_pixel_classifier uses Tensorflow 2, but it is not declared via `requirements.txt` but as an `extra_requires` with either `tf_cpu` or `tf_gpu` flag, because of the separate packages for CPU and GPU. This could be changed by using tensorflow >= 2.1.0, which no longer has that distinction and was already tested, but I rolled back because of this bug: ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation#7

bertsky: > This could be changed by using tensorflow >= 2.1.0, which no longer has that distinction and was already tested, but I rolled back because of this bug: [ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation#7](https://github.com/ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation/issues/7)

I see. This is unfortunate. Then we have a stalemate situation: We need the "`tensorflow-gpu` for TF1, `tensorflow` for TF2" scheme to avoid the conflicting requirements in the ocrd_all venv, but your module is not quite TF2 enough to be able to rely on the "`tensorflow` for both CPU and GPU environments" scheme. Am I correct?

Generally IMHO you should try to change the `extras_require` to a regular `install_requires` because otherwise `pip install` won't pick up the dependency (unless the user knows about the extras and uses the Makefile accordingly). But for now I don't know exactly what to recommend.

Maybe we should be pragmatic and – condering that most users will run the Docker version in which we still cannot offer GPU support anyway – set the requirement to `tensorflow>=2.0.0,<2.1.0` unconditionally (and later when you support TF2.1 change to `tensorflow>=2.0`).

crater2150: > but your module is not quite TF2 enough to be able to rely on the "tensorflow for both CPU and GPU environments" scheme. Am I correct?
The module runs fine with TF2.1, I haven't been able to reproduce the bug with a current distribution. If @kba can confirm that it is a problem only with Debian 8, I can just change the requirement to `tensorflow>=2.0` directly without problems.

kba: I haven't had a problem in Ubuntu 18.04, so https://github.com/ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation/issues/7 is likely not an issue with modern distros.

bertsky:
Is there any reason for this to hang loose, or can we pick the fruit?

kba:
the pr can be merged but the discussion on venv isolation for conflicting versions is still unresolved, maybe track in an issue?

stweil:
What about Volker's change request for the README?

bertsky:
> the pr can be merged but the discussion on venv isolation for conflicting versions is still unresolved, maybe track in an issue?

This is already tracked by #69 IMO. But I can copy the comment there...

bertsky:
> What about Volker's change request for the README?

It's not a change request / suggestion.

stweil:
@bertsky, I was talking about https://github.com/OCR-D/ocrd_all/pull/86#pullrequestreview-406374041.

bertsky:
>  I was talking about [#86 (review)](https://github.com/OCR-D/ocrd_all/pull/86#pullrequestreview-406374041).

@stweil sorry, I don't understand. There's the comment relating to #69 and another that's been resolved.

bertsky:
> and another that's been resolved.

Oh, maybe that was premature! You mean that `ocrd_pc_segmentation` should be mentioned in the list of TF2 dependents, don't you?

bertsky:
> You mean that `ocrd_pc_segmentation` should be mentioned in the list of TF2 dependents, don't you?

Done.

stweil:
Thanks.

PR #88, opened 2020-05-03
Makefile: export PYTHON variable...

bertsky:
- don't overwrite PYTHON variable from caller's env, if any
- (re-)export PYTHON variable to subprocesses
  (also affects builds in Docker)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,11 @@
 -include local.mk
 
 # Python version (python3 required).
-PYTHON := python3
+PYTHON ?= python3
 # Python packaging
 PIP ?= pip3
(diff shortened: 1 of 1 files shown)


bertsky:
@stweil I have a faint memory we somehow discussed whether or not to use `export` in this manner, but cannot find it anymore.

Rationale here is that we forgot to export `PYTHON` (which is picked up by some modules' Makefiles) to the shell in all our recipes. So if the caller did not set it herself (as is the case in our CircleCI Docker build rules), then it will not get defined.

See OCR-D/ocrd_olena#51

PR #91, opened 2020-05-09
Support user provided PKG_CONFIG_PATH

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,11 @@ define WGET
 $(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
 endef
 
+ifeq ($(PKG_CONFIG_PATH),)
 PKG_CONFIG_PATH := $(VIRTUAL_ENV)/lib/pkgconfig
+else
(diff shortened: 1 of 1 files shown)


stweil:
macOS with Homebrew requires a special `PKG_CONFIG_PATH` setting.

Makefile
```diff
@@ -27,7 +27,11 @@ define WGET
 $(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
 endef
 
+ifeq ($(PKG_CONFIG_PATH),)
 PKG_CONFIG_PATH := $(VIRTUAL_ENV)/lib/pkgconfig
+else
+PKG_CONFIG_PATH := $(VIRTUAL_ENV)/lib/pkgconfig:$(PKG_CONFIG_PATH)
+endif
```
bertsky, line 34: Do you actually need both, the user-defined path _and_ the one under venv/lib/pkgconfig? If not, you could just change it to:
```suggestion
PKG_CONFIG_PATH ?= $(VIRTUAL_ENV)/lib/pkgconfig
```

stweil: The Tesseract build puts its `tesseract.pc` in `venv/lib/pkgconfig`, and that is used by `tesserocr`. So that part is not optional. I'm afraid it would not work to ask users to handle that correctly in their own `PKG_CONFIG_PATH`.

PR #94, opened 2020-05-12
Add build rules for Tesseract training tools

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ Targets:
 	modules: download all submodules to the managed revision
 	all: installs all executables of all modules
 	install-tesseract: download, build and install Tesseract (with required models)
+	install-tesseract-training: build and install Tesseract training tools
 	fix-pip: try to repair conflicting requirements
 	clean: removes the virtual environment directory, and clean-*
 	clean-tesseract: remove the build directory for tesseract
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -454,6 +454,37 @@ $(BIN)/tesseract: tesseract/configure
...
+TESSTRAIN_EXECUTABLES += $(BIN)/set_unicharset_properties
+TESSTRAIN_EXECUTABLES += $(BIN)/shapeclustering
+TESSTRAIN_EXECUTABLES += $(BIN)/text2image
+TESSTRAIN_EXECUTABLES += $(BIN)/unicharset_extractor
+TESSTRAIN_EXECUTABLES += $(BIN)/wordlist2dawg
+
+.PHONY: install-tesseract-training
+install-tesseract-training: $(TESSTRAIN_EXECUTABLES)
+
+$(TESSTRAIN_EXECUTABLES): tesseract/configure
```
bertsky, line 479: This will ensure that the rule does not fire up N times for N tools, but make will treat the recipe as multi-target (without waiting for make 4.3's proper "grouped target" support):

```suggestion
$(call multirule,$(TESSTRAIN_EXECUTABLES)): tesseract/configure
```

stweil:
Thanks for the review. I added a commit with your suggestions (I slightly modified the help text).

PR #100, opened 2020-05-25
Update ocrd-pixelclassifier-segmentation -> 0.1.5

kba:
Model wasn't bundled because of https://github.com/ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation/issues/10#issuecomment-630676588

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,7 +315,7 @@ ifneq ($(findstring ocrd_pc_segmentation, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_PC_SEGMENTATION)
 OCRD_PC_SEGMENTATION := $(BIN)/ocrd-pc-segmentation
 $(OCRD_PC_SEGMENTATION): ocrd_pc_segmentation
-	$(pip_install)
+	. $(ACTIVATE_VENV) && cd $< && $(MAKE) install
 endif
 
 ifneq ($(findstring ocrd_anybaseocr, $(OCRD_MODULES)),)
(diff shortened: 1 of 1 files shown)


bertsky:
Don't we also need to change https://github.com/OCR-D/ocrd_all/blob/47df98ad26f979a89e70008512417cf644bb7473/Makefile#L318 so that `make install` gets used, not just `pip install`? (Background: the dependency `ocr4all-pixel-classifier` along with its GPU/CPU TF distinction is _not_ in requirements.txt / setup.py's `install_requires`, only in that makefile recipe. I wonder how this could have ever worked in the Docker builds...)

kba:
@crater2150

bertsky:
CI failure is unrelated. Looks like a Sourceforge mirror failed during ocr-fileformat installation.

Makefile
```diff
@@ -315,7 +315,7 @@ ifneq ($(findstring ocrd_pc_segmentation, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_PC_SEGMENTATION)
 OCRD_PC_SEGMENTATION := $(BIN)/ocrd-pc-segmentation
 $(OCRD_PC_SEGMENTATION): ocrd_pc_segmentation
-	$(pip_install)
+	. $(ACTIVATE_VENV) && cd $< && $(MAKE) install
```
bertsky, line 318: How about best of both worlds?

```suggestion
	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
	$(pip_install)
```

That way we get all dependencies as needed, but the package itself will be installed such that timestamps are guaranteed to be updated (and in `-e` mode if requested).

PR #103, opened 2020-05-30
make install-models to install commonly used models for tesseract, ocropy, calamari

kba:
Adds targets `install-models{,-tesseract,-ocropy,-calamari}` to install recognition models by @stweil, @mikegerber, @jze, @chreul to sensible locations in `$(VIRTUAL_ENV)/share`.

The list of installed models is far from exhaustive and can/should be easily amended. 

This is a stopgap solution until we have a proper model repository. I had to re-download models over the last two weeks for different environments and I thought I might as well create a script and `ocrd_all` seemed like a good place.

I tried to use GitHub locations where possible. If potential traffic spikes from adding model download here are of concern, I can set up a mirror.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -418,7 +418,8 @@ CUSTOM_DEPS += g++ make automake libleptonica-dev
 # but since we are building statically, we need more (static) libs at build time
 CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
 
-TESSDATA := $(VIRTUAL_ENV)/share/tessdata
+TESSDATA_PREFIX ?= $(VIRTUAL_ENV)/share
+TESSDATA := $(TESSDATA_PREFIX)/tessdata
 TESSDATA_URL := https://github.com/tesseract-ocr/tessdata_fast
 TESSERACT_TRAINEDDATA = $(ALL_TESSERACT_MODELS:%=$(TESSDATA)/%.traineddata)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -423,6 +424,53 @@ TESSERACT_TRAINEDDATA = $(ALL_TESSERACT_MODELS:%=$(TESSDATA)/%.traineddata)
 
 stripdir = $(patsubst %/,%,$(dir $(1)))
 
+# {{{ Install commonly used models
+.PHONY: install-tesseract
```
stweil, line 428:
```suggestion
.PHONY: install-models
```

Makefile
```diff
@@ -423,6 +424,53 @@ TESSERACT_TRAINEDDATA = $(ALL_TESSERACT_MODELS:%=$(TESSDATA)/%.traineddata)
...
+install-models: \
+	install-models-tesseract \
+	install-models-ocropus \
+	install-models-calamari \
+
+UB_MANNHEIM_BACKUP_URL = https://ub-backup.bib.uni-mannheim.de/~stweil/ocrd-train/data/Fraktur_5000000
+.PHONY: install-models-tesseract
+install-models-tesseract: \
+	$(TESSDATA)/Fraktur_50000000.334_450937-fast.traineddata \
```
stweil, line 437: I personally prefer `$(TESSDATA)/fast/Fraktur_50000000.334_450937.traineddata`, ... and using `-l fast/Fraktur_50000000.334_450937` (or a parameter `fast/Fraktur_50000000.334_450937` for `ocrd-tesserocr`) at run time.

stweil:
Did I already mention that I don't like environment variables (with only a handful of exceptions like `PATH`)? I nearly never use `TESSDATA_PREFIX`. It is not needed when Tesseract was build for a virtual environment, and I would not propagate its use.

kba:
I removed the instructions on `TESSDATA_PREFIX`, they [were not working for ocrd_tesserocr anyways](https://github.com/OCR-D/ocrd_tesserocr/issues/129).

mikegerber:
> I tried to use GitHub locations where possible. If potential traffic spikes from adding model download here are of concern, I can set up a mirror.

In my experience, the GitHub downloads are painfully slow at times, you may want to consider mirroring them. (I mirored tessdata_best for that reason here: https://qurator-data.de/mirror/github.com/tesseract-ocr/tessdata_best/archive/)

stweil:
Regarding Tesseract models, my idea is that in the future models will be selected by their URL, downloaded on demand and cached locally.

kba:
> downloaded on demand and cached locally

That is also what we have in mind for processors in general, hence the `cacheable` and `content-type` attributes to mark parameters as (cacheable) remote files.

As I said before, this is merely a stopgap solution to let users benefit from the enhanced models out there now, while model repository infrastructure and processors improve.

kba:
Since we're already describing the mechanism [in our documentation](https://ocr-d.de/en/models), I'll merge. Better solution work bundling data with processors is upcoming.

mikegerber:
>     2. How about adding (some of) the Calamari models [here](https://github.com/Calamari-OCR/calamari_models)? 

(Late to the party.) I have never gotten any useful result from those models, but that may have been my mistake. → should be tested before integration.

Issue #101, opened 2020-05-26
Basic Releasemanagement


M3ssman:
As an user of OCR-D-Tools it is the easiest approach to run the Tools the containerized way.

For a production environment release builds of ocrd-all-containers are a key requirement for stable workflows. 

By now, the situation is as follows:
If an OCR-D-Container is pulled one week later on a different system, it must be possible to pull exactly the same version than it was the time before on the first system. Without tagging the container images, this can't be guaranteed, since they implicit tagged as `latest`.

Any common CI/CD-Plattform has the ability to run post-commit actions, so has CircleCI.

M3ssman:
The current container naming doesn't comply to common image naming conventions, as the part after the double colon is usually used for the tag:

`ocrd/all:maximum` => `vendor/product:tag` => `ocrd/maximum:2020.05.26`

M3ssman:
Another point that makes me wonder: Why 6 image flavors? 
Having a small one and one really maximum should do the job. 
Please keep in mind, that more products blow continuous deliverance and need additional documentation, too.

stweil:
> As a user of OCR-D-Tools it is the easiest approach to run the Tools the containerized way.

That depends. If you run OCR-D on a limited number of hosts, building it locally might be the better choice. At least I nearly never use containers for OCR-D. That allows easy local tagging of the installed release and parallel installations of different releases using virtual environments. For some hosts (ARM, PowerPC) there exist no containers at all.

bertsky:
> Another point that makes me wonder: Why 6 image flavors?
> Having a small one and one really maximum should do the job.

I must object. We have 3x2 flavours of pre-built images precisely because different users have different needs. Every module pre-selection (mini/medi/maxi-mum) has a different trade-off between costs (disk size and network bandwidth) and benefits (processor and workflow options available). And there's an orthogonal distinction between binary-only ("wait for next update") and source+binary ("autonomous mini-update") delivery. This flexibility seems natural for a complex federated system like OCR-D.

> Please keep in mind, that more products blow continuous deliverance and need additional documentation, too.

All this has been solved for automatic continous integration and deployment long ago. Even documentation is in a respectable state nowadays. (If you think otherwise, please don't hesitate to open issues specifically!)

> If an OCR-D-Container is pulled one week later on a different system, it must be possible to pull exactly the same version than it was the time before on the first system.

[...]

PR #104, opened 2020-05-31
Update ocrd_olena, tesseract, ocrd_cis

kba:
Update:

  * `ocrd_olena`: e000f24 (1.1.7)
  * `tesseract`: 62eae84
  * `ocrd_cis`: a6a2ecd
  * `ocrd_tesserocr`: 9d2253e (0.8.3 +1)
  * `sbb_textline_detector`: 3593506
  * `workflow-configuration`: 403ef75

Added:

  * `CHANGELOG.md` to track changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# CHANGELOG
+
+## 2020-05-31
+
+Update:
+
+  * `ocrd_olena`: e000f24 (1.1.7)
+  * `tesseract`: 62eae84
+  * `ocrd_cis`: a6a2ecd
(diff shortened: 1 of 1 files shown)


kba:
> I think we should discuss the content of CHANGELOG.md 

Totally, this is just a proposal and the format is not that useful (I was mostly curious whether GitHub would be able to identify the short commit strings to the commits in the submodule).

I'm also well aware of #101 and thinking about a compromise. In the meantime, starting with a CHANGELOG seemed constructive but is not meant to preempt a proper solution to #101.

PR #110, opened 2020-06-14
:arrow_up: 2020-06-14

kba:
Fixes in OCR-D/core bashlib and ocrd_fileformat.


stweil:
>Fixes in OCR-D/core bashlib and ocrd_fileformat.

... and Tesseract. That one was nagging since a couple of years.

PR #113, opened 2020-06-17
update 2020-06-17.1

kba:
Ideally we could also include a new version of ocrd_pagetopdf if https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/13 is merged, so all bashlib-based processors would be functional again.


stweil:
@kba, I updated `ocrd_pagetopdf` now. Do you want to update your PR to include it as well?

Issue #68, opened 2020-02-25
Outdated info about conflicting requirements


mikegerber:
1. https://github.com/OCR-D/ocrd_all#conflicting-requirements states that ocrd_calamari would depend on tensorflow-gpu 1.14.x, but it depends on 1.15.2 since recently.

2. There is also still some solvable(!) problem/confusion about the different TensorFlow flavours. For tensorflow 1.15.*, one can simply depend on `tensorflow-gpu == 1.15.*` *for CPU **and** GPU* support. I am not aware of any issues using `tensorflow-gpu`'s CPU fallback on CPU, I use it every day. (There *was* some source of additional confusion because TF changed their recommendation for 1.15 only.)

2b. I just recently discovered that one can depend on an approximate version, e.g. `tensorflow-gpu ~= 1.15.2` or `tensorflow == 1.15.*`

TL&DR: My recommendation would be that our TF1 projects just use `tensorflow-gpu == 1.15.*` *for CPU and GPU support* and be done with this problem.

bertsky:
>     1. https://github.com/OCR-D/ocrd_all#conflicting-requirements states that

Yes, that section needs to be updated (cf. #35). But the real problem is that TF2 dependencies are lurking everywhere, so we will _very soon_ have the unacceptable state that no catch-all venv (satisfying both TF1 and TF2 modules) is possible anymore. By then, a new solution needs to be in place, which (at least partially) isolates venvs from each other again.

>     2\. For tensorflow 1.15.*, one can simply depend on `tensorflow-gpu == 1.15.*` _for CPU **and** GPU_ support. I am not aware of any issues using `tensorflow-gpu`'s CPU fallback on CPU

But isn't that equally true for using `tensorflow == 1.15.*`? It is the variant with a `-gpu` suffix that is going to be dropped eventually IIUC.

mikegerber:
> > 2. For tensorflow 1.15.*, one can simply depend on `tensorflow-gpu == 1.15.*` _for CPU **and** GPU_ support. I am not aware of any issues using `tensorflow-gpu`'s CPU fallback on CPU
> >
> But isn't that equally true for using `tensorflow == 1.15.*`? It is the variant with a `-gpu` suffix that is going to be dropped eventually IIUC.

Nah, they *had* recommended `tensorflow-gpu` for TF2 CPU+GPU but changed it again to just `tensorflow` 🤣 So if `tensorflow == 1.15.*` has GPU support I am happy with that convention, too.

[...]

PR #118, opened 2020-06-26
group some modules into isolated venvs…

bertsky:
- for modules with certain sets of dependencies,
  delegate to another (isolated) virtual env and
  place a shell script in the top-level venv
- update some executable names
- remove `fix-pip`
- update documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,9 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
 
-BIN := $(VIRTUAL_ENV)/bin
-SHARE := $(VIRTUAL_ENV)/share
-ACTIVATE_VENV := $(VIRTUAL_ENV)/bin/activate
+BIN = $(VIRTUAL_ENV)/bin
+SHARE = $(VIRTUAL_ENV)/share
+ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
 define WGET
 $(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
@@ -70,7 +70,6 @@ Targets:
 	all: installs all executables of all modules
 	install-tesseract: download, build and install Tesseract (with required models)
 	install-tesseract-training: build and install Tesseract training tools
-	fix-pip: try to repair conflicting requirements
 	clean: removes the virtual environment directory, and clean-*
 	clean-tesseract: remove the build directory for tesseract
 	clean-olena: remove the build directory for ocrd_olena
@@ -189,7 +188,13 @@ OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-train
 OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-eval
 OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-repl
 $(call multirule,$(OCRD_COR_ASV_ANN)): cor-asv-ann
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_COR_ASV_ANN))
+	$(call delegate_venv,$(OCRD_COR_ASV_ANN))
+$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring cor-asv-fst, $(OCRD_MODULES)),)
@@ -198,8 +203,14 @@ OCRD_EXECUTABLES += $(OCRD_COR_ASV_FST)
 OCRD_COR_ASV_FST := $(BIN)/ocrd-cor-asv-fst-process
 OCRD_COR_ASV_FST += $(BIN)/cor-asv-fst-train
 $(call multirule,$(OCRD_COR_ASV_FST)): cor-asv-fst
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_COR_ASV_FST))
+	$(call delegate_venv,$(OCRD_COR_ASV_FST))
+$(OCRD_COR_ASV_FST): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(MAKE) -C $< deps
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_keraslm, $(OCRD_MODULES)),)
@@ -207,7 +218,13 @@ OCRD_EXECUTABLES += $(OCRD_KERASLM)
 OCRD_KERASLM := $(BIN)/ocrd-keraslm-rate
 OCRD_KERASLM += $(BIN)/keraslm-rate
 $(call multirule,$(OCRD_KERASLM)): ocrd_keraslm
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_KERASLM))
+	$(call delegate_venv,$(OCRD_KERASLM))
+$(OCRD_KERASLM): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_im6convert, $(OCRD_MODULES)),)
@@ -258,11 +275,21 @@ endif
 ifneq ($(findstring ocrd_segment, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_SEGMENT)
 OCRD_SEGMENT := $(BIN)/ocrd-segment-evaluate
+OCRD_SEGMENT += $(BIN)/ocrd-segment-from-masks
+OCRD_SEGMENT += $(BIN)/ocrd-segment-from-coco
 OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-lines
 OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-regions
+OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-pages
+OCRD_SEGMENT += $(BIN)/ocrd-segment-replace-original
 OCRD_SEGMENT += $(BIN)/ocrd-segment-repair
 $(call multirule,$(OCRD_SEGMENT)): ocrd_segment
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_SEGMENT))
+	$(call delegate_venv,$(OCRD_SEGMENT))
+$(OCRD_SEGMENT): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_tesserocr, $(OCRD_MODULES)),)
@@ -319,15 +346,27 @@ ifneq ($(findstring ocrd_calamari, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_CALAMARI)
 OCRD_CALAMARI := $(BIN)/ocrd-calamari-recognize
 $(OCRD_CALAMARI): ocrd_calamari
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_CALAMARI))
+	$(call delegate_venv,$(OCRD_CALAMARI))
+$(OCRD_CALAMARI): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_pc_segmentation, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_PC_SEGMENTATION)
 OCRD_PC_SEGMENTATION := $(BIN)/ocrd-pc-segmentation
 $(OCRD_PC_SEGMENTATION): ocrd_pc_segmentation
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_PC_SEGMENTATION))
+	$(call delegate_venv,$(OCRD_PC_SEGMENTATION))
+$(OCRD_PC_SEGMENTATION): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+else
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_anybaseocr, $(OCRD_MODULES)),)
@@ -341,7 +380,13 @@ OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-textline
 OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-layout-analysis
 OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-block-segmentation
 $(call multirule,$(OCRD_ANYBASEOCR)): ocrd_anybaseocr
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_ANYBASEOCR))
+	$(call delegate_venv,$(OCRD_ANYBASEOCR))
+$(OCRD_ANYBASEOCR): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_typegroups_classifier, $(OCRD_MODULES)),)
@@ -349,14 +394,26 @@ OCRD_EXECUTABLES += $(OCRD_TYPECLASS)
 OCRD_TYPECLASS := $(BIN)/ocrd-typegroups-classifier
 OCRD_TYPECLASS += $(BIN)/typegroups-classifier
 $(call multirule,$(OCRD_TYPECLASS)): ocrd_typegroups_classifier
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_TYPECLASS))
+	$(call delegate_venv,$(OCRD_TYPECLASS))
+$(OCRD_TYPECLASS): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring sbb_textline_detector, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(SBB_LINE_DETECTOR)
-SBB_LINE_DETECTOR := $(BIN)/ocrd-sbb-line-detector
+SBB_LINE_DETECTOR := $(BIN)/ocrd-sbb-textline-detector
 $(SBB_LINE_DETECTOR): sbb_textline_detector
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(SBB_LINE_DETECTOR))
+	$(call delegate_venv,$(SBB_LINE_DETECTOR))
+$(SBB_LINE_DETECTOR): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
+else
 	$(pip_install)
+endif
 endif
 
 ifneq ($(findstring ocrd_repair_inconsistencies, $(OCRD_MODULES)),)
@@ -384,6 +441,25 @@ define pip_install
 . $(ACTIVATE_VENV) && cd $< && $(PIP) install --no-deps --force-reinstall $(PIP_OPTIONS) .
 endef
 
+# echo a shell script that relays to
+# the (currently active) sub-venv
+# (replacing the outer by the inner
+#  venv directory to ensure there
+#  is no infinite recursion when
+#  the sub-venv does not have the
+#  executable)
+# TODO: variant for relay to Docker
+define delegator
+#!/bin/bash
+. $(ACTIVATE_VENV) && $(BIN)/$(notdir $(1)) "$$@"
+endef
+# create shell scripts that relay to
+# the (currently active) sub-venv
+define delegate_venv
+$(foreach executable,$(1),$(file >$(executable),$(call delegator,$(executable))))
+chmod +x $(1)
+endef
+
 # avoid making these .PHONY so they do not have to be repeated:
 # clstm tesserocr
 $(SHARE)/%: % | $(ACTIVATE_VENV) $(SHARE)
@@ -396,18 +472,6 @@ $(SHARE):
 # At last, add venv dependency (must not become first):
 $(OCRD_EXECUTABLES) $(BIN)/wheel: | $(ACTIVATE_VENV)
 $(OCRD_EXECUTABLES): | $(BIN)/wheel
-
-.PHONY: fix-pip
-# temporary workaround for conflicting requirements between modules:
-# - opencv-python instead of opencv-python-headless (which needs X11 libs)
-#   (pulled by ocrd_anybaseocr and segmentation-runner)
-# - tensorflow>=2.0, tensorflow_gpu in another version
-# - pillow==5.4.1 instead of >=6.2
-fix-pip:
-	. $(ACTIVATE_VENV) && $(PIP) install --force-reinstall $(PIP_OPTIONS_E) \
-		opencv-python-headless \
-		"pillow>=7.1.0" \
-		$$($(PIP) list | grep tensorflow | sed -nE '/tensorflow +1[.]/s/ +/>=2.0/p;/tensorflow-gpu +2[.]/s/ +/~=1.15.3/p')
 
 # At last, we know what all OCRD_EXECUTABLES are:
 all: $(OCRD_EXECUTABLES)
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ endif
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
 	$(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && $(PIP) install --upgrade $(PIP_OPTIONS_E) pip
+	. $(ACTIVATE_VENV) && $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
 
 .PHONY: wheel
 wheel: $(BIN)/wheel
@@ -364,7 +364,7 @@ $(OCRD_PC_SEGMENTATION): ocrd_pc_segmentation
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_PC_SEGMENTATION))
 	$(call delegate_venv,$(OCRD_PC_SEGMENTATION))
-$(OCRD_PC_SEGMENTATION): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+$(OCRD_PC_SEGMENTATION): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf21
 else
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
 	$(pip_install)
@@ -385,7 +385,7 @@ $(call multirule,$(OCRD_ANYBASEOCR)): ocrd_anybaseocr
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_ANYBASEOCR))
 	$(call delegate_venv,$(OCRD_ANYBASEOCR))
-$(OCRD_ANYBASEOCR): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+$(OCRD_ANYBASEOCR): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf22
 else
 	$(pip_install)
 endif
@@ -399,7 +399,7 @@ $(call multirule,$(OCRD_TYPECLASS)): ocrd_typegroups_classifier
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_TYPECLASS))
 	$(call delegate_venv,$(OCRD_TYPECLASS))
-$(OCRD_TYPECLASS): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf2
+$(OCRD_TYPECLASS): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-torch14
 else
 	$(pip_install)
 endif
--- a/Makefile
+++ b/Makefile
@@ -449,7 +449,6 @@ endif
 define pip_install
 . $(ACTIVATE_VENV) && cd $< && $(PIP) install $(PIP_OPTIONS_E) .
 . $(ACTIVATE_VENV) && cd $< && $(PIP) install --no-deps --force-reinstall $(PIP_OPTIONS) .
-. $(ACTIVATE_VENV) && $(PIP) check
 endef
 
 # pattern for recursive make:
--- a/Makefile
+++ b/Makefile
@@ -441,6 +441,7 @@ endif
 define pip_install
 . $(ACTIVATE_VENV) && cd $< && $(PIP) install $(PIP_OPTIONS_E) .
 . $(ACTIVATE_VENV) && cd $< && $(PIP) install --no-deps --force-reinstall $(PIP_OPTIONS) .
+. $(ACTIVATE_VENV) && $(PIP) check
 endef
 
 # pattern for recursive make:
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,8 @@ endif
 all: modules # add OCRD_EXECUTABLES at the end
 
 clean: # add more prerequisites for clean below
-	$(RM) -r $(CURDIR)/venv
+	$(RM) -r $(VIRTUAL_ENV)/share/venv-*
+	$(RM) -r $(CURDIR)/venv # deliberately not using VIRTUAL_ENV here
 
 define HELP
 cat <<"EOF"
@@ -75,6 +76,7 @@ Targets:
 	clean: removes the virtual environment directory, and clean-*
 	clean-tesseract: remove the build directory for tesseract
 	clean-olena: remove the build directory for ocrd_olena
+	deinit: clean, then deinit and rmdir all submodules
 	docker: (re)build a docker image including all executables
 	dockers: (re)build docker images for some pre-selected subsets of modules
 
@@ -109,6 +111,12 @@ $(OCRD_MODULES): always-update
 		git submodule update --init $(GIT_RECURSIVE) $@ && \
 		touch $@; fi
 endif
+
+deinit: clean
+.PHONY: deinit
+deinit:
+	git submodule deinit --all # --force
+	git submodule status | while read stat dir ver; do rmdir $$dir; done
 
 # Get Python modules.
 
--- a/Makefile
+++ b/Makefile
@@ -619,6 +619,8 @@ docker%: Dockerfile $(DOCKER_MODULES)
 	--build-arg PIP_OPTIONS="$(PIP_OPTIONS)" \
 	-t $(DOCKER_TAG):$(or $(*:-%=%),latest) .
 
+docker: DOCKER_MODULES ?= $(OCRD_MODULES)
+docker: docker-latest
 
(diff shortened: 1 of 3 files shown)


bertsky:
Fix #35
Fix #68
Fix #69

This is merely a starting point. One can easily imagine making better/more groups. Also, maybe there is some way to have less writing effort for each module. (But apparently you cannot have control structues or dependencies in canned recipes, only the recipe itself.)
I kept the top-level venv for most modules for now. It _should_ work inside the Docker containers, too.

Makefile
```diff
@@ -384,6 +441,25 @@ define pip_install
...
+# the (currently active) sub-venv
+# (replacing the outer by the inner
+#  venv directory to ensure there
+#  is no infinite recursion when
+#  the sub-venv does not have the
+#  executable)
+# TODO: variant for relay to Docker
+define delegator
+#!/bin/bash
+. $(ACTIVATE_VENV) && $(BIN)/$(notdir $(1)) "$$@"
```
stweil, line 454: That's not sufficient, because arguments which contain whitespace will be split. This can be solved by using a compiled delegator, implemented in C or C++.

Makefile
```diff
@@ -258,12 +275,22 @@ endif
 ifneq ($(findstring ocrd_segment, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_SEGMENT)
 OCRD_SEGMENT := $(BIN)/ocrd-segment-evaluate
+OCRD_SEGMENT += $(BIN)/ocrd-segment-from-masks
```
stweil, line 278: This change is unrelated to the isolated venv, isn't it? I suggest to make it in a separate commit which is easy to review and apply.

Makefile
stweil: I was wrong. With `"$@"` the shell script does the right thing.

bertsky: That's not true. You can read about `$@` in bash manual: it expands to a _list of quoted arguments_.

Makefile
bertsky: Seriously?

Makefile
stweil: Yes, that's what I said.

Makefile
```diff
@@ -258,12 +275,22 @@ endif
...
+OCRD_SEGMENT += $(BIN)/ocrd-segment-from-coco
 OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-lines
 OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-regions
+OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-pages
+OCRD_SEGMENT += $(BIN)/ocrd-segment-replace-original
 OCRD_SEGMENT += $(BIN)/ocrd-segment-repair
 $(call multirule,$(OCRD_SEGMENT)): ocrd_segment
+ifeq (0,$(MAKELEVEL))
```
stweil, line 286: Instead of running `$(pip_install)`, you could also define a new `$(headless_tf1_pip_install)`, avoid the recursive make and get Makefile code which is better readable.

bertsky: No, you cannot avoid the recursive make (because so many things depend on `VIRTUAL_ENV`), and other rules have more than just `pip_install`.

Makefile
```diff
@@ -341,23 +380,41 @@ OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-textline
...
 	$(pip_install)
 endif
+endif
 
 ifneq ($(findstring ocrd_typegroups_classifier, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_TYPECLASS)
 OCRD_TYPECLASS := $(BIN)/ocrd-typegroups-classifier
 OCRD_TYPECLASS += $(BIN)/typegroups-classifier
 $(call multirule,$(OCRD_TYPECLASS)): ocrd_typegroups_classifier
+ifeq (0,$(MAKELEVEL))
```
stweil, line 397: Why does `ocrd_typegroups_classifier` need a separate venv?

bertsky: We do not strictly require it currently. But we already had conflicts between `torch>=1.0` and `torch<1.0` in the past, and I thought it might be useful to group all GPU-enabled modules into sub-venvs for now. (Plus `ocrd_anybaseocr` which has the same `torch` dependency is in that sub-venv now.) This can easily be placed in an extra sub-venv or lifted into the top-venv in the future.

We should also separate into OpenCV with vs without X11 (headless) ASAP. IIRC some modules _implicitly_ still require X11 (for various purposes), so we should dig into transitive dependencies to get this right.

This PR is not meant to solve all these problems at once, but to bring a mechanism for resolving conflicts as we see them. 

Once we gathered some experience, we can also decide whether _isolate when necessary_ is generally better than _isolate when possible_.

bertsky:
I have successfully tested this inside a Docker container. I first had to re-instate our custom `make docker` rule though (see 4ba6628ff534b827de0b1b0e3e2f778f20110a67). Must have become dysfunctional when we introduced the automatic mini/medi/maxi builds via CI.

Makefile
kba: Seems like premature optimization, better to keep it consistent.

Makefile
kba: regarding ocrd_typegroup_classifier and torch, c.f. https://github.com/OCR-D/ocrd_typegroups_classifier/pull/1#discussion_r450880756

Makefile
kba: Neat solution.

stweil:
A non-recursive solution could be based on https://github.com/OCR-D/ocrd_all/compare/master...stweil:tf1.

Makefile
bertsky: It's also directly related in this case: Since the delegation shell scripts under the top-level venv will only be created when listed as `OCRD_EXECUTABLES`, you won't be able to see/run programs anymore if you forget to mention them in the rule.

bertsky:
> A non-recursive solution could be based on [master...stweil:tf1](https://github.com/OCR-D/ocrd_all/compare/master...stweil:tf1). It uses symbolic links to run processors in the TF1 venv.

It's not just about separating TF1. There's many more potential and actual incompatibilities (TF<2.2 vs >=2.2, Keras 2.3.* vs 2.4.*, Torch <1.0 vs <1.5 vs >=1.5, opencv-python vs -headless, Pillow ...). Some of these have merely run under the radar until now. Plus think about going into Docker thin containers (wrapping CLIs or even REST API) as the next step. 

So merely copying all variables and rules (as ub your proposal) we would quickly run into a mess (esp. since you have to drag along any change to them in all others as well). Plus symlinks are not supported by all currently relevant FS IIRC.

stweil:
> Plus symlinks are not supported by all currently relevant FS IIRC.

They are supported, at least on any machine and operating system which I tried so far. We don't have plans to support Windows without WSL, do we?

bertsky:
> > Plus symlinks are not supported by all currently relevant FS IIRC.
> 
> They are supported, at least on any machine and operating system which I tried so far. We don't have plans to support Windows without WSL, do we?

Oh, I think I confused HPFS (which doesn't support them but is irrelevant) and HFS+. And yes, I guess at least within WSL even Windows should be fine with symlinks nowadays.

stweil:
> There's many more potential and actual incompatibilities.

Indeed, there are. Luckily we only have to handle actual incompatibilities.

Are there still problems with opencv-python? I thought all of them are using opencv-python-headless now?

Makefile
```diff
@@ -21,9 +21,9 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
 # (but re-use if already active):
 VIRTUAL_ENV ?= $(CURDIR)/venv
```
stweil, line 22: I suggest to add macros for all other virtual environments and document what they are used for here, directly after line 22.

Makefile
```diff
@@ -189,26 +188,44 @@ OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-train
 OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-eval
 OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-repl
 $(call multirule,$(OCRD_COR_ASV_ANN)): cor-asv-ann
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) $(notdir $(OCRD_COR_ASV_ANN))
+	$(call delegate_venv,$(OCRD_COR_ASV_ANN))
+$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(VIRTUAL_ENV)/share/venv-headless-tf1
```
stweil, line 194: Here (and for the other sub virtual environments) I'd use a new macro name instead of `$(VIRTUAL_ENV)/share/venv-headless-tf1`. See my comment for line 22 above. That improves readability, helps avoiding typos and better documents the purpose.

bertsky:
> > There's many more potential and actual incompatibilities.
> 
> Indeed, there are. Luckily we only have to handle actual incompatibilities.

The main point was: there are too many for your kind of solution

The other point was: we don't really know what incompatibilities we already have. They rarely surface as long as no one here uses complex workflows in Docker images and we don't have regression tests for that.

Besides, potential can always become actual when a module changes in the future.

kba:
Code is still compiling but so far it looks good :)

> We don't have plans to support Windows without WSL, do we?

IIUC there is a potential project partner for phase 3 that intends to roll out OCR-D on a pure Windows (Server) environment. We'll continue to target Ubuntu 18.04 as the main platform and when in doubt, we'll stick to POSIX. However, symlinks seem like a less maintainable approach and unless there's an advantage to it over the more generalized multi-venv solution here, I'd avoid them.

stweil:
> IIUC there is a potential project partner for phase 3 that intends to roll out OCR-D on a pure Windows (Server) environment.

I noticed that, too, but think that's not feasible, so IMO even trying it is a waste of time. With Docker or a virtual Linux machine there exist solutions which are known to work. And with newer Windows WSL is also a good solution.

stweil:
> However, symlinks seem like a less maintainable approach and unless there's an advantage to it over the more generalized multi-venv solution here, I'd avoid them.

Both @bertsky's and my solution use multi-venv. Running a `venv/sub/bin/processor` (not in `PATH`) from `venv/bin/processor` (in `PATH`) can be implemented in different ways. One way are individual wrapper scripts `venv/sub/bin/processor` (solution here, wrapper scripts must be generated during make and use disk space), another way are symbolic links `venv/sub/bin/processor` which link to a single wrapper script which is placed in `venv/sub/bin` (my solution, only links must be created, less disk space required). Both variants are not difficult to maintain.

stweil:
> There's many more potential and actual incompatibilities.

One known existing problem is that some prebuild Tensorflow modules don't work with every Python version. This is not a problem as long as we stick to Ubuntu 18.04, but for newer Debian / Ubuntu it could be required to use different Python versions for the different venvs.

kba:
> > However, symlinks seem like a less maintainable approach and unless there's an advantage to it over the more generalized multi-venv solution here, I'd avoid them.
> 
> Both @bertsky's and my solution use multi-venv. Running a `venv/sub/bin/processor` (not in `PATH`) from `venv/bin/processor` (in `PATH`) can be implemented in different ways. One way are individual wrapper scripts `venv/sub/bin/processor` (solution here, wrapper scripts must be generated during make and use disk space), another way are symbolic links `venv/sub/bin/processor` which link to a single wrapper script which is placed in `venv/sub/bin` (my solution, only links must be created, less disk space required). Both variants are not difficult to maintain.

Yes, both solutions are multi-env but I like that @bertsky's approach is more generic, not just for the TF1/TF2 issue but possible other dependency incompatibilities. The venv delegation mechanism is the same by and large. Wrapper scripts over symlinks have the advantage that they are explicitly in the script whereas symlinks contain that information implicitly. If the OS or the filesystem - think a removable FAT or NTFS drive - don't support them, this needlessly breaks the deployment IMHO.

> > (Server) environment.
> 
> I noticed that, too, but think that's not feasible, so IMO even trying it is a waste of time. With Docker or a virtual Linux machine there exist solutions which are known to work. And with newer Windows WSL is also a good solution.

I agree with you that Linux, Docker or even WSL are the best option. And we will continue to target Ubuntu 18.04. But there are Windows Server environments out there where Linux in a VM or Docker are not viable for whatever reason. We're not actively supporting that at the moment but if somebody made the effort, I would not want to prevent it with platform-dependent design choices.

> One known existing problem is that some prebuild Tensorflow modules don't exist for every Python version. This is not a problem as long as we stick to Ubuntu 18.04, but for newer Debian / Ubuntu it could help to support different Python versions for the different venvs.

Fully agree, it is unfortunate that we cannot support Python 3.8 at the moment. I have been thinking about eleviating this by setting up a PPA or other Debian repo and generating `.deb` packages for the tools in ocrd_all, as well as common models. This would give us the freedom to create compiled packages for tesseract, tensorflow, possibly python, as well as python packages like tensorflow, torch and simplify the docker builds.

stweil:
> Yes, both solutions are multi-env but I like that @bertsky's approach is more generic.

Both approaches are identically generic. I only have implemented TF1 as an example, but of course any number of other venvs can be added in the same way, and it would also be possible to use hard files instead of symbolic links when needed, so that's not the essential difference.

The difference is how the Makefile looks like and whether it requires recursive make or not.

stweil:
As we all agree that there will be more venvs needed than two, I think it is important to have also a common understanding of the contents of those different venvs. It should be clearly defined what goes into the primary venv and what is the difference for the sub-venvs. The PR currently uses `venv-headless-tf1` and `venv-headless-tf2` which are currently not documented explicitly, and I simply don't know what they stand for. `tf1` and `tf2` are clear, but why is `headless` an important attribute of both venvs?

I'd try to put most modules into the primary venv. In my understanding that includes modules without peculiarities and modules which follow best / recommended practice (using current software versions => TF2, using OpenCV headless, ...). So I would not make a primary venv without and TF software and put all TF software in different sub-venvs, and similar with OpenCV headless or not (where we currently don't have a problem as far as I know).

bertsky:
> less disk space required

I think we all agree this is a non-argument (for a few bytes of shell script).

We also (more or less) agreed that FS should be able to cope with symlinks.

> Both variants are not difficult to maintain.

I argued precisely against that above. @stweil's solution needs to mirror many variables and recipes for each additional sub-venv:
- `pip_install` → `*_pip_install`
- `VIRTUAL_ENV` → `*_VIRTUAL_ENV`
- `BIN` → `*_BIN`
- `ACTIVATE_VENV` → `ACTIVATE_*_VENV`
- `OCRD_EXECUTABLES` → `*_EXECUTABLES`

IMO this not only bloats the Makefile, it also makes it less readable and more error-prone.

> Both approaches are identically generic. I only have implemented TF1 as an example, but of course any number of other venvs can be added in the same way

They are not IMO: the non-recursive approach cannot yet deal with non-Python (or not purely Python) install targets. It's not trivial to wrap recipes like `make -C $< install`, which some modules will need, in the same way that `pip_install` has been redefined. In contrast, the recursive approach is fully transparent w.r.t. all venv-dependent choices (variables, rules and recipes alike).

> The PR currently uses `venv-headless-tf1` and `venv-headless-tf2` which are currently not documented explicitly, and I simply don't know what they stand for. `tf1` and `tf2` are clear, but why is `headless` an important attribute of both venvs?

I have explained that above already. The current choices are merely illustrations to get the new mechanism going and fix the most pressing inconsistencies. Headless stands for `opencv-python-headless` (as opposed to `opencv-python`, which some modules might still need explicitly or implicitly). We will have the need for _combinations_ of certain requirements, like TF1 with headless, or TF1 with X11 (think `pyplot`), or TF2 with older Torch etc.

> So I would not make a primary venv without any TF software and put all TF software in different sub-venvs, and similar with OpenCV headless or not (where we currently don't have a problem as far as I know).

Like I said above, I believe it's better to _group all GPU-enabled modules into sub-venvs for now_. Not only because they usually have mutually exclusive sets of dependencies, but because we still have no mechanism to make GPU work in Docker yet, so grouping them might facilitate future solutions.

kba:
I've tested and unfortunately, I cannot get it working. After `make all` in a fresh checkout, there is no `venv/share/venv-headless-tf2` directory, therefore the wrapper script for e.g. `ocrd-anybaseocr-binarize`, which looks like

```
#!/bin/bash
. /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/bin/activate && /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/bin/ocrd-anybaseocr-binarize "$@"
```

fails to run. Any ideas what to look for to fix this?

bertsky:
> I've tested and unfortunately, I cannot get it working. After `make all` in a fresh checkout, there is no `venv/share/venv-headless-tf2` directory, therefore the wrapper script

Sorry, I had tested the final code only with venv already active, which ensured `VIRTUAL_ENV` got exported to sub-make because it was already defined in the environment (not by make itself). I now fixed that by exporting all variables to sub-make.

I have also improved the handling of submodule directories in recursive calls: Now sub-make uses `-B` to ensure the nested recipe is run when the outer recipe was deemed necessary, and `-o $<` to avoid repeating the actual submodule update in doing so.

Makefile
bertsky: This goes against the principle of localizing anything specific for targets near their rules, and would necessitate duplicating all other macros/variables that depend on `VIRTUAL_ENV` for all venv configs we want to define (which will likely be quite a few in the future). In contrast, this PR's conceptual idea is to define venv configs "ad-hoc" in the module rules. Whenever we feel the need to differentiate a little more, we can simply rewrite more modules into sub-venv form, or distinguish more venv configs from each other.

Makefile
bertsky: There's a trade-off between readability (in the sense of _intelligibility_) and writing effort (targeting _brevity_) here. I have already speculated on the possibilities to condense this pattern further in my [opening statement](https://github.com/OCR-D/ocrd_all/pull/118#issuecomment-650432407), but decided against them to ensure readibility.

(For example, it's trivial to condense the `MAKE` and `delegate_venv` line into one, but then it's not so obvious anymore that this recipe contains a recursive `MAKE` call. There are no constructs to my knowledge which can encompass both the recipes and the additional target-specific variable definition in the recursive branch.) 

Ideas/proposals how to optimise both are still welcome, as long as they stay faithful to the general idea of transparent recursive rules.

kba:
Thanks for the update. think the issue was a spurious entry in `$PATH` from a previous venv activation, which lead to `python3` to point to a (deleted) venv. venv-headless-tf{1,2} are created now, waiting for olena to compile now.

kba:
:tada: It works now, great!

bertsky:
> It works now, great!

One thing we could add now is `pip check` after `make all` or even in each `pip_install`. This should now work everywhere. And if not – we get an early (build-time) alarm where to send the fire brigade...

bertsky:
> `pip check` after `make all` or even in each `pip_install`.

The latter seemed like the best option (will target each venv, and any error will appear right after the respective module). Thumbs crossed for CI!

kba:
I have another issue: Not all the wrappers are executable:

```sh
$ ls -la venv/share/bin

-rw-rw-r-- 1 kba kba      224 Jul 17 12:34 ocrd-anybaseocr-binarize
-rw-rw-r-- 1 kba kba      234 Jul 17 12:34 ocrd-anybaseocr-block-segmentation
-rw-rw-r-- 1 kba kba      220 Jul 17 12:34 ocrd-anybaseocr-crop
-rw-rw-r-- 1 kba kba      222 Jul 17 12:34 ocrd-anybaseocr-deskew
-rw-rw-r-- 1 kba kba      222 Jul 17 12:34 ocrd-anybaseocr-dewarp
-rw-rw-r-- 1 kba kba      231 Jul 17 12:34 ocrd-anybaseocr-layout-analysis
-rw-rw-r-- 1 kba kba      224 Jul 17 12:34 ocrd-anybaseocr-textline
-rw-rw-r-- 1 kba kba      221 Jul 17 12:34 ocrd-anybaseocr-tiseg
-rwxrwxr-x 1 kba kba      223 Jul 17 12:32 ocrd-calamari-recognize
-rwxrwxr-x 1 kba kba      288 Jul 17 13:20 ocrd-cis-align
-rwxrwxr-x 1 kba kba      272 Jul 17 13:20 ocrd-cis-data
-rwxrwxr-x 1 kba kba      309 Jul 17 13:20 ocrd-cis-ocropy-binarize
-rwxrwxr-x 1 kba kba      301 Jul 17 13:20 ocrd-cis-ocropy-clip
```

This also concerns `ocrd-typegroup-classifier` and the non-OCR-D `typegroups-classifier` executables. `chmod a+x venv/bin/ocrd-*` fixes this but apparently, ocrd_anybaseocr was not properly installed:

```
$ ocrd-anybaseocr-binarize --version
/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/bin/ocrd-anybaseocr-binarize: line 2: /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/bin/ocrd-anybaseocr-binarize: No such file or directory
```

I did run into "No space left on device" in the middle of `make all`, that might be the problem. Deleted a few venvs and obsolete models, trying again (with the fix-pip commit).
```
```

kba:
Did `rm -rf venv/bin/ocrd-anybaseocr-` to trigger a rebuild on `make all`. Installed correctly, scripts are executable but now

```
$ ocrd-anybaseocr-binarize --version
  File "/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/lib/python3.6/site-packages/keras/__init__.py", line 3, in <module>
    from tensorflow.keras.layers.experimental.preprocessing import RandomRotation
ModuleNotFoundError: No module named 'tensorflow.keras.layers.experimental.preprocessing'

During handling of the above exception, another exception occurred:

  File "/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/bin/ocrd-anybaseocr-binarize", line 5, in <module>
    from ocrd_anybaseocr.cli.cli import ocrd_anybaseocr_binarize
  File "/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/lib/python3.6/site-packages/ocrd_anybaseocr/cli/cli.py", line 8, in <module>
    from ocrd_anybaseocr.cli.ocrd_anybaseocr_tiseg import OcrdAnybaseocrTiseg
  File "/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/lib/python3.6/site-packages/ocrd_anybaseocr/cli/ocrd_anybaseocr_tiseg.py", line 35, in <module>
    from keras.models import load_model
  File "/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/share/venv-headless-tf2/lib/python3.6/site-packages/keras/__init__.py", line 6, in <module>
    'Keras requires TensorFlow 2.2 or higher. '
ImportError: Keras requires TensorFlow 2.2 or higher. Install TensorFlow via `pip install tensorflow`
``

**Comment by bertsky** [comment]
> Thumbs crossed for CI!

argh:

```
tensorboard 1.15.0 has requirement setuptools>=41.0.0, but you have setuptools 39.0.1.
```

I've seen that before and still don't understand why `pip` does not resolve that on its own (by updating `setuptools`).

**Comment by bertsky** [comment]
> ImportError: Keras requires TensorFlow 2.2 or higher. Install TensorFlow via `pip install tensorflow`

Probably need to differentiate between 2.1 (for ocrd_pc_segmentation, maybe others) and 2.2 (ocrd_anybaseocr?). Using `venv-headless-tf22`?

**Comment by kba** [comment]
> > ImportError: Keras requires TensorFlow 2.2 or higher. Install TensorFlow via `pip install tensorflow`
> 
> Probably need to differentiate between 2.1 (for ocrd_pc_segmentation, maybe others) and 2.2 (ocrd_anybaseocr?). Using `venv-headless-tf22`?

Fine by me, if that's what it takes. I can only tell for sure once the new run finishes.

**Comment by kba** [comment]
Sorry for the noise but here's the newest hickup:

```
. /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/bin/activate && make -C core install PIP_INSTALL="pip3 install --force-reinstall "
make[1]: Entering directory '/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/core'
make[1]: *** No rule to make target 'install'.  Stop.
make[1]: Leaving directory '/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/core'
Makefile:135: recipe for target '/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/bin/ocrd' failed
make: *** [/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/venv/bin/ocrd] Error 2
```

Apparently not all modules have been checked out:


```
du -hs *|sort -h|head
4.0K	clstm
4.0K	cor-asv-fst
4.0K	core
4.0K	Dockerfile
4.0K	LICENSE
4.0K	ocrd_kraken
4.0K	ocrd_ocropy
4.0K	opencv-python
```

**Comment by bertsky** [comment]
> Running `make modules` does not fix the problem.

That's strange. The `core` module is not even part of the changes in this PR. Do you perhaps have a `Makefile.local` or `OCRD_MODULES` in your environment? And what does `git submodule status` say?

**Comment by bertsky** [comment]
> I need a bigger hard drive or the ability to build ocrd_all on an external FAT/NTFS HDD. Trying again from scratch, fingers crossed that remaining space is enough...

I recommend against vfat. For fresh start, I always do:
```sh
git submodule deinit --all -f
git submodule status | while read stat dir ver; do rmdir $dir; done
```

kba:
> > Running `make modules` does not fix the problem.
> 
> That's strange. The `core` module is not even part of the changes in this PR. Do you perhaps have a `Makefile.local` or `OCRD_MODULES` in your environment? And what does `git submodule status` say?

This was a clean checkout, so no Makefile.local or variable overrides. Cannot say what `git submodule status` says because I manually ran `git submodule update --init --recursive`. But since the build failed again with `No space left on device` I'll have to run it yet again anyway and will report back if the issue persists.

> I recommend against vfat.

That won't work anyway (I tried and failed before).

> ```
> git submodule deinit --all -f
> git submodule status | while read stat dir ver; do rmdir $dir; done
> rm -fr venv
> ```

Thx. I'll try with a fresh checkout but this is useful documentation for debugging. Perhaps add to the contributor guide?

bertsky:
> because I manually ran `git submodule update --init --recursive`.

That would take substantially more space than just `make modules` (not all modules are needed, and not all modules are needed recursively).

> but this is useful documentation for debugging. Perhaps add to the contributor guide?

I could introduce it as `clean-modules`. Should `clean` also depend on it?

bertsky:
> But since the build failed again with `No space left on device`

BTW your problems could also be explained by that alone. As soon as `git submodule` becomes inconsistent with the actual subdirectories, it often cannot recover by its own means.

bertsky:
```
tensorflow 2.2.0 has requirement scipy==1.4.1; python_version >= "3", but you have scipy 1.5.1.
```

Damn!

bertsky:
TF's handling of GH issues is a joke:
https://github.com/tensorflow/tensorflow/issues/40884

_Oh, this issue hasn't been solved in 7 days! Makes us look bad. Better close the issue automatically. Users seriously wounded will come back to our knife anyway._

kba:
> > But since the build failed again with `No space left on device`
> 
> BTW your problems could also be explained by that alone. As soon as `git submodule` becomes inconsistent with the actual subdirectories, it often cannot recover by its own means.

The space ran out after this, there was at least enough space left for `git submodule update --init --recursive`.

Cannot reproduce now, though, fortunately:

```
Submodule 'core' (https://github.com/OCR-D/core.git) registered for path 'core'
Cloning into '/home/kba/build/github.com/OCR-D/monorepo/ocrd_all/core'...
Submodule path 'core': checked out 'f708420d806ab6363917e3a53519d85505327568'
```

Maybe a network problem.


> > but this is useful documentation for debugging. Perhaps add to the contributor guide?
> 
> I could introduce it as `clean-modules`. Should `clean` also depend on it?

`clean-modules`: :+1: But keep it explicit, I wouldn't want to accidentally lose all local changes to the modules because of `make clean`.

bertsky:
> TF's handling of GH issues is a joke:
> [tensorflow/tensorflow#40884](https://github.com/tensorflow/tensorflow/issues/40884)

Looks like this has since been removed [upstream](https://github.com/tensorflow/tensorflow/commit/78026d6a66f7f0fc80c69b1a2f8843616f4cd2a7) – 4 days ago. How long do they take to update their releases on PyPI?

Who knows what other failures are lurking behind this one. Perhaps `pip check` was too ambitious for a system like `pip` which is incapable of conflict management?

bertsky:
> `clean-modules`: +1 But keep it explicit, I wouldn't want to accidentally lose all local changes to the modules because of `make clean`.

And what if we made that `deinit --all` (without `--force`)?

kba:
:( No space issues this time but again, something went wrong with anybaseocr and typegroups classifier. In fact, only these tools have been installed:

```
$ ls -l share/venv*/bin/ocrd-*
-rwxrwxr-x 1 kba kba 329 Jul 17 16:35 share/venv-headless-tf1/bin/ocrd-calamari-recognize
-rwxrwxr-x 1 kba kba 344 Jul 17 16:14 share/venv-headless-tf1/bin/ocrd-cor-asv-ann-evaluate
-rwxrwxr-x 1 kba kba 342 Jul 17 16:14 share/venv-headless-tf1/bin/ocrd-cor-asv-ann-process
-rwxrwxr-x 1 kba kba 296 Jul 17 16:13 share/venv-headless-tf1/bin/ocrd-dummy
-rwxrwxr-x 1 kba kba 324 Jul 17 16:14 share/venv-headless-tf1/bin/ocrd-keraslm-rate
-rwxrwxr-x 1 kba kba 347 Jul 17 16:46 share/venv-headless-tf1/bin/ocrd-sbb-textline-detector
-rwxrwxr-x 1 kba kba 324 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-evaluate
-rwxrwxr-x 1 kba kba 334 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-extract-lines
-rwxrwxr-x 1 kba kba 334 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-extract-pages
-rwxrwxr-x 1 kba kba 338 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-extract-regions
-rwxrwxr-x 1 kba kba 326 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-from-coco
-rwxrwxr-x 1 kba kba 328 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-from-masks
-rwxrwxr-x 1 kba kba 320 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-repair
-rwxrwxr-x 1 kba kba 340 Jul 17 16:33 share/venv-headless-tf1/bin/ocrd-segment-replace-original
-rwxrwxr-x 1 kba kba 297 Jul 17 16:35 share/venv-headless-tf21/bin/ocrd-dummy
-rwxrwxr-x 1 kba kba 331 Jul 17 16:36 share/venv-headless-tf21/bin/ocrd-pc-segmentation
-rwxrwxr-x 1 kba kba 300 Jul 17 16:44 share/venv-headless-torch14/bin/ocrd-dummy
-rwxrwxr-x 1 kba kba 315 Jul 17 16:46 share/venv-headless-torch14/bin/ocrd-typegroups-classifier
```

Wrapper scripts for `ocrd-anybaseocr-*` are not executable. Of course I ran this `make all` outside tmux and don't have the output to debug, because it's Friday afternoon and things cannot just work for once.

Off to another fresh start, again.

bertsky:
> something went wrong with anybaseocr and typegroups classifier
[...]
> Wrapper scripts for `ocrd-anybaseocr-*` are not executable

Are you sure you didn't just fall into the above `pip check` trap, as does CI now?

kba:
> > something went wrong with anybaseocr and typegroups classifier
> > [...]
> > Wrapper scripts for `ocrd-anybaseocr-*` are not executable
> 
> Are you sure you didn't just fall into the above `pip check` trap, as does CI now?

Yes, there was no error that broke `make all` AFAICS. This time I'm doing `make all|tee make-all.log` so at least I can inspect the results later...

kba:
No luck

```
ls -l venv/share/venv*/bin/ocrd-*
-rwxrwxr-x 1 kba kba 329 Jul 17 17:40 venv/share/venv-headless-tf1/bin/ocrd-calamari-recognize
-rwxrwxr-x 1 kba kba 344 Jul 17 17:19 venv/share/venv-headless-tf1/bin/ocrd-cor-asv-ann-evaluate
-rwxrwxr-x 1 kba kba 342 Jul 17 17:19 venv/share/venv-headless-tf1/bin/ocrd-cor-asv-ann-process
-rwxrwxr-x 1 kba kba 296 Jul 17 17:18 venv/share/venv-headless-tf1/bin/ocrd-dummy
-rwxrwxr-x 1 kba kba 324 Jul 17 17:19 venv/share/venv-headless-tf1/bin/ocrd-keraslm-rate
-rwxrwxr-x 1 kba kba 324 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-evaluate
-rwxrwxr-x 1 kba kba 334 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-extract-lines
-rwxrwxr-x 1 kba kba 334 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-extract-pages
-rwxrwxr-x 1 kba kba 338 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-extract-regions
-rwxrwxr-x 1 kba kba 326 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-from-coco
-rwxrwxr-x 1 kba kba 328 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-from-masks
-rwxrwxr-x 1 kba kba 320 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-repair
-rwxrwxr-x 1 kba kba 340 Jul 17 17:39 venv/share/venv-headless-tf1/bin/ocrd-segment-replace-original
-rwxrwxr-x 1 kba kba 297 Jul 17 17:41 venv/share/venv-headless-tf21/bin/ocrd-dummy
-rwxrwxr-x 1 kba kba 331 Jul 17 17:42 venv/share/venv-headless-tf21/bin/ocrd-pc-segmentation
-rwxrwxr-x 1 kba kba 338 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-binarize
-rwxrwxr-x 1 kba kba 358 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-block-segmentation
-rwxrwxr-x 1 kba kba 338 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-crop
-rwxrwxr-x 1 kba kba 334 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-deskew
-rwxrwxr-x 1 kba kba 334 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-dewarp
-rwxrwxr-x 1 kba kba 352 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-layout-analysis
-rwxrwxr-x 1 kba kba 338 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-textline
-rwxrwxr-x 1 kba kba 332 Jul 17 17:47 venv/share/venv-headless-tf22/bin/ocrd-anybaseocr-tiseg
-rwxrwxr-x 1 kba kba 297 Jul 17 17:46 venv/share/venv-headless-tf22/bin/ocrd-dummy
```

But a log
[make-all.log](https://github.com/OCR-D/ocrd_all/files/4939191/make-all.log)

bertsky:
> `ocrd-anybaseocr` wrapper scripts still not executable

thanks for the log. This is proof to me that you _did_ fall into the `pip check` trap. That failed (with above scipy issue), and then the recipe stopped short, right before the `chmod +x` call.

`make all` should have stopped right there. Maybe you used `make -j` and did not notice the error further up?

Anyway, since I have reverted the `pip check` step, because it's unclear when a fix for TF is released via PyPI, this should be gone.

Makefile
stweil: `share` is normally used for files which can be shared between different architectures. Here I'd not use it because the sub venv also includes binaries.

bertsky: > `share` is normally used for files which can be shared between different architectures. Here I'd not use it because the sub venv also includes binaries.

Good point. What else would you propose?

stweil: I have no strong preference, maybe `$(VIRTUAL_ENV)/venv-headless-tf1` (if we except less than 10 venvs), `$(VIRTUAL_ENV)/venvs/headless-tf1` or even `$(VIRTUAL_ENV)/venvs/tf1/headless` (if we expect a large number of venvs).

bertsky: Ok, so I have used `$(VIRTUAL_ENV)/local/sub-venv/headless-tf1` etc.

kba:
> make all should have stopped right there. Maybe you used make -j and did not notice the error further up?

No, it was plain `make all`. But I have since tried again and so far have no more issues. @cneud also verified this. Merging :tada:

PR #119, opened 2020-06-30
release.sh: use master instead of dev on ocrd_cis

bertsky:
...as this is now the default branch again.


bertsky:
(Should probably update the submodule as well?)

kba:
> (Should probably update the submodule as well?)

I'll do another round of `release.sh` and send a PR.

PR #121, opened 2020-07-10
Avoid repeated Tesseract builds

stweil:
Running autogen.sh creates a new configure file. That updates the timestamp for tesseract.
With the dependency introduced in commit e1b30b95, that triggered a new run of autogen.sh
when make was run again and so on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -649,12 +649,12 @@ $(TESSDATA)/%.traineddata:
 	$(call WGET,$@,$(TESSDATA_URL)/raw/master/$(notdir $(call stripdir,$@))/$(notdir $@)) || \
 		{ $(RM) $@; false; }
 
-tesseract/configure: tesseract
+tesseract/Makefile.in: tesseract
 	cd tesseract && ./autogen.sh
 
 # Build and install Tesseract.
 TESSERACT_CONFIG ?= --disable-openmp --disable-shared CXXFLAGS="-g -O2 -fPIC"
-$(BIN)/tesseract: tesseract/configure
+$(BIN)/tesseract: tesseract/Makefile.in
 	mkdir -p $(VIRTUAL_ENV)/build/tesseract
 	cd $(VIRTUAL_ENV)/build/tesseract && $(CURDIR)/tesseract/configure --prefix="$(VIRTUAL_ENV)" $(TESSERACT_CONFIG)
 	cd $(VIRTUAL_ENV)/build/tesseract && $(MAKE) install
@@ -681,7 +681,7 @@ TESSTRAIN_EXECUTABLES += $(BIN)/wordlist2dawg
 .PHONY: install-tesseract-training
 install-tesseract-training: $(TESSTRAIN_EXECUTABLES)
 
-$(call multirule,$(TESSTRAIN_EXECUTABLES)): tesseract/configure
+$(call multirule,$(TESSTRAIN_EXECUTABLES)): tesseract/Makefile.in
 	mkdir -p $(VIRTUAL_ENV)/build/tesseract
 	$(MAKE) -C $(VIRTUAL_ENV)/build/tesseract training-install
 


bertsky:
> Running autogen.sh creates a new configure file. That updates the timestamp for tesseract.

Correct so far.

But note that this will make `tesseract` and `tesseract/configure` have **the exact same timestamp**.

> With the dependency introduced in commit [e1b30b9](https://github.com/OCR-D/ocrd_all/commit/e1b30b951bf0beef835954cb75051a8b43aa4b81), that triggered a new run of autogen.sh
> when make was run again and so on.

I cannot reproduce any such behaviour (even after simulating different conditions with `-W`).

And I fail to see how this could ever happen, because `tesseract` has the same mtime/ctime as `tesseract/configure` (see above).

@stweil can you state a recipe how to end up with your dependency loop?

stweil:
> Makes sense. To trigger a rebuild, one can always do make clean-tesseract, right?

Right. But a rebuild should also be triggered when the tesseract submodule change.

> If you remove the dependency on tesseract, nothing will work anymore: not cloning the submodule after cloning a fresh ocrd_all, and not updating the submodule after updating ocrd_all.

You are right. The dependency on the tesseract submodule is still needed, but for the tesseract executable, not for the configure file. I fixed that now.

> But note that this will make tesseract and tesseract/configure have the exact same timestamp.

Try `ls -lat tesseract`, then you will see that's not the case. Several files and directories are written after `configure`, so the timestamp of `tesseract` is slightly newer than `tesseract/configure`. The delta is not always large enough to trigger a rebuild with the old code, but for me that happened in more than every second case.

Makefile
```diff
@@ -459,11 +459,11 @@ $(TESSDATA)/%.traineddata:
 	$(call WGET,$@,$(TESSDATA_URL)/raw/master/$(notdir $(call stripdir,$@))/$(notdir $@)) || \
 		{ $(RM) $@; false; }
 
-tesseract/configure: tesseract
+tesseract/configure:
 	cd tesseract && ./autogen.sh
 
 # Build and install Tesseract.
-$(BIN)/tesseract: tesseract/configure
+$(BIN)/tesseract: tesseract/configure tesseract
```
bertsky, line 466: That still would not work. For one, there's no `tesseract/autogen.sh` after a fresh clone of ocrd_all. But even if you would add `tesseract` (the module dir) as an order-only dependency (`tesseract/configure: | tesseract`), you don't know when Tesseract updates will require a reconfigure (not just a rebuild).

bertsky:
> > But note that this will make tesseract and tesseract/configure have the exact same timestamp.
> 
> Try `ls -lat tesseract`, then you will see that's not the case. Several files and directories are written after `configure`, so the timestamp of `tesseract` is slightly newer than `tesseract/configure`. The delta is not always large enough to trigger a rebuild with the old code, but for me that happened in more than every second case.

Well `ls -lat` does not tell you more than `ls -l`, it just sorts different. I have compared the timestamps of `tesseract` and `tesseract/configure` via `stat -c %Y` and it _always_ gives me the same Epoch seconds. I mean what you say sounds plausible, but I would like to be able to reproduce. I wonder if there's anything else this depends on – like FS or make options like `-j`...

Makefile
stweil: Did you try it? I did a fresh clone, and `make all` works perfectly. Tesseract's `make` handles reconfigure automatically, so there is no need to run `autogen.sh` more than once.

stweil:
Sure, it depends on a lot of factors, for example CPU performance or disk speed. Here is the result for my "normal" desktop PC running Debian with EXT4 on a SSD directly after running `./autogen.sh` in the `tesseract` directory:

    $ ls --full-time -d . configure
    drwxr-xr-x 17 stweil stweil   4096 2020-07-13 08:47:53.586047367 +0200 .
    -rwxr-xr-x  1 stweil stweil 657295 2020-07-13 08:47:52.470047314 +0200 configure

So you see that the timestamps differ by a little bit more than a second. Maybe for you the difference is smaller. Then chances are higher that the timestamps rounded to seconds are identical and no rebuild will occur.

bertsky:
> Maybe for you the difference is smaller. Then chances are higher that the timestamps rounded to seconds are identical and no rebuild will occur.

As I said, I _always_ have the same timestamp for both, including equal nanoseconds. It's not a race. There must be some other reason it's different for you.

Makefile
bertsky: > Did you try it?
> I did a fresh clone, and `make all` works perfectly.

I have. It does indeed work single-threaded if you try `make all` or `make install-tesseract`. 

But that's merely by chance: For `$(BIN)/tesseract` there are 2 direct prerequisites now, and make just happens to go for `tesseract` (the submodule) before doing `tesseract/configure`. But if you build with multiple jobs (e.g. `-j2`) or if you want to have the training utils (`make lstmtraining`), then you do end up with:

```
cd tesseract && ./autogen.sh
/bin/sh: 1: cd: can't cd to tesseract
```

> Tesseract's `make` handles reconfigure automatically, so there is no need to run `autogen.sh` more than once.

That's good to know, maybe we can work something out. But the primary dependency on the submodule must be made explicit IMO. We should best try to isolate the real cause of why your `autogen.sh` behaves different (creating more files after `configure`) than mine.

stweil:
That's indeed surprising. Here is a more complete list of the timestamps after `./autogen.sh`:

    $ ls -lat --full-time | head
    insgesamt 1780
    drwxr-xr-x  6 stweil stweil   4096 2020-07-13 11:31:49.582516916 +0200 unittest
    drwxr-xr-x  2 stweil stweil   4096 2020-07-13 11:31:49.550516915 +0200 config
    drwxr-xr-x  4 stweil stweil   4096 2020-07-13 11:31:49.434516909 +0200 tessdata
    drwxr-xr-x  3 stweil stweil   4096 2020-07-13 11:31:49.294516902 +0200 java
    drwxr-xr-x  2 stweil stweil   4096 2020-07-13 11:31:49.278516902 +0200 doc
    drwxr-xr-x 18 stweil stweil   4096 2020-07-13 11:31:49.258516901 +0200 .
    -rw-r--r--  1 stweil stweil 490990 2020-07-13 11:31:49.258516901 +0200 Makefile.in
    drwxr-xr-x  2 stweil stweil   4096 2020-07-13 11:31:48.722516875 +0200 autom4te.cache
    -rwxr-xr-x  1 stweil stweil 657716 2020-07-13 11:31:47.778516830 +0200 configure

As you can see, `Makefile.in` was made after `configure` and `autom4te.cache`, so it sets the timestamp of the directory. That's also the order to be expected from what `autogen.sh` does.

Makefile
stweil: > But if you build with multiple jobs (e.g. -j2) or if you want to have the training utils (make lstmtraining), then you do end up with ...

Again my question: did you try it? I did, and `make lstmtraining` works. The build always starts with syncing the submodules. `make -j` does not work. Parallel builds fail because they try parallel submodule syncs and fail there.

bertsky: > Again my question: did you try it?

I have, as I already stated. See above.

> I did, and `make lstmtraining` works. The build always starts with syncing the submodules.

Not for me.

> Parallel builds fail because they try parallel submodule syncs and fail there.

I have never seen a concurrent attempt at syncing any single submodule since we introduced that. Would you care to elaborate how you reached that conclusion? (A `make -d` log might help.)

AFAICT the parallel build fails precisely because now both prereqs `tesseract/configure tesseract` get built "first", and since the former _implicitly_ depends on the latter now, the former gets run too soon.

stweil: `make all -j` does not build anything at all. It terminates without error after syncing all submodules.
`make all -j6` fails with an error message which is completely unrelated to Tesseract:

    error: Konnte Konfigurationsdatei '.git/config' nicht sperren.: Die Datei existiert bereits

bertsky: > `make all -j6` fails with an error message which is completely unrelated to Tesseract:
> 
> ```
> error: Konnte Konfigurationsdatei '.git/config' nicht sperren.: Die Datei existiert bereits
> ```

Looks like we have a bigger problem than we thought. How reproducible is that for you? Have you ever observed that before? Does it really depend on the _number of jobs_, or is it rather a random event (which you also get after running `make all -j` for a couple of times)?

bertsky:
> Here is a more complete list of the timestamps after `./autogen.sh`:

For me, that's:
```
-rw-r--r--  1 xbert xbert   8270 2020-04-08 14:43:25.081349667 +0200 README.md
-rw-r--r--  1 xbert xbert  29796 2020-06-17 14:55:22.092285032 +0200 Makefile.am
-rw-r--r--  1 xbert xbert  18687 2020-06-17 14:55:22.092285032 +0200 configure.ac
-rw-r--r--  1 xbert xbert  21209 2020-06-17 14:55:22.092285032 +0200 CMakeLists.txt
drwxr-xr-x 30 xbert xbert   4096 2020-06-29 13:44:17.914725302 +0200 ..
drwxr-xr-x  2 xbert xbert   4096 2020-07-11 22:43:15.753194472 +0200 m4
drwxr-xr-x  2 xbert xbert   4096 2020-07-11 22:43:17.465229289 +0200 config
-rw-r--r--  1 xbert xbert  53655 2020-07-11 22:43:25.561393938 +0200 aclocal.m4
drwxr-xr-x  2 xbert xbert   4096 2020-07-11 22:43:26.137405651 +0200 autom4te.cache
-rw-r--r--  1 xbert xbert 447972 2020-07-11 22:43:26.865420457 +0200 Makefile.in
drwxr-xr-x  2 xbert xbert   4096 2020-07-11 22:43:26.893421027 +0200 doc
drwxr-xr-x  3 xbert xbert   4096 2020-07-11 22:43:26.913421433 +0200 java
drwxr-xr-x  4 xbert xbert   4096 2020-07-11 22:43:27.113425501 +0200 tessdata
drwxr-xr-x  6 xbert xbert   4096 2020-07-11 22:43:27.309429486 +0200 unittest
-rwxr-xr-x  1 xbert xbert 671032 2020-07-11 22:43:27.617435751 +0200 configure
drwxr-xr-x 18 xbert xbert   4096 2020-07-11 22:43:27.617435751 +0200 .
```

> As you can see, `Makefile.in` was made after `configure` and `autom4te.cache`

Yes, from what `autogen.sh` (or `autoreconf`) do, this should be the correct result state. The oddity is on my system!

But then, why not just replace `tesseract/configure` (both as a prerequisite to `$(BIN)/tesseract` and as a target for the `autogen` recipe) with `Makefile.in`, the last file created?

Makefile
stweil: It is 100 % reproducible with n jobs, 1 < n < 7: I currently cannot run parallel builds (tested on several platforms) although I used to do that some (long) time ago.

Today it was the first try with `make all -j` (no limit on parallel jobs), and I was really surprised to see that it did not show an error, but did not build anything either.

stweil:
> But then, why not just replace tesseract/configure (both as a prerequisite to $(BIN)/tesseract and as a target for the autogen recipe) with Makefile.in, the last file created?

I replaced it now in a 2nd commit.

Makefile
bertsky: > It is 100 % reproducible with n jobs, 1 < n < 7: I currently cannot run parallel builds (tested on several platforms) although I used to do that some (long) time ago.

I can reproduce that behaviour. It's a race for the lock file, alright. We need to explicate this to `make`.

> Today it was the first try with `make all -j` (no limit on parallel jobs), and I was really surprised to see that it did not show an error, but did not build anything either.

I always use it this way, and I can confirm this still works. (If it did not build anything for you, then only because everything was already up to date.)

I guess when there is no limit on the number of jobs there is something of an extra check what the current load-level is, which helps avoid the race.

Anyway, let's discuss this in a separate issue!

Makefile
```diff
@@ -459,11 +459,11 @@ $(TESSDATA)/%.traineddata:
 	$(call WGET,$@,$(TESSDATA_URL)/raw/master/$(notdir $(call stripdir,$@))/$(notdir $@)) || \
 		{ $(RM) $@; false; }
 
-tesseract/configure:
+tesseract/Makefile.in:
```
bertsky, line 462:
```suggestion
tesseract/Makefile.in: tesseract
```

Makefile
```diff
@@ -459,11 +459,11 @@ $(TESSDATA)/%.traineddata:
...
 	cd tesseract && ./autogen.sh
 
 # Build and install Tesseract.
-$(BIN)/tesseract: tesseract/configure tesseract
+$(BIN)/tesseract: tesseract/Makefile.in tesseract
```
bertsky, line 466:
```suggestion
$(BIN)/tesseract: tesseract/Makefile.in
```

Makefile
bertsky: > Anyway, let's discuss this in a separate issue!

https://github.com/OCR-D/ocrd_all/issues/123

Makefile
stweil: I'd prefer a solution which restores `tesseract` (instead of `install-tesseract`) as a target to build the `tesseract` executable (similar to other executables). The suggested change would prevent that.

bertsky: > I'd prefer a solution which restores `tesseract` (instead of `install-tesseract`) as a target to build the `tesseract` executable (similar to other executables).

That's not possible. The submodule itself is already called `tesseract`. You'd have to rename it for that.

What's so bad about `install-tesseract`?

> The suggested change would prevent that.

The suggested change is unrelated to that.

stweil: > That's not possible. The submodule itself is already called tesseract.

Sure, but why should that be a problem. You already claimed in commit  e1b30b9 that `tesseract` cannot be both PHONY and concrete, but did not explain why. I did not notice a problem. What exactly was the problem why that commit was needed?

> What's so bad about install-tesseract?

It's not bad. But we offer abbreviated forms for most executables, then why not for tesseract?

> The suggested change is unrelated to that.

Adding `tesseract` as a requirement for `tesseract/Makefile.in` would not work if `tesseract` were PHONY. And I still don't see why that requirement should be needed.

bertsky: > > That's not possible. The submodule itself is already called tesseract.
> 
> Sure, but why should that be a problem.

Because makefiles are rule systems, and you cannot have two different rules for the same target.

> You already claimed in commit [e1b30b9](https://github.com/OCR-D/ocrd_all/commit/e1b30b951bf0beef835954cb75051a8b43aa4b81) that `tesseract` cannot be both PHONY and concrete, but did not explain why

Then you should have asked back then, or just read the `make` manual.

> I did not notice a problem. 

The problem was there nevertheless.

> What exactly was the problem why that commit was needed?

What I already stated above: you cannot have `tesseract` as both an file/executable and a directory/submodule – regardless of how it is modelled in make.

This discussion gets us nowhere. If you don't want to understand the syntax yourself, then why insist on something that _does not work_ as I have shown you above?

> > What's so bad about install-tesseract?
> 
> It's not bad. But we offer abbreviated forms for most executables, then why not for tesseract?

For the very reason stated above. (You'll also notice that none of the other modules have an executable of the same name.)

> > The suggested change is unrelated to that.
> 
> Adding `tesseract` as a requirement for `tesseract/Makefile.in` would not work if `tesseract` were PHONY. And I still don't see why that requirement should be needed.

This is so unproductive! I **already explained and demonstrated** to you twice that without that dependency – which you try to remove in this PR and my suggestion merely re-instates after the substitution `s/configure/Makefile.in/` – `make` simply does not know where to find `autogen.sh`. The file depends on the module being cloned/updated, it's as simple as that. (And I have no idea why you start being concerned that `tesseract` is PHONY. It's a concrete directory as far as `make` is concerned.)

bertsky:
Despite the recent activity (force-push), still not resolved: re-instating the actual module dependency:
* https://github.com/OCR-D/ocrd_all/pull/121#discussion_r453645810
* https://github.com/OCR-D/ocrd_all/pull/121#discussion_r453645989

(Remember, the _only_ problem was that `configure` was not early enough, it had to be `Makefile.in`.)

bertsky (approved):

Issue #123, opened 2020-07-13
avoid race for git lockfile when syncing modules in parallel


bertsky:
There's a race condition for the `.git/config.lock` file between parallel jobs when they want to synchronize / update some submodules. It must have been around for some time already, and only surfaces when using `make -j N` (i.e. with some fixed number of jobs), but not `make -j` (i.e. relative to the current system load-level).

_Originally posted by @bertsky in https://github.com/OCR-D/ocrd_all/pull/121_

bertsky:
It's again unfortunate that all which GNU make offers here is global breakdown of concurrency via `.NOTPARALLEL`, where we instead want only local serialization.

For a synchronization mechanism with blocking behaviour on the shell, I can only think of using GNU parallel (which we already install under `deps-ubuntu` when `workflow-configuration` is not disabled). This would allow for example:
```make
$(OCRD_MODULES): always-update
	sem --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
		sem --id ocrd_all_git git submodule update --init $(GIT_RECURSIVE) $@ && \
		touch $@; fi
```

@stweil @kba opinions?

stweil:
IMO pull request #124 is a simply and sufficient solution for this issue.

bertsky:
> IMO pull request #124 is a simply and sufficient solution for this issue.

#124 would sacrifice the independence of builds for different modules, using `modules` as a single dependency for all executables, and would require downloading all modules even when (re-)installing only a single executable.

The `sem` solution is simple, too. It's also just a tiny change.

I would have started a PR for it already, but I still have difficulty reliably reproducing the above problem. The race seems to be more related to CPU load than network load. I want to be sure to understand the actual cause before rushing to conclusions.

PR #124, opened 2020-07-14
Change rule for submodules update (fix issue #123)

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,13 +100,20 @@ help: ;	@eval "$$HELP"
 # - then updates the time stamps of the module directories
 #   so each directory can be used as a dependency
 $(OCRD_MODULES): modules
+RECURSIVE_MODULES = ocrd_fileformat ocrd_olena opencv-python
 modules:
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
-	git submodule sync $(GIT_RECURSIVE)
-	for module in $(OCRD_MODULES); do \
+	git submodule sync --recursive
+	for module in $(filter-out $(RECURSIVE_MODULES),$(OCRD_MODULES)); do \
 	    if git submodule status $(GIT_RECURSIVE) $$module | grep -qv '^ '; then \
 		git submodule update --init $(GIT_RECURSIVE) $$module && \
+		touch $$module; \
+	    fi; \
+	done
+	for module in $(filter $(RECURSIVE_MODULES),$(OCRD_MODULES)); do \
+	    if git submodule status --recursive $$module | grep -qv '^ '; then \
+		git submodule update --init --recursive $$module && \
 		touch $$module; \
 	    fi; \
 	done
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -99,6 +99,7 @@ help: ;	@eval "$$HELP"
 #   unless they are already up-to-date
 # - then updates the time stamps of the module directories
 #   so each directory can be used as a dependency
+$(OCRD_MODULES): modules
```
bertsky, line 102: This introduces a circular dependency, because we already have `modules: $(OCRD_MODULES)`

stweil:
@bertsky, @kba, the PR now passes CI and should be ready for merging.

stweil:
> it sacrifices the ability to build modules independent of each other

Building modules was not changed. The pull request only changes the way how the submodules are synchronized and updated. This results in a changed behaviour for some less common scenarios: running `make ocrd` not only synchronizes and updates the `core` submodule, but also all other submodules according to `OCRD_MODULES`. By setting that macro it is possible to enforce the old behaviour.

bertsky:
> > it sacrifices the ability to build modules independent of each other
> 
> Building modules was not changed. The pull request only changes the way how the submodules are synchronized and updated.

It throws them all together into a single rule (named `modules`). This can slow down building severely, especially for modules with large tails. Parallel build should start using the CPU as soon as the network transfer is ready for the first modules. (Not all submodule actions need a lock file and thus have to be synchronized.)

> This results in a changed behaviour for some less common scenarios: running `make ocrd` for example not only synchronizes and updates the `core` submodule, but also all other submodules according to `OCRD_MODULES`.

True! That's another minus for this approach IMO.

Makefile
```diff
@@ -93,20 +93,30 @@ export HELP
...
-#   unless it is already up-to-date
-# - then updates the time stamp of the module directory
-#   so the directory can be used as a dependency
-modules: $(OCRD_MODULES)
+# - updates the submodules to the revision registered here,
+#   unless they are already up-to-date
+# - then updates the time stamps of the module directories
+#   so each directory can be used as a dependency
+$(OCRD_MODULES): modules
+RECURSIVE_MODULES = ocrd_fileformat ocrd_olena opencv-python
```
bertsky, line 103: Defining this in a central place undermines our principle of encapsulating all modules' knowledge together. You could easily use `RECURSIVE_MODULES += ocrd_fileformat` (asf.) style.

Makefile
```diff
@@ -93,20 +93,30 @@ export HELP
...
+	    if git submodule status $(GIT_RECURSIVE) $$module | grep -qv '^ '; then \
+		git submodule update --init $(GIT_RECURSIVE) $$module && \
+		touch $$module; \
+	    fi; \
+	done
+	for module in $(filter $(RECURSIVE_MODULES),$(OCRD_MODULES)); do \
+	    if git submodule status --recursive $$module | grep -qv '^ '; then \
+		git submodule update --init --recursive $$module && \
+		touch $$module; \
+	    fi; \
```
bertsky, line 118: Why split the recipe into two parts, copying all commands? Copying always introduces the danger of someone forgetting to update the other accordingly.

You can easily use some `$(if ...)` within the command to discern your `RECURSIVE_MODULES` from the others.

Makefile
kba: Eek, that cannot be good. @stweil?

bertsky: @stweil already avoided that by also removing the `modules: $(OCRD_MODULES)` dependency and force-pushing. But I stand by my other criticism (both on how this is done and on the general approach) and still recommend for #125

stweil:
This pull request was superseded by the merged pull request #125, so it can be closed.

kba:
thanks @stweil, I appreciate your continued contributions!

PR #125, opened 2020-07-15
synchronize concurrent git submodule actions via mutex

### LINKED ISSUE #123: avoid race for git lockfile when syncing modules in parallel

bertsky:
Fixes #123 

@stweil please test this, too!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ endif
 export PKG_CONFIG_PATH
 
 OCRD_EXECUTABLES = $(BIN)/ocrd # add more CLIs below
-CUSTOM_DEPS = unzip wget python3-venv parallel # add more packages for deps-ubuntu below (or modules as preqrequisites)
+CUSTOM_DEPS = unzip wget python3-venv parallel git # add more packages for deps-ubuntu below (or modules as preqrequisites)
 
 DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_kraken clstm ocrd_ocropy
 # Default to all submodules, but allow overriding by user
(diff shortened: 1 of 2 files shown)


bertsky:
@kba CI fails because GNU `parallel` is already needed for `deps-ubuntu` now (just like `make` and `git` are needed a priori). Do you think I should add a rule to CircleCI (before `setup_remote_docker`)?

kba:
> Do you think I should add a rule to CircleCI (before setup_remote_docker)?

:+1: That is what I would do. Before/after setup_remote_docker as you please but before `run: make docker-maximum`.

bertsky:
> That is what I would do. Before/after setup_remote_docker as you please but before `run: make docker-maximum`.

Ok, but until the new rule is in place, we won't see whether it works :frowning: 

Alternatively, I could re-organize `deps-ubuntu` such that it first installs all packages for `ocrd_all` itself, and only then goes to look at submodules' `deps-ubuntu`...

bertsky:
> Alternatively, I could re-organize `deps-ubuntu` such that it first installs all packages for `ocrd_all` itself, and only then goes to look at submodules' `deps-ubuntu`...

That worked. Except for `make` itself, everything can be installed via `make deps-ubuntu` now.

kba:
@stweil I cannot reproduce the `git` lock file race condition issue. I notice no overhead from using parallel's `sem` here but does it fix #123 for you?

What I **did** notice was that `make modules` in this branch will not update opencv-python but it does so in master. Any idea why that is?

bertsky:
> What I **did** notice was that `make modules` in this branch will not update opencv-python but it does so in master. Any idea why that is?

In a fresh clone, `opencv-python` is under `DISABLED_MODULES`.

> I cannot reproduce the `git` lock file race condition issue.

Try doing `make -j4 all MODULES="core sbb_textline_detector ocrd_anybaseocr ocrd_tesserocr"` (perhaps repeatedly). You will soon get an `error: could not lock config file .git/config` message with a non-zero retval from make.

kba:
> `make -j4 all MODULES="core sbb_textline_detector ocrd_anybaseocr ocrd_tesserocr"`

That did the trick in master after several tries. Can confirm that with this fix, no more "lock file exists" messages no matter how often I try. Thanks, merging.

PR #137, opened 2020-08-04
make all SKIP_MODULES=1 to not update modules

kba:


kba:
> Yes, it would be useful for development (e.g. running make after changing submodules but before committing) to have such an option.
> 
> But IMHO `SKIP_MODULES` is not a good name: It sounds like `DISABLED_MODULES`. How about `NO_UPDATES=1`?

I thought `SKIP_MODULES` because it "skips" the `make modules` target. But `NO_UPDATES` is fine with me.

Makefile
```diff
@@ -108,11 +111,13 @@ modules: $(OCRD_MODULES)
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
+ifneq ($(NO_UPDATE),1)
```
stweil, line 114: `ifeq ($NO_UPDATE),0)` would also work with `NO_UPDATE=yes` or even `NO_UPDATE=anything_not_0`. Do you also plan to add information to `README.md`?

bertsky: > Do you also plan to add information to `README.md`?

I wouldn't recommend advertising it too much. After all, it's a developer option...

kba: > `ifeq ($NO_UPDATE),0)` would also work with `NO_UPDATE=yes` or even `NO_UPDATE=anything_not_0`. Do you also plan to add information to `README.md`?

You're right, that would make it easier to use but as @bertsky says, this is not something we want users to "try if anything else fails", so I think it's best not to document in the README.md (perhaps in the contributor guide?).

Issue #140, opened 2020-08-04
sem timing out


kba:
@mnoelte reports that `make modules` fails with timeouts from `sem`:

```
parallel: Warning: Semaphore stuck for 30 seconds. Consider using --semaphoretimeout.
```

I'm not able to reproduce the issue, open to any suggestions. Might be a permission issue.

However, since the semaphore file locks are only required when make is called with `j n` and `n > 1`, a simple workaround would be to check whether more than a single job is to be run and to remove the `sem` locks if `n == 1`.

kba:
Confirmed with @mnoelte that the problem was indeed the wrong permissions in `$HOME/.parallel`.

The problem probably arose from him running `sudo make deps-ubuntu` as the first command, which depends on `modules` and creates the `$HOME/.parallel` directory with the wrong permissions.

We should document that as a first step, users should either call `parallel --citation` (#139) or `make modules` which should both create the `$HOME/.parallel` directory with the right permissions.

bertsky:
Independent of whether and how we [document](#139) that problem, we should also fix our `deps-ubuntu` rule. Here's my suggestion (again):

> IMO we should run that command (piping `will cite` to stdin) as part of the `custom-deps-ubuntu` recipe as an automatic workaround. Not sure about non-Debian/Ubuntu users though (so yes, documentation might be needed anyway).

PR #141, opened 2020-08-04
workaround for #140

kba:
A trivial workaround that will call `git submodule` without `sem` if `-j` was not given or was 1.


kba:
Not necessary anymore, root cause for #140 identified.

Issue #139, opened 2020-08-04
Document `parallel --citation`


kba:
Users should run `parallel --will-cite` or they will be flooded with spam about citing the paper that GNU parallel is based on.

mnoelte:
Or they should run `parallel --citation` and type `will cite` when asked for!?

bertsky:
IMO we should run that command (piping `will cite` to stdin) as part of the `custom-deps-ubuntu` recipe as an automatic workaround. Not sure about non-Debian/Ubuntu users though (so yes, documentation might be needed anyway).

Sorry about all that BTW. If I had considered that problem earlier, I would not have suggested using `parallel` for our deployment in the first place.

kba:
I think the easiest solution is to touch `$HOME/.parallel/will-cite`.

bertsky:
> I think the easiest solution is to touch `$HOME/.parallel/will-cite`.

But what if `parallel` changes this implementation detail?

The advantage of merely touching the config file is that we can use the same assumption to work around our `sudo` context in native installations, e.g. by `chown -R --reference=$(CURDIR) .parallel` (as in the existing recipe after recursive `deps-ubuntu`).

Other alleys could be using the sudo-defined environment variables `SUDO_UID` and `SUDO_GID` to switch back to the original user.

bertsky:
> > I think the easiest solution is to touch `$HOME/.parallel/will-cite`.
> 
> But what if `parallel` changes this implementation detail?
> 
> The advantage of merely touching the config file is that we can use the same assumption to work around our `sudo` context in native installations, e.g. by `chown -R --reference=$(CURDIR) .parallel` (as in the existing recipe after recursive `deps-ubuntu`).

An additional problem here is that – depending on sudoers policy – `HOME` might or might not get set to the target user by `sudo`. Since we need `parallel` both in priviledged (`sudo make deps-ubuntu` submodules) and non-priviledged (`make all` submodules) context, we have to "register" it for both users. But we cannot even be sure where their HOME is (without knowing the sudoers policy).

So we have no choice but to use the functional approach, i.e. `echo will-cite | parallel --citation`. And since we _also_ cannot know for sure when `sudo` is used, let's just do this **every time** we want to use `parallel` or `sem` (in a canned recipe, like our `WGET` wrapper).

kba:
[...]

PR #145, opened 2020-08-13
Fix sub-venv

### LINKED ISSUE #140: sem timing out

### LINKED ISSUE #144: sub-venv logic ineffective when main venv was already active
When the user already has a virtual environment set up for herself and activated, ocrd_all picks it up (does not create a new one). But under this circumstance, creating sub-venvs does not work as expected: `python -m venv DIR` does **not** create new `pip` / `pip3` entry points then, which causes subsequent `pip` calls to use the main venv again.

This does _not_ happen when using `virtualenv` proper to create the sub-venv – which makes me wonder why we moved back from `python3-virtualenv` to `python3-venv` in the first place. From `virtualenv`'s [manual](https://virtualenv.pypa.io/en/latest/):
> The `venv` module does not offer all features of this library, to name just a few more prominent:
> ...
> - is not upgrade-able via `pip`,

This does also _not_ happen when the main venv gets deactivated before the sub-venv is created, even with the `venv` module.

So which of these 2 options is preferable as a fix?

bertsky:
Fixes #140 
Fixes #144

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,9 +84,9 @@ RUN ldconfig
 
 # remove (dated) security workaround preventing use of
 # ImageMagick's convert on PDF/PS/EPS/XPS:
-RUN sed -i 's/rights="none"/rights="read|write"/g' /etc/ImageMagick-6/policy.xml
+RUN sed -i 's/rights="none"/rights="read|write"/g' /etc/ImageMagick-6/policy.xml || true
 # prevent cache resources exhausted errors
-RUN sed -i 's/name="disk" value="1GiB"/name="disk" value="4GiB"/g' /etc/ImageMagick-6/policy.xml
+RUN sed -i 's/name="disk" value="1GiB"/name="disk" value="8GiB"/g' /etc/ImageMagick-6/policy.xml || true
 
 # reset to interactive
 ENV DEBIAN_FRONTEND teletype
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ ENV VIRTUAL_ENV $PREFIX
 ENV DEBIAN_FRONTEND noninteractive
 
 # make apt system functional
-RUN apt-get update \
+RUN apt-get -y update \
  && apt-get install -y apt-utils
 
 # allow passing build-time parameter for list of tools to be installed
--- a/Makefile
+++ b/Makefile
@@ -286,7 +286,7 @@ $(BIN)/ocrd-olena-binarize: ocrd_olena
 clean: clean-olena
 .PHONY: clean-olena
 clean-olena:
-	$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
+	-$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
 endif
 
 ifneq ($(findstring dinglehopper, $(OCRD_MODULES)),)
--- a/Makefile
+++ b/Makefile
@@ -286,7 +286,8 @@ $(BIN)/ocrd-olena-binarize: ocrd_olena
 clean: clean-olena
 .PHONY: clean-olena
 clean-olena:
-	-$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
+	test ! -f ocrd_olena/Makefile || \
+	$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
 endif
 
 ifneq ($(findstring dinglehopper, $(OCRD_MODULES)),)
--- a/Makefile
+++ b/Makefile
@@ -706,9 +706,9 @@ endif
 # install CUSTOM_DEPS first (which is required for the module updates)
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	@chown -R --reference=$(CURDIR) .git $^
+	chown -R --reference=$(CURDIR) .git $^
 # prevent the sem commands during above module updates from imposing sudo perms on HOME:
-	@chown -R --reference=$(HOME) $(HOME)/.parallel
+	chown -R --reference=$(HOME) $(HOME)/.parallel
 
 custom-deps-ubuntu:
 	apt-get -y update
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ export PIP
 # (the option --editable must always be last and only applies to src install)
 PIP_OPTIONS_E = $(filter-out -e,$(PIP_OPTIONS))
 # Set to 1 to skip all submodule updates. For development.
-SKIP_MODULES = 0
+NO_UPDATE = 0
 GIT_RECURSIVE = # --recursive
 # Required and optional Tesseract models.
 ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
@@ -89,6 +89,7 @@ Variables:
 	PIP: name of the Python packaging binary
 	PIP_OPTIONS: extra options for the `pip install` command like `-q` or `-v` or `-e`
 	GIT_RECURSIVE: set to `--recursive` to checkout/update all submodules recursively
+	NO_UPDATE: set to `1` to omit git submodule sync and update
 	OCRD_MODULES: list of submodules to include (defaults to all git submodules, see `show`)
 	DISABLED_MODULES: list of disabled modules. Default: $(DISABLED_MODULES)
 	TESSERACT_MODELS: list of additional models/languages to download for Tesseract
@@ -110,7 +111,7 @@ modules: $(OCRD_MODULES)
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
-ifneq ($(SKIP_MODULES),1)
+ifneq ($(NO_UPDATE),1)
 	sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
 		sem --fg --id ocrd_all_git git submodule update --init $(GIT_RECURSIVE) $@ && \
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ endif
 clean: # add more prerequisites for clean below
 	$(RM) -r $(SUB_VENV)/*
 	$(RM) -r $(CURDIR)/venv # deliberately not using VIRTUAL_ENV here
+	$(RM) -r $(HOME)/.parallel/semaphores/id-ocrd_all_git/
 
 define HELP
 cat <<"EOF"
--- a/Makefile
+++ b/Makefile
@@ -4,11 +4,9 @@
 -include local.mk
 
 # Python version (python3 required).
-PYTHON ?= python3
+export PYTHON ?= python3
 # Python packaging
-PIP ?= pip3
-export PYTHON
-export PIP
+export PIP ?= pip3
 # PIP_OPTIONS ?= # empty
 # Derived variable to allow filtering -e, or inserting other options
 # (the option --editable must always be last and only applies to src install)
@@ -23,17 +21,19 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
 # (but re-use if already active); overriden
 # to nested venv in recursive calls for modules
 # that have known dependency clashes with others
-VIRTUAL_ENV ?= $(CURDIR)/venv
+export VIRTUAL_ENV ?= $(CURDIR)/venv
 SUB_VENV = $(VIRTUAL_ENV)/local/sub-venv
 
 BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
-SEM = sem --nn --fg --id ocrd_all_git
+define SEM
+$(if $(shell sem --version),sem --nn --fg --id ocrd_all_git,$(error cannot find package GNU parallel))
+endef
 
 define WGET
-$(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
+$(if $(shell wget --version),wget -nv -O $(1) $(2),$(if $(shell curl --version),curl -L -o $(1) $(2),$(error found no cmdline downloader (wget/curl))))
 endef
 
 ifeq ($(PKG_CONFIG_PATH),)
@@ -120,7 +120,6 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 ifneq ($(NO_UPDATE),1)
 $(OCRD_MODULES): always-update
-	$(SEM) --version >/dev/null 2>&1 || { echo "cannot find package GNU parallel" >&2; false; }
 	$(SEM) git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
 		$(SEM) git submodule update --init $(GIT_RECURSIVE) $@ && \
@@ -154,7 +153,7 @@ $(SHARE)/numpy: | $(ACTIVATE_VENV) $(SHARE)
 
 .PHONY: ocrd
 ocrd: $(BIN)/ocrd
-deps-ubuntu: core
+deps-ubuntu-modules: core
 $(BIN)/ocrd: core
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< install PIP_INSTALL="$(PIP) install --force-reinstall $(PIP_OPTIONS)"
 	# workaround for core#351:
@@ -224,7 +223,7 @@ endif
 endif
 
 ifneq ($(findstring cor-asv-fst, $(OCRD_MODULES)),)
-deps-ubuntu: cor-asv-fst
+deps-ubuntu-modules: cor-asv-fst
 OCRD_EXECUTABLES += $(OCRD_COR_ASV_FST)
 OCRD_COR_ASV_FST := $(BIN)/ocrd-cor-asv-fst-process
 OCRD_COR_ASV_FST += $(BIN)/cor-asv-fst-train
@@ -254,7 +253,7 @@ endif
 endif
 
 ifneq ($(findstring ocrd_im6convert, $(OCRD_MODULES)),)
-deps-ubuntu: ocrd_im6convert
+deps-ubuntu-modules: ocrd_im6convert
 OCRD_EXECUTABLES += $(BIN)/ocrd-im6convert
 $(BIN)/ocrd-im6convert: ocrd_im6convert
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< install
@@ -280,7 +279,7 @@ endif
 
 ifneq ($(findstring ocrd_olena, $(OCRD_MODULES)),)
 ocrd_olena: GIT_RECURSIVE = --recursive
-deps-ubuntu: ocrd_olena
+deps-ubuntu-modules: ocrd_olena
 OCRD_EXECUTABLES += $(BIN)/ocrd-olena-binarize
 $(BIN)/ocrd-olena-binarize: ocrd_olena
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< install BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
@@ -322,7 +321,7 @@ ifneq ($(findstring ocrd_tesserocr, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_TESSEROCR)
 # only add custom PPA when not building tesseract from source
 ifeq ($(findstring tesseract, $(OCRD_MODULES)),)
-deps-ubuntu: ocrd_tesserocr
+deps-ubuntu-modules: ocrd_tesserocr
 # convert Tesseract model names into Ubuntu/Debian pkg names
 # (does not work with names under script/ though)
 CUSTOM_DEPS += $(filter-out tesseract-ocr-equ,$(subst _,-,$(ALL_TESSERACT_MODELS:%=tesseract-ocr-%)))
@@ -361,7 +360,7 @@ $(call multirule,$(OCRD_CIS)): ocrd_cis
 endif
 
 ifneq ($(findstring ocrd_pagetopdf, $(OCRD_MODULES)),)
-deps-ubuntu: ocrd_pagetopdf
+deps-ubuntu-modules: ocrd_pagetopdf
 OCRD_EXECUTABLES += $(OCRD_PAGETOPDF)
 OCRD_PAGETOPDF := $(BIN)/ocrd-pagetopdf
 $(OCRD_PAGETOPDF): ocrd_pagetopdf
@@ -451,7 +450,7 @@ $(OCRD_REPAIR_INCONSISTENCIES): ocrd_repair_inconsistencies
 endif
 
 ifneq ($(findstring workflow-configuration, $(OCRD_MODULES)),)
-deps-ubuntu: workflow-configuration
+deps-ubuntu-modules: workflow-configuration
 OCRD_EXECUTABLES += $(WORKFLOW_CONFIGURATION)
 WORKFLOW_CONFIGURATION := $(BIN)/ocrd-make
 WORKFLOW_CONFIGURATION += $(BIN)/ocrd-import
@@ -503,9 +502,6 @@ $(foreach executable,$(1),$(file >$(executable),$(call delegator,$(executable)))
 chmod +x $(1)
 endef
 
-# export VIRTUAL_ENV etc to recursive make
-.EXPORT_ALL_VARIABLES:
-
 # avoid making these .PHONY so they do not have to be repeated:
 # clstm tesserocr
 $(SHARE)/%: % | $(ACTIVATE_VENV) $(SHARE)
@@ -534,7 +530,7 @@ $(OCRD_EXECUTABLES:$(BIN)/%=%): %: $(BIN)/%
 ifneq ($(findstring tesseract, $(OCRD_MODULES)),)
 # Tesseract.
 
-# when not installing via PPA, we must cope without ocrd_tesserocr's deps-ubuntu
+# when not installing via PPA, we must cope without ocrd_tesserocr's deps-ubuntu-modules
 CUSTOM_DEPS += g++ make automake libleptonica-dev
 # but since we are building statically, we need more (static) libs at build time
 CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
@@ -703,18 +699,19 @@ endif
 # (mainly intended for docker, not recommended to use directly for live systems)
 # reset ownership of submodules to that of ocrd_all
 # (in case deps-ubuntu has been used with sudo and some modules were updated)
-# install CUSTOM_DEPS first (which is required for the module updates)
-deps-ubuntu: | custom-deps-ubuntu
-	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
+# install git and parallel first (which is required for the module updates)
+deps-ubuntu:
+	apt-get -y install git parallel
+	$(MAKE) deps-ubuntu-modules
 	chown -R --reference=$(CURDIR) .git $^
 # prevent the sem commands during above module updates from imposing sudo perms on HOME:
 	chown -R --reference=$(HOME) $(HOME)/.parallel
 
-custom-deps-ubuntu:
-	apt-get -y update
+deps-ubuntu-modules:
+	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
 	apt-get -y install $(CUSTOM_DEPS)
 
-.PHONY: deps-ubuntu custom-deps-ubuntu
+.PHONY: deps-ubuntu deps-ubuntu-modules
 
 # Docker builds.
 DOCKER_TAG ?= ocrd/all
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,8 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+SEM = sem --nn --fg --id ocrd_all_git
+
 define WGET
 $(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
 endef
@@ -112,9 +114,10 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
 ifneq ($(NO_UPDATE),1)
-	sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
+	sem --version >/dev/null 2>&1 || { echo "cannot find package GNU parallel" >&2; false; }
+	$(SEM) git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
-		sem --fg --id ocrd_all_git git submodule update --init $(GIT_RECURSIVE) $@ && \
+		$(SEM) git submodule update --init $(GIT_RECURSIVE) $@ && \
 		touch $@; fi
 endif
 endif
@@ -697,7 +700,9 @@ endif
 # install CUSTOM_DEPS first (which is required for the module updates)
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	chown -R --reference=$(CURDIR) .git $^
+	@chown -R --reference=$(CURDIR) .git $^
+# prevent the sem commands during above module updates from imposing sudo perms on HOME:
+	@chown -R --reference=$(HOME) $(HOME)/.parallel
 
 custom-deps-ubuntu:
 	apt-get -y update
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ export PIP
 # Derived variable to allow filtering -e, or inserting other options
 # (the option --editable must always be last and only applies to src install)
 PIP_OPTIONS_E = $(filter-out -e,$(PIP_OPTIONS))
+# Set to 1 to skip all submodule updates. For development.
+SKIP_MODULES = 0
 GIT_RECURSIVE = # --recursive
 # Required and optional Tesseract models.
 ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
@@ -108,10 +110,12 @@ modules: $(OCRD_MODULES)
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
+ifneq ($(SKIP_MODULES),1)
 	sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
 		sem --fg --id ocrd_all_git git submodule update --init $(GIT_RECURSIVE) $@ && \
 		touch $@; fi
+endif
 endif
 
 deinit: clean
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -107,6 +107,8 @@ help: ;	@eval "$$HELP"
 modules: $(OCRD_MODULES)
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
+# workaround for parallel#47076 in sudo and non-sudo contexts:
+$(shell echo will-cite | parallel --citation >/dev/null 2>&1; chown -R --reference=$(HOME) $(HOME)/.parallel)
```
stweil, line 111: This would break builds on macOS:

    usage: chown [-fhnv] [-R [-H | -L | -P]] owner[:group] file ...
           chown [-fhnv] [-R [-H | -L | -P]] :group file ...

Makefile
```diff
@@ -123,7 +125,7 @@ deinit:
 # Get Python modules.
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
-	$(PYTHON) -m venv $(VIRTUAL_ENV)
+	if type deactivate >/dev/null 2>&1; then deactivate; fi; $(PYTHON) -m venv $(VIRTUAL_ENV)
```
stweil, line 128: I'd prefer running `deactivate` just before entering a recursive `make`. Then we know that there exists `$(VIRTUAL_ENV)/bin/deactivate`.

Makefile
bertsky: Damn it. But then ocrd_all **already** cannot work on macOS, because we use the same recipe in the `deps-ubuntu` recipe.

So perhaps (in both instances) we should use `chown -R $(stat --format "%u:%g" FILE) TARGETS`. Would that work on macOS?

Makefile
bertsky: > I'd prefer running `deactivate` just before entering a recursive `make`. Then we know that there exists `$(VIRTUAL_ENV)/bin/deactivate`.

There's no `$(VIRTUAL_ENV)/bin/deactivate` (it's only a function defined in `$(VIRTUAL_ENV)/bin/activate`). 

But you're right, we could (activate and) deactivate before entering recursive make. However, that would further complicate the recipes.

Is there any particular reason for not having `deactivate` up there that you had in mind?

Makefile
stweil: > Damn it. But then ocrd_all **already** cannot work on macOS, because we use the same recipe in the `deps-ubuntu` recipe.

`deps-ubuntu` won't work and will never be needed for macOS for obvious reasons.

stweil: > So perhaps (in both instances) we should use `chown -R $(stat --format "%u:%g" FILE) TARGETS`. Would that work on macOS?

`stat --format` is not supported, but the short form `stat -f` works on macOS, too:

    usage: stat [-FlLnqrsx] [-f format] [-t timefmt] [file ...]

stweil: There remains the question whether it is okay to set "will cite" automatically in a script. As I already said my answer is no.

bertsky: > but the short form `stat -f` works on macOS, too:

That's inconsistent with GNU stat again, which uses `-c` as shorthand (`-f` is short for `--file-system`). 

What's the portable answer to this?

Makefile
```diff
@@ -108,7 +108,7 @@ modules: $(OCRD_MODULES)
 # but bypass updates if we have no repo here (e.g. Docker build)
 ifneq (,$(wildcard .git))
 # workaround for parallel#47076 in sudo and non-sudo contexts:
-$(shell echo will-cite | parallel --citation >/dev/null 2>&1; chown -R --reference=$(HOME) $(HOME)/.parallel)
+$(shell echo will-cite | parallel --citation >/dev/null 2>&1; chown -R `stat -c %u:%g $(HOME)` $(HOME)/.parallel)
```
stweil: Using backticks in shell scripts is deprecated.
And `stat -c` is unsupported on macOS.

bertsky: > Using backticks in shell scripts is deprecated.

It's still POSIX though. And we are in a makefile, so backticks help avoid less readable `$$(...)` syntax.

> And `stat -c` is unsupported on macOS.

Let's keep discussing this in the original thread.

Makefile
bertsky: > What's the portable answer to this?

Probably not using `stat` at all. How about `ls -ldn | { read perm size user group rest; echo $user:$group; }` – does _that_ work under BSD/macOS, too?

bertsky: > There remains the question whether it is okay to set "will cite" automatically in a script. As I already said my answer is no.

As I [already said in the respective thread](https://github.com/OCR-D/ocrd_all/issues/139#issuecomment-673408513) GPL's answer is _yes_, and fundamentally so.

bertsky: > Probably not using `stat` at all. How about `ls -ldn | { read perm size user group rest; echo $user:$group; }` – does _that_ work under BSD/macOS, too?

@stweil please have a look at the current, `ls` based replacement for `chown --reference` – does it work on macOS?

Makefile
kba: Since it's a shell function, only `type deactivate` will work. Couldn't we do that as part of the `ACTIVATE_VENV` variable:

```make
ACTIVATE_VENV = deactivate; $(VIRTUAL_ENV)/bin/activate
```
?

Makefile
```diff
@@ -692,11 +693,16 @@ endif
...
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	chown -R --reference=$(CURDIR) .git $^
+	@chown -R `ls -ldn $(CURDIR) | { read perm size user group rest; echo $$user:$$group; }` .git $^
 
 custom-deps-ubuntu:
 	apt-get -y update
 	apt-get -y install $(CUSTOM_DEPS)
+# workaround for parallel#47076 in sudo and non-sudo contexts:
+	@echo will cite | parallel --citation >/dev/null 2>&1
```
kba, line 702: I have no qualms about circumventing the (GPL-incompatible) change per se but I wonder whether it is easier to just instruct users to do `sudo rm -rf ~/.parallel;echo 'will cite'|parallel --citation` on their own?

Makefile
bertsky: > Couldn't we do that as part of the `ACTIVATE_VENV` variable:
> 
> ```makefile
> ACTIVATE_VENV = deactivate; $(VIRTUAL_ENV)/bin/activate
> ```
> 
> ?

No. For two reasons:
- we use `ACTIVATE_VENV` in a `source` context. You cannot source a literal command sequence, only a file.
- the problem arose with outside venvs already activate before running `make`, which would not be affected by this.

Makefile
bertsky: Some people read documentation (more or less), some don't. Those users which use `deps-ubuntu` should get the best automatic solution. This does not rule out documentation of course, as I have [said before](https://github.com/OCR-D/ocrd_all/issues/140#issuecomment-671916924).

bertsky:
703735d will also fix #151 (in the sense of offering a way out of stale/inconsistent state of `~/.parallel/semaphores`; the original issue was duplicate of #140).

bertsky:
@stweil @kba (after CI shows green) can we merge this?

Makefile
stweil: Let me repeat my preference with a fixed argument:

I'd prefer running deactivate just before entering a recursive make. Then we know that we are able to activate and deactivate the parent venv, and `deactivate` runs in the correct context (not with a different value of `VIRTUAL_ENV`).

stweil:
> @stweil @kba (as soon as CI shows green) can we merge this?

This is again a pull request which mixes different fixes. That makes reviews and merging really difficult.

Makefile
bertsky: > I'd prefer running deactivate just before entering a recursive make.

The problem with that was it would require adding the `deactivate` check/call to every instance of recursive `MAKE` call, which makes the recipes less readable.

> Then we know that we are able to activate and deactivate the parent venv

What do you mean by that? (We never know for sure whether the parent venv actually _is_ active already until we look up the `deactivate` function.)

> and `deactivate` runs in the correct context (not with a different value of `VIRTUAL_ENV`).

Granted, that is a _potential_ problem. But venvs created by `virtualenv` and the `venv` module have so far never relied on `VIRTUAL_ENV` during `deactivate` (and why should they?) – on the contrary, they merely `unset VIRTUAL_ENV`.

stweil: > What do you mean by that? (We never know for sure whether the parent venv actually _is_ active already until we look up the `deactivate` function.)

The top level make can activate its venv before running deactivate and the recursive make, no need to look up `deactivate` as it must the defined after `activate`.

bertsky: > The top level make can activate its venv before running deactivate and the recursive make, no need to look up `deactivate` as it must the defined after `activate`. And like that it does not matter whether there was a venv activated outside of make.

Ah, so you refer to the `. $(ACTIVATE_VENV); deactivate` pattern. But there's nothing to learn from the fact that we can activate the parent venv at that point, because we just created it (successfully), or it was already active (in which case activation is a no-op).

So to sum up (in my view) the lasting arguments for deactivating before entering recursively:
- pro: would give up the assumption that `deactivate` will (continue to) work even after `VIRTUAL_ENV` has been reassigned
- con: bloats the rules of all modules which need sub-venvs

Example:
https://github.com/OCR-D/ocrd_all/blob/adb5209f95d0b1499ebc91de948f63bc498d4c38/Makefile#L207-L215

…would have to become…

```make
$(call multirule,$(OCRD_COR_ASV_ANN)): cor-asv-ann
ifeq (0,$(MAKELEVEL))
	. $(ACTIVATE_VENV) && deactivate && $(MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_ANN))
	$(call delegate_venv,$(OCRD_COR_ASV_ANN))
$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
else
	$(pip_install)
endif
endif
```

If only I knew a way to reduce the amount and complexity of the recursive invocation (makelevel check, conditional, target-specific variable override, call recursively, call canned recipe for shell script wrapper)... then the choice would be easier.

@kba what's your preference?

stweil: It's always the same pattern, so maybe the modified line could be rewritten to be even smaller than the current code?

stweil: Would this work?

    SUB_VENV_MAKE := . $(ACTIVATE_VENV) && deactivate && $(MAKE)
    [...]
    ifeq (0,$(MAKELEVEL))
        $(SUB_VENV_MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_ANN))

bertsky: > Would this work?
> 
> ```
> SUB_VENV_MAKE := . $(ACTIVATE_VENV) && deactivate && $(MAKE)
> [...]
> ifeq (0,$(MAKELEVEL))
>     $(SUB_VENV_MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_ANN))
> ```

Yes, it would.

bertsky: >     * would give up the assumption that `deactivate` will (continue to) work even after `VIRTUAL_ENV` has been reassigned

BTW, it's the other way round if we switch to deactivation before recursive call: We assume that `activate` will (continue to) work even after `activate` has been called already. (The `virtualenv` and `venv` implementations both internally run `deactivate nondestructive` as a first step, so this does work.) Not that I recommend against it – just saying that we merely trade one assumption on future venv implementation details with another.

kba: > Would this work?
> 
> ```
> SUB_VENV_MAKE := . $(ACTIVATE_VENV) && deactivate && $(MAKE)
> [...]
> ifeq (0,$(MAKELEVEL))
>     $(SUB_VENV_MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_ANN))
> ```

That would be a good compromise IMHO.

bertsky: > That would be a good compromise IMHO.

But is it worth the change? Mind you:

> just saying that we merely trade one assumption on future venv implementation details with another.

We are not gaining anything AFAICS, but sacrifice some comprehensibility in our rules. 

Unless there are still other, new arguments?

kba: The original intention with the `deactivate` call at this point in the makefile was to fix https://github.com/OCR-D/ocrd_all/issues/144, right? What issue doees the deactivate/activate dance for recursive $(MAKE) calls fix? I have no problem with adding a bit of complexity in the, rather formulaic anyway, recursive recipes if that prevents future problems but IIUC @bertsky that wouldn't be the case here.

Makefile
```diff
@@ -697,11 +704,16 @@ endif
 # install CUSTOM_DEPS first (which is required for the module updates)
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	chown -R --reference=$(CURDIR) .git $^
+	@chown -R `ls -ldn $(CURDIR) | { read perm size user group rest; echo $$user:$$group; }` .git $^
```
kba, line 707: Should we set `SHELL = /bin/bash` to make sure this works across platforms?

Makefile
bertsky: > The original intention with the `deactivate` call at this point in the makefile was to fix #144, right?

Correct. But then @stweil argued that it would be even better to deactivate _before_ entering the recursive make. Both solutions _work_ currently, so this is largely a discussion about which might not under unknown, changed circumstances.

I have presented the analysis that _some_ assumptions on venv's `activate` or `deactivate` implementation are necessary either way.

> if that prevents future problems but IIUC @bertsky that wouldn't be the case here.

Exactly – it just trades one possible future problem with another, similar one.

Makefile
bertsky: I don't see bashisms here. But maybe there are elsewhere. On the other hand, we would have detected this, because bash behaves like POSIX sh when called as such by make (which is the default).

kba: I thought the `{ .. }` construct wasn't POSIX sh, I stand corrected.

Makefile
stweil: I disagree with this solution because I think it is not the right one. Citing the [GNU parallel FAQ](https://git.savannah.gnu.org/cgit/parallel.git/tree/doc/citation-notice-faq.txt):

    > Do automated scripts break if the notice is not silenced?

    No. Not a single time has that happened. This is due to the notice
    only being printed, if the output is to the screen - not if the output
    is to a file or a pipe.

So the message can just be ignored, or it could be sent to `/dev/null`.

Makefile
stweil: > The original intention with the deactivate call at this point in the makefile was to fix #144, right?

Good point. Why don't I get the issue described in #144? Up to now, I never missed the deactivate call, although I mostly run `make all` with an activated venv.

Makefile
```diff
@@ -697,11 +704,16 @@ endif
...
 
 custom-deps-ubuntu:
 	apt-get -y update
 	apt-get -y install $(CUSTOM_DEPS)
+# workaround for parallel#47076 in sudo and non-sudo contexts:
+	@echo will cite | parallel --citation >/dev/null 2>&1
+# prevent the last cmd imposing sudo perms on HOME:
+	@chown -R `ls -ldn $(HOME) | { read perm size user group rest; echo $$user:$$group; }` $(HOME)/.parallel
+
```
stweil, line 716: That part is not needed with the alternate solutions for the citation issue.

Makefile
bertsky: > I disagree with this solution because I think it is not the right one. Citing the [GNU parallel FAQ](https://git.savannah.gnu.org/cgit/parallel.git/tree/doc/citation-notice-faq.txt):
> 
> ```
> > Do automated scripts break if the notice is not silenced?
> 
> No. Not a single time has that happened. This is due to the notice
> only being printed, if the output is to the screen - not if the output
> is to a file or a pipe.
> ```
> 
> So the message can just be ignored, or it could be sent to `/dev/null`.

Let's untangle again:
```
parallel: Warning: Semaphore stuck for 30 seconds. Consider using --semaphoretimeout.
```
This was our failure in #140. It turned out to be caused by `~/.parallel` being created with sudo permissions (and subsequently using it in non-priviledged context).
We then not only fixed the permissions in a follow-up line, but addressed the problem with the citation notices, which just confuse users.
We did not want to suppress output in general, only the notice specifically.

Makefile
stweil: > We assume that activate will (continue to) work even after activate has been called already.

We already activate our venv without checking whether a venv was activated before, so that is a common practice.

Makefile
bertsky: But yes, the (undocumented) solution `--no-notice` which you [dug up](https://github.com/OCR-D/ocrd_all/pull/155#issuecomment-675472138) would be even better!

bertsky: (Or, it turns out, just `--nn`)

Makefile
bertsky: The `chown` part is needed independent of the citation warning (see above). So let's use a variable `SEM` encapsulating  `sem --nn --fg --id ocrd_all_git` (as requested by @kba), but keep the last line in `deps-ubuntu` (as we don't know whether or not `sudo` sets `HOME` or not in the host's configuration).

stweil: I only was speaking about the `chown` part for `$(HOME)/.parallel`.

stweil: Ah, sorry, my fault. Yes, that part is still required because not only `will-cite` is written there but also the semaphore information.

Makefile
bertsky: > Why don't I get the issue described in #144? Up to now, I never missed the deactivate call, although I mostly run `make all` with an activated venv.

So, as I've said, I cannot reproduce it myself anymore. I'll drop this and rebase the PR.

Makefile
```diff
@@ -681,7 +691,7 @@ $(TESSTRAIN_EXECUTABLES:$(BIN)/%=%): %: $(BIN)/%
 clean: clean-tesseract
 .PHONY: clean-tesseract
 clean-tesseract:
-	$(RM) -r $(VIRTUAL_ENV)/build/tesseract
+	-$(RM) -r $(VIRTUAL_ENV)/build/tesseract
```
stweil, line 694: This change is not needed.

Makefile
```diff
@@ -697,11 +707,14 @@ endif
 # install CUSTOM_DEPS first (which is required for the module updates)
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	chown -R --reference=$(CURDIR) .git $^
+	@chown -R `ls -ldn $(CURDIR) | { read perm size user group rest; echo $$user:$$group; }` .git $^
```
stweil, line 710: This change is not needed.

Makefile
```diff
@@ -697,11 +707,14 @@ endif
...
 
 custom-deps-ubuntu:
 	apt-get -y update
 	apt-get -y install $(CUSTOM_DEPS)
+# prevent the last cmd imposing sudo perms on HOME:
+	@chown -R `ls -ldn $(HOME) | { read perm size user group rest; echo $$user:$$group; }` $(HOME)/.parallel
```
stweil, line 716: As we are running on Debian or Ubuntu, the shorter old syntax could also be used. I'd prefer that because it is better readable.

Makefile
```diff
@@ -112,18 +122,18 @@ modules: $(OCRD_MODULES)
...
 		touch $@; fi
 endif
 endif
 
 deinit: clean
 .PHONY: deinit
 deinit:
-	git submodule deinit --all # --force
-	git submodule status | while read stat dir ver; do rmdir $$dir; done
+	-git submodule deinit --all # --force
```
stweil, line 135: Why do you want to ignore errors here? Wouldn't it be better to fail if the `deinit`target cannot deinit all submodules?

Makefile
bertsky: > This change is not needed.

You mean because non-Ubuntu users won't call this anyway?

Makefile
bertsky: Because not all modules might be checked out (initialized) before. git will still return an error in that case.

Makefile
bertsky: It is: `make clean` (which depends on `clean-tesseract`) should work, regardless of whether Tesseract has already been built. This is common practice. (No-one expects `clean` to fail if there was nothing to clean.)

Makefile
stweil: Well, I call it on Debian, too, from time to time.

I just expect that any host which provides `apt-get` also has a recent `chown`. And running `make deps-ubuntu` without `apt-get` (for example on macOS) will fail anyway. It does not matter whether it fails here or some lines later, does it?

Makefile
stweil: I cannot confirm that. Deinit only fails for me when a submodule has modified contents, and for that case that's the desired behaviour.

Makefile
stweil: The old code does not fail for me. As long as `RM` is defined to be `rm -f` I would be surprised if it failed for you.

Dockerfile
```diff
@@ -84,9 +84,9 @@ RUN ldconfig
 
 # remove (dated) security workaround preventing use of
 # ImageMagick's convert on PDF/PS/EPS/XPS:
-RUN sed -i 's/rights="none"/rights="read|write"/g' /etc/ImageMagick-6/policy.xml
+RUN sed -i 's/rights="none"/rights="read|write"/g' /etc/ImageMagick-6/policy.xml || true
```
stweil, line 87:
```suggestion
RUN test -f /etc/ImageMagick-6/policy.xml && sed -i 's/rights="none"/rights="read|write"/g' /etc/ImageMagick-6/policy.xml
```

stweil: Otherwise unexpected errors from `sed` would not be noticed.

Dockerfile
```diff
@@ -84,9 +84,9 @@ RUN ldconfig
...
 # prevent cache resources exhausted errors
-RUN sed -i 's/name="disk" value="1GiB"/name="disk" value="4GiB"/g' /etc/ImageMagick-6/policy.xml
+RUN sed -i 's/name="disk" value="1GiB"/name="disk" value="8GiB"/g' /etc/ImageMagick-6/policy.xml || true
```
stweil, line 89: Dto.

Makefile
```diff
@@ -53,13 +57,19 @@ ifeq ($(strip $(OCRD_MODULES)),)
...
+# delete file targets (e.g. executables / venvs) when recipes fail
+# to ensure they are re-built (not considered up-to-date) when re-entering
+.DELETE_ON_ERROR:
 
 .PHONY: all modules clean help show always-update
 
 clean: # add more prerequisites for clean below
-	$(RM) -r $(SUB_VENV)/*
-	$(RM) -r $(CURDIR)/venv # deliberately not using VIRTUAL_ENV here
+	-$(RM) -r $(SUB_VENV)/*
```
stweil, line 70: No need to ignore errors from `$(RM)` which usually is `rm -f` and does not fail if there is nothing to remove.

Makefile
```diff
@@ -278,7 +288,7 @@ $(BIN)/ocrd-olena-binarize: ocrd_olena
 clean: clean-olena
 .PHONY: clean-olena
 clean-olena:
-	$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
+	-$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
```
stweil, line 291:
```suggestion
	test ! -f ocrd_olena/Makefile || \
	$(MAKE) -C ocrd_olena clean-olena BUILD_DIR=$(VIRTUAL_ENV)/build/ocrd_olena
```

Makefile
stweil, line 694:
```suggestion
	$(RM) -r $(VIRTUAL_ENV)/build/tesseract
```

Makefile
```diff
@@ -112,18 +122,18 @@ modules: $(OCRD_MODULES)
...
 endif
 endif
 
 deinit: clean
 .PHONY: deinit
 deinit:
-	git submodule deinit --all # --force
-	git submodule status | while read stat dir ver; do rmdir $$dir; done
+	-git submodule deinit --all # --force
+	-git submodule status | while read stat dir ver; do rmdir $$dir; done
```
stweil, line 136: Do you have a test case which requires this modification?

stweil (changes requested):
Please update the PR to make it ready for merging:

- Update my commits from PR #153 to fix a misplaced argument and to fix Circle CI.
- Don't ignore failing commands. Some of the related changes are not needed, and for others it is better to improve the command than to ignore errors.
- Use the old `chown` variant for `deps-ubuntu` because portability it not an issue for that target and the old variant is shorter and better readable.

Makefile
bertsky: Yeah, I agree `deinit` should fail under these conditions.

Dockerfile
bertsky: > Otherwise unexpected errors from `sed` would not be noticed.

Which errors do you expect? The _only_ thing that can go wrong here is if the file is not present, or not readable/writable – all of which is covered by the original change. It's also more concise, showing that this step should never break the build.

Makefile
bertsky: > No need to ignore errors from `$(RM)` which usually is `rm -f` and does not fail if there is nothing to remove.

Not if you simply don't have the necessary priviledges. But you might argue that it _does_ make sense to let `clean` fail under these circumstances.

stweil: When I run `make clean`, I expect a certain outcome, namely a defined set of files which is ready for a fresh build. So yes, `make clean` should fail if it is not able to do that.

Dockerfile
stweil: The only acceptable condition is that the file is missing. Then there is no need to patch it of course.

If there is a problem with access permissions, that's a problem which should abort the process, because the code here fails to patch the file although that is required.

If it does not matter whether the existing file is patched or not, I suggest to remove the whole line.

Makefile
kba: It's not necessary to ignore errors from `rm -rf` because it will virtually never throw one. But it doesn't hurt either, making clear the intent to "remove no matter what" which fits the target.

Makefile
kba: Since the target is ubuntu/debian specific, a portable solution, though meritful per se, is not sensible here. Unless there's a reason beyond portability for the new construct.

Dockerfile
kba: > 
> 
> > Otherwise unexpected errors from `sed` would not be noticed.
> 
> Which errors do you expect? The _only_ thing that can go wrong here is if the file is not present, or not readable/writable – all of which is covered by the original change. It's also more concise, showing that this step should never break the build.

Why would the file not be present in the Docker image? We know exactly which base image we build upon, so we can be sure whether the image magick config files exist. If they do not, because the image changed e.g., then we should probably take notice by the build failing. So unless I overlook something, I'm against hiding such errors.

bertsky: > Why would the file not be present in the Docker image? We know exactly which base image we build upon, so we can be sure whether the image magick config files exist. If they do not, because the image changed e.g., then we should probably take notice by the build failing. So unless I overlook something, I'm against hiding such errors.

Because not all module selections depend on IM. For example, if you exclude `ocrd_olena`, Docker will probably not have ImageMagick. (That Dockerfile is not only used for pre-built configs!)

kba: Oh right, hadn't thought of that. Then `|| true` in those statements makes sense.

bertsky: > The only acceptable condition is that the file is missing. Then there is no need to patch it of course.

This is the case (see above)

> If there is a problem with access permissions, that's a problem which should abort the process, because the code here fails to patch the file although that is required.

Agreed, but in our Docker build we are always priviledged, so this is actually out.

> If it does not matter whether the existing file is patched or not, I suggest to remove the whole line.

It does – IM (even the library) does not work otherwise, due to the overly restrictive defaults. (And I think IM's policy workaround to a security problem is a bad design choice...)

kba:
> * Update my commits from PR #153 to fix a misplaced argument and to fix Circle CI.

CI passes.

> * Don't ignore failing commands. Some of the related changes are not needed, and for others it is better to improve the command than to ignore errors.

This has been resolved. ("Because not all module selections depend on IM. For example, if you exclude ocrd_olena, Docker will probably not have ImageMagick. (That Dockerfile is not only used for pre-built configs!)")

> 
> * Use the old `chown` variant for `deps-ubuntu` because portability it not an issue for that target and the old variant is shorter and better readable.

Done

> 
> * Please consider also my suggestions for the `Dockerfile` changes.

Done (see above)

So @stweil @bertsky anything missing or can we merge?

Makefile
```diff
@@ -702,7 +706,9 @@ endif
 # install CUSTOM_DEPS first (which is required for the module updates)
 deps-ubuntu: | custom-deps-ubuntu
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu; done
-	chown -R --reference=$(CURDIR) .git $^
+	@chown -R --reference=$(CURDIR) .git $^
```
stweil, line 709: @bertsky, why do you want to hide the `chown` commands in the build output?

Makefile
stweil: Still unresolved ...

stweil (changes requested):

Makefile
kba: Yes, let's keep those for debugging build.log.

Makefile
bertsky: The problem still is that if `clean-olena` fails, then `clean` would not happen. IMO by convention the semantics of `clean` never includes a notion of success, only _try to remove anything that might be in the way_.

Makefile
bertsky: done.

bertsky:
> anything missing or can we merge?

There's still a similar problem for [wget](https://github.com/OCR-D/ocrd_all/pull/153#discussion_r473116099)

Makefile
stweil: Then please let us wait until such a failure occurs. Do you have some reference for the convention which I did not know?

README.md
```diff
@@ -404,3 +404,13 @@ _(Solved by maintaining these requirements under `deps-ubuntu` here.)_
...
+
+## Acknowledgments
+
+This software uses GNU parallel.
+GNU Parallel is a general parallelizer to run multiple serial command line
+programs in parallel without changing them.
+
+### Reference
+
+Tange, Ole: GNU Parallel 20200722 ('Privacy Shield'). July 2020. https://doi.org/10.5281/zenodo.3956817
```
stweil, line 416: That's not the latest version. The newer one is better.

bertsky:
Wow.

> [must replace which with command -v in Bourne sh](/OCR-D/ocrd_all/pull/145/commits/94040ca8bc96723feaae8be1e17553bd06b52a08)

This is yet another portability nightmare. We [apparently](https://github.com/OCR-D/ocrd_all/runs/1004210143) cannot rely on `which`, because some cache does not get updated soon enough after `apt install`. Also, they say `which` is actually from old csh, and has never been standardized. The proper way for Bourne shells or POSIX sh is said to be `command -v`, but bash _in POSIX mode_ offers [neither](https://github.com/OCR-D/ocrd_all/runs/1004274175) `command` nor `type`, which is stunning. That aspect is not even covered by this (otherwise exhaustive) [SO thread](https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then) on the subject.

Now what do we do? Crawl out and beg for `SHELL = /usr/bin/env bash` (using `command -v`), or find another tool, or debug the [cache problem](https://github.com/OCR-D/ocrd_all/runs/1004210143)?

bertsky:
> This is yet another portability nightmare. We [apparently](https://github.com/OCR-D/ocrd_all/runs/1004210143) cannot rely on `which`, because some cache does not get updated soon enough after `apt install`. Also, they say `which` is actually from old csh, and has never been standardized. The proper way for Bourne shells or POSIX sh is said to be `command -v`, but bash _in POSIX mode_ offers [neither](https://github.com/OCR-D/ocrd_all/runs/1004274175) `command` nor `type`, which is stunning. That aspect is not even covered by this (otherwise exhaustive) [SO thread](https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then) on the subject.
> 
> Now what do we do? Crawl out and beg for `SHELL = /usr/bin/env bash` (using `command -v`), or find another tool, or debug the [cache problem](https://github.com/OCR-D/ocrd_all/runs/1004210143)?

Oh, got it: For one, `/bin/sh` is not bash on Debian, but dash. But more importantly, `make` itself (regardless of whether via `shell` function or recipe) will decide for itself _not_ to invoke a shell, but search the executable in PATH, if it sees only one instruction. Thus, builtins like `command` don't work.

stweil:
> Thus, builtins like command don't work.

Well, `command` works for me without a hack, see this `Makefile`:

    TEST := $(shell command -v make)
    all:
    	@echo test=$(TEST)

How did you get a different result? If it works for you, too, I suggest to remove all hacks again.

And there is no cache problem with `which`.

stweil:
@bertsky, I now merged most of your commits, especially the urgent bug fixes. Please rebase your PR, so we can discuss the remaining ones easier.

bertsky:
> Well, `command` works for me without a hack, see this `Makefile`:
> 
> ```
> TEST := $(shell command -v make)
> all:
> 	@echo test=$(TEST)
> ```
> 
> How did you get a different result?

You didn't say what your result was ;-)
With `make` version 4.1, I get:
```
make: command: Command not found
test=
```
But if I add a semicolon to the end of the `shell` call, then this becomes:
```
test=/usr/bin/make
```

Anyway, we don't need any `which` or `command` test, we can just `$(shell PROG --version)`. 

Meanwhile, I have also found the actual reason why our `$(error ...)` calls were all evaluated out of context: `.EXPORT_ALL_VARIABLES` breaks the usual expansion conditions of course. So removing this and using `export VIRTUAL_ENV` fixes all our wget/sem problems during `deps-ubuntu`. 

It's even possible to run `deps-ubuntu` without having `git` before now (as the README already claimed).

bertsky:
> I now merged most of your commits, especially the urgent bug fixes. Please rebase your PR, so we can discuss the remaining ones easier.

What the ...

bertsky:
> > I now merged most of your commits, especially the urgent bug fixes. Please rebase your PR, so we can discuss the remaining ones easier.
> 
> What the ...

@stweil you must be joking! You made me undo most of my PR after 3 days back and forth, and then just before I finally come around fixing the last issues, you just drag the commits out of the PR without even asking? I just reset master to the previous state, so we can wait for CI in the _current, fixed_ state here and _then_ merge if everyone is satisfied (as it should be).

kba:
> @bertsky, I now merged most of your commits, especially the urgent bug fixes. Please rebase your PR, so we can discuss the remaining ones easier.

@stweil why did you do that? We were close to finishing the discussions, weren't we?

stweil:
> why did you do that

I had the impression that it was time to fix some issues after a whole week of discussion. So I applied those patches where we agreed that they are fine.

What is the problem with this? Rebasing @bertsky's remaining commits is easy. I have already done in locally and could push that state to a new branch if that helps.

README.md
bertsky: So all I need is adding https://github.com/OCR-D/ocrd_all/pull/153/commits/635068e9fba6f341df47b7cedeefbd3e2f4d0410 right?

stweil: I already added the reference to latest git, so you can remove that commit from your branch.

kba:
> > why did you do that
> 
> I had the impression that it was time to fix some issues after a whole week of discussion. So I applied those patches where we agreed that they are fine.

We are all eager for a fix, but it should be based on consensus and successful CI. Cherry-picking individual commits and pushing them to master without consultation is just not good practice. We all have the power to push to master but we should only use it once we're in agreement. We were close but not yet there.

> 
> What is the problem with this? Rebasing @bertsky's remaining commits is easy. I have already done in locally and could push that state to a new branch if that helps.

Again, it's not a question of technicality but of courtesy. And you are - rightly - an adherent of concise pull requests and static code analysis. Cherry-picking from an almost finished PR with 90+ discussion threads without asking first here or in gitter is not good. I appreciate that the pace of discussions can sometimes be frustrating but finding consensus improves the product in the long-term.

README.md
kba: https://github.com/OCR-D/ocrd_all/pull/145/commits/7a57df7b3fba53e218a395b4d1e28f06dc28d519

kba:
CI succeeded, the last commit only adapted the parallel citation in the README, which will not break building. So thanks a lot @bertsky and @stweil for the continued effort!

bertsky:
> What is the problem with this? Rebasing @bertsky's remaining commits is easy. I have already done in locally and could push that state to a new branch if that helps.

@stweil Also, you forget that rebasing then becomes much harder and a one-way direction. In this concrete case, I had already rebased this PR locally, fixing the early commits regarding the sem and wget failures (which you yourself have been pointing out mostly). We were in the middle of a discussion about these. By dragging them onto master directly, you made it impossible to edit/undo/reorder them. After all, that's what a PR is for: aggregating a set of mutually supporting/consistent commits that have to be considered and developed together until mature enough to integrate at once.

PR #149, opened 2020-08-15
sem: warn if parallel --citation was not run before

### LINKED ISSUE #140: sem timing out

kba:
Replaces the `sem` calls with a macro that will remind users to answer `will cite` to `parallel --citation` if the latter prompts for it.

Fix #140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,12 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
 VIRTUAL_ENV ?= $(CURDIR)/venv
 SUB_VENV = $(VIRTUAL_ENV)/local/sub-venv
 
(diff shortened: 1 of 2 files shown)


bertsky:
I don't understand the reason for this. I already made a fix for #140 in #145. And we wanted to get a workaround for our scripts (which normally won't be attended by users anyway).

kba:
> I don't understand the reason for this. I already made a fix for #140 in #145. And we wanted to get a workaround for our scripts (which normally won't be attended by users anyway).

Worked on the issues in order, didn't see that #140 is adressed in #145 until after I opened this.

PR #153, opened 2020-08-17
Report missing sem instead of failing with error

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,10 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+define SEM
+$(if $(shell which sem),sem --fg --id ocrd_all_git ,$(error "Please install GNU parallel"))
+endef
+
 define WGET
 $(if $(shell which wget),wget -nv -O $(1) $(2),$(if $(shell which curl),curl -L -o $(1) $(2),$(error "found no cmdline downloader (wget/curl)")))
 endef
@@ -117,9 +121,9 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
 ifneq ($(NO_UPDATE),1)
-	sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
+	$(SEM) git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
-		sem --fg --id ocrd_all_git git submodule update --init $(GIT_RECURSIVE) $@ && \
+		$(SEM) git submodule update --init $(GIT_RECURSIVE) $@ && \
 		touch $@; fi
 endif
 endif
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,15 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+ifneq ($(MAKECMDGOALS),)
+ifneq ($(MAKECMDGOALS),deps-ubuntu)
+ifneq ($(MAKECMDGOALS),help)
 define SEM
-$(if $(shell which sem),sem --fg --id ocrd_all_git ,$(error "Please install GNU parallel"))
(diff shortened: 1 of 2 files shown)


stweil:
PR #145 includes a different solution.

The current PR can be extended to work without `sem` (for example for non-parallel make) or with an alternative semaphore implementation.

kba:
I like introducing `$(SEM)` to replace the call, c.f. https://github.com/OCR-D/ocrd_all/pull/149. But I'd prefer it if we agreed on #145.

stweil:
@bertsky, please review. If this PR is okay for you, I suggest to rebase your PR #145 after this one was merged.

stweil:
@kba, I added a commit which fixes the citation issue, so please review.

stweil:
Regarding the GNU parallel citation issue, there remains the question how to respect it. I suggest to add the citation to the README.md here or to the OCR-D website. Which variant is preferred?

bertsky:
> please review. If this PR is okay for you, I suggest to rebase your PR #145 after this one was merged.

Good idea to combine with the `error`. But I am against merging this alone, because (as you see) CI fails here. #145 has a different solution for that. So I'll just cherry-pick your 2 commits here and continue rebasing #145 (so let's see if CI agrees with all that).

bertsky:
> and then let's add the citation GNU parallel's authors crave to the Readme

ok, just what section were you thinking of? Below the link to contributing.md perhaps?

bertsky:
> > please review. If this PR is okay for you, I suggest to rebase your PR #145 after this one was merged.
> 
> Good idea to combine with the `error`

Damnit. No, now I remember there was a good reason _not_ to include the `error` function into the `sem` call unconditionally. I already had this correct in my branch... (get a feeling I am working backwards all day)

Makefile
```diff
@@ -30,6 +30,10 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+define SEM
+$(if $(shell which sem),sem --fg --id --nn ocrd_all_git ,$(error "Please install GNU parallel"))
```
bertsky, line 34: BTW, this does not work. You placed `--nn` between `--id` and the identifier.

But CI also shows this does not work in general. For some reason (I just cannot seem to remember it) the canned recipe gets evaluated even outside of its usage, right in the middle of `custom-deps-ubuntu`. It's probably a `make` bug. In my solution I ended placing the `error` function into the semaphore recipe directly.

bertsky (changes requested):

Makefile
stweil: Ah, sorry. I updated the PR with the right order of the `sem` arguments and will investigate the CI error.

stweil: @bertsky, the CI problem is caused by `make deps-ubuntu` which fails when GNU parallel is missing. Either we must require an installation of GNU parallel before using `make` with ocrd_all, or we must disable lines 33 and 34 in `Makefile` if the target is `deps-ubuntu`.

stweil: CI should be fixed now.

stweil:
@bertsky, @kba, do you suggest more modifications? CI is fixed now, and I think this PR should be applied (and PR #145 rebased then).

stweil:
I now enhanced `README.md` with a reference for GNU parallel, see https://github.com/stweil/ocrd_all/tree/doc#acknowledgments.

Makefile
bertsky: > the CI problem is caused by `make deps-ubuntu` which fails when GNU parallel is missing.

That's what I've just said.

>  Either we must require an installation of GNU parallel before using `make` with ocrd_all, or we must disable lines 33 and 34 in `Makefile` if the target is `deps-ubuntu`.

No target guessing is necessary. There's a third option (illustrated by the state of my PR before I had rebased it on here), which is better: Checking sem availability only when you actually need it. See current #145

stweil: I would not call a test for `help` or `deps-ubuntu` target guessing, and we need that code for the `WGET` check, too (running make on Ubuntu with neither wget nor curl installed will currently always fail). So that's not wasted code. Therefore I decided against your alternative solution.

bertsky: > I would not call a test for `help` or `deps-ubuntu` target guessing

Look, we want to look at the `sem` business this once, and then forget about it. As soon as other targets appear, which are legitimate before doing `deps-ubuntu` (or similar), or if we make something dependent on them, then these will have to be included in your list. Hence _guessing_.

> and we need that code for the `WGET` check, too

oh, right – I didn't notice that before. I'll add a fix to #145 for that along the same lines.

stweil:
This PR can be closed. Some of the commits are merged now, others will be replaced by PR #145.

PR #154, opened 2020-08-18
Don't use sub-venv for ocrd_segment

stweil:
That processors don't use TensorFlow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,13 +305,7 @@ OCRD_SEGMENT += $(BIN)/ocrd-segment-extract-pages
 OCRD_SEGMENT += $(BIN)/ocrd-segment-replace-original
 OCRD_SEGMENT += $(BIN)/ocrd-segment-repair
 $(call multirule,$(OCRD_SEGMENT)): ocrd_segment
-ifeq (0,$(MAKELEVEL))
-	$(MAKE) -B -o $< $(notdir $(OCRD_SEGMENT))
-	$(call delegate_venv,$(OCRD_SEGMENT))
-$(OCRD_SEGMENT): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
-else
 	$(pip_install)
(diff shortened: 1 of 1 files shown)


bertsky:
> That processors don't use TensorFlow.

Oh, but they do (or will in the near future): look at the `maskrcnn-cli` branch... This is still work in progress, but I depend on this branch in other ocrd_all-based deployments.

stweil:
> Oh, but they do (or will in the near future)

That's bad news. Maybe we should split such projects in two, one with processors which don't need a sub-venv, one with processors which need one? Or will all processors depend on TensorFlow in the near future?

Ideally we should not add new processors depending on old versions of TensorFlow ...

bertsky:
> That's bad news. Maybe we should split such projects in two, one with processors which don't need a sub-venv, one with processors which need one? Or will all processors depend on TensorFlow in the near future?

If you look at the README.md (outlining the original plan/idea for this repo), the TF-based MaskRCNN segmentation was always at the core of what we wanted to do here. The other (non-TF) processors were just preliminaries and fall-off which we exposed/exported early on, because they had value in the OCR-D ecosphere independently. I am reluctant to change that, because of the interdependencies. @wrznr what do you think?

> Ideally we should not add new processors depending on old versions of TensorFlow ...

Agreed, but our MaskRCNN implementation has become unmaintained since we started using it, and it's not trivial to make the migration (there are already dozens of competing PRs targeting different TF2 versions in different ways). ocrd_anybaseocr is just one example of that effort (but they decided to just copy the source files into their own repo, so there's no point in follow them). Perhaps if I find some time to look at it, I'll merge and refine on of these PRs into my fork's master...

bertsky:
BTW ocrd_segment's TF branch `maskrcnn-cli` will be based on TF2 soon, too. Closing for reasons explained above.

PR #155, opened 2020-08-18
Silence citation notice from GNU parallel

stweil:
According to the GNU parallel FAQ, the citation notice can be suppressed
in scripts by sending the output to a file or a pipe.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,9 +117,9 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 $(OCRD_MODULES): always-update
 ifneq ($(NO_UPDATE),1)
-	sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@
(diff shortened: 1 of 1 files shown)


stweil:
`sem --fg --id ocrd_all_git git submodule sync $(GIT_RECURSIVE) $@ 2>&1 | tee` would also work and preserve any other output to `STDERR`, but requires `tee`.

stweil:
See https://git.savannah.gnu.org/cgit/parallel.git/tree/doc/citation-notice-faq.txt for the GNU parallel FAQ which explains the background.

I think it would also be good and polite to add the suggested citation somewhere in the OCR-D documentation, for example in `ocrd_all/README.md` and / or on the OCR-D website.

stweil:
`sem --fg --id ocrd_all_git --no-notice git submodule sync $(GIT_RECURSIVE) $@` is another good solution.

stweil:
I added the "best" solution now to PR #153, so closing this one here.

PR #158, opened 2020-08-20
make check: check for parallel/wget/curl, fix #157

### LINKED ISSUE #157: Parallel build terminates with unexpected message when wget and curl are missing
Running `make all -j4` on a fresh clone of ocrd_all terminated with these build messages:

    configure.ac:382: installing 'config/compile'
    configure.ac:86: installing 'config/config.guess'
    configure.ac:86: installing 'config/config.sub'
    configure.ac:27: installing 'config/install-sh'
    configure.ac:27: installing 'config/missing'
    Makefile.am: installing 'config/depcomp'
    parallel-tests: installing 'config/test-driver'

    All done.
    To build the software now, do something like:

    $ ./configure [--enable-debug] [...other options]

The error was caused because neither `curl` nor `wget` was installed. The related error message was hidden somewhere in the middle of the build protocol:

    Makefile:641: *** found no cmdline downloader (wget/curl).  Stop.
    make: *** Waiting for unfinished jobs....

kba:


Makefile
```diff
@@ -62,7 +62,12 @@ endif
 # to ensure they are re-built (not considered up-to-date) when re-entering
 .DELETE_ON_ERROR:
 
-.PHONY: all modules clean help show always-update
+.PHONY: all modules check clean help show always-update
+
+check:
+	@sem --version >/dev/null && echo "ok - parallel installed"
```
stweil, line 68: Then an error would be reported like this:

/bin/sh: 1: sem: not found

```suggestion
.PHONY: all modules check clean help show always-update

check:
	@sem --version >/dev/null 2>&1 || echo "error - please install GNU parallel"
```

stweil: I have no strong opinion regarding the success message, so `ok - parallel installed` could be added for the success case if desired.

bertsky (changes requested):
I don't see the point. You already get appropriate error messages now when one of these tools is missing. No need to duplicate that under a new target. It will only confuse users. Other missing dependencies will produce errors when due, respectively.

Makefile
```diff
@@ -62,7 +62,12 @@ endif
...
+	@{ wget --version || curl --version; } > /dev/null && echo "ok - wget/curl installed"
```
stweil, line 69: As we already have hard dependencies on `wget`, cleaning the current code would be reasonable - no need to support `curl` as well.

stweil (changes requested):
Error messages should be added / improved, and instead of checking for curl or wget, a check for wget only would be better.

Makefile
kba: > I don't see the point. You already get appropriate error messages now when one of these tools is missing. No need to duplicate that under a new target. It will only confuse users. Other missing dependencies will produce errors when due, respectively.

I see this more as a documentation feature. If users do encounter problems, you could ask them: "What does `make check` say?" to quickly exclude missing dependencies. I'd say that is easier than capturing `make all` output and spot the error.

kba: `> As we already have hard dependencies on `wget`, cleaning the current code would be reasonable - no need to support `curl` as well.

Since the `$(WGET)` macro checks `wget --version`, then `curl --version`, `make check` should also check in that order. The `||` should stop early if wget is available and test whether curl is available only if wget isn't.

bertsky: > I see this more as a documentation feature. If users do encounter problems, you could ask them: "What does `make check` say?" to quickly exclude missing dependencies. I'd say that is easier than capturing `make all` output and spot the error.

What do you gain? – Nothing. Whether wget and parallel is missing is already reported by `make all` – at the very bottom (unless running parallel and for the first time).

What do you loose? – All the other possible sources of error. You'll usually have to ask for `make all` output anyway.

stweil: Cleaning the `WGET` macro to only use `wget`is a separate task which can be done before or after this PR (but please in a separate PR). I just wanted to say that there is no need to support `curl` here and to remove that later again.

`make deps-ubuntu` installs `wget`.

cneud:
Just to note that ``curl`` is required for the [OCR-D Research Data Repository](https://ocr-d.de/en/setup#testing-the-installation-1) and since this is included with [taverna_workflow](https://github.com/OCR-D/taverna_workflow), users of that will see their workflows fail when ``curl`` is not installed by ``ocrd_all``.

stweil:
> Just to note that `curl` is required for the [OCR-D Research Data Repository](https://ocr-d.de/en/setup#testing-the-installation-1) and since this is included with [taverna_workflow](https://github.com/OCR-D/taverna_workflow), users of that will see their workflows fail when `curl` is not installed by `ocrd_all`.

`make deps-ubuntu` already installs `wget`. Should it also install `curl` then?

As far as I remember, a default installation of Ubuntu already includes `curl`, so most users will have it. For those who install a minimal Ubuntu adding `curl` to `CUSTOM_DEPS` might be a good idea.

PR #164, opened 2020-08-20
Model dependency of tesserocr on tesseract

stweil:
It is not needed when a pre-installed Tesseract is used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,6 +338,12 @@ OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-segment-region
 OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-segment-word
 $(call multirule,$(OCRD_TESSEROCR)): ocrd_tesserocr $(SHARE)/tesserocr
 	$(pip_install)
+
+# tesserocr must wait for tesseract in parallel builds.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -338,6 +338,12 @@ OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-segment-region
 OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-segment-word
 $(call multirule,$(OCRD_TESSEROCR)): ocrd_tesserocr $(SHARE)/tesserocr
 	$(pip_install)
+
+# tesserocr must wait for tesseract in parallel builds.
```
kba, line 342: Looks reasonable 👍 

Since the bugs you discovered today only appeared with testing `-j` parallelization, perhaps we should add/adapt our pull-request-CI with `-j` calls? @stweil @bertsky

bertsky: > perhaps we should add/adapt our pull-request-CI with `-j` calls?

Yes, running with `-j` is harder and gives us faster CI, so it should suffice to use it exclusively. But not for the `deps-ubuntu` calls of course.

bertsky (approved):

Issue #165, opened 2020-08-20
Missing pip in sub-venv with parallel build


stweil:
`make all -j6` fails because `pip` is missing:

        Uninstalling pip-9.0.1:
        Uninstalling pip-9.0.1:
          Successfully uninstalled pip-9.0.1
    Exception:
    Traceback (most recent call last):
      File "/usr/lib/python3.6/shutil.py", line 544, in move
        os.rename(src, real_dst)
    FileNotFoundError: [Errno 2] No such file or directory: '/root/ocrd_all/venv/local/sub-venv/headless-tf1/bin/pip' -> '/tmp/pip-07jk920k-uninstall/root/ocrd_all/venv/local/sub-venv/headless-tf1/bin/pip'

    During handling of the above exception, another exception occurred:

It looks like one or two processes tried to upgrade pip at the same time.

stweil:
It can be fixed by using the existing semaphores with `$(SEM)`. PR will come soon.

Issue #142, opened 2020-08-11
Upgrade pip to 20.2+


kba:
https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-2-2020

> The most significant changes to the resolver are:

> It will reduce inconsistency: it will no longer install a combination of packages that is mutually inconsistent. In older versions of pip, it is possible for pip to install a package which does not satisfy the declared requirements of another installed package. For example, in pip 20.0, pip install "six<1.12" "virtualenv==20.0.2" does the wrong thing, “successfully” installing six==1.11, even though virtualenv==20.0.2 requires six>=1.12.0,<2 (defined here). The new resolver, instead, outright rejects installing anything if it gets that input.

> It will be stricter - if you ask pip to install two packages with incompatible requirements, it will refuse (rather than installing a broken combination, like it did in previous versions).

> So, if you have been using workarounds to force pip to deal with incompatible or inconsistent requirements combinations, now’s a good time to fix the underlying problem in the packages, because pip will be stricter from here on out.

We should explore that new dependency algorithm and consider upgrading.

bertsky:
True, this must be examined ASAP. I'd expect things will not improve for us unambiguously, but stir up some unresolved problems (which might necessitate downgrading pip for the time being). For example, I recall we could not use `pip check` because of some stupid Tensorflow packaging bugs like [its scipy dependency](https://github.com/OCR-D/ocrd_all/pull/118#issuecomment-660130530) and others.

stweil:
`pip` is upgraded automatically when `make all` creates a new virtual Python environment. So all we have to do is suggest users who create the venv manually or use an older venv to run `pip install --upgrade pip` from time to time.

bertsky:
Another example of unresolved problems that the new pip behaviour will make us stumble over: Our core installation procedure!

There we have a couple of sub-packages which mutually depend on each other. But pip cannot install all of them at once (at least to my knowledge), so our `make install` is a loop over each of them. Each iteration updates a single sub-package but leaves the others in their previous (probably inconsistent) state:
```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

ocrd 2.11.0 requires ocrd-modelfactory==2.11.0, but you'll have ocrd-modelfactory 2.13.0 which is incompatible.
[...]

PR #166, opened 2020-08-20
Avoid parallel creation of virtual environment and pip upgrade (fix i…

stweil:
…ssue #165)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,8 +136,8 @@ deinit:
 # Get Python modules.
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
-	$(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
+	$(SEM) $(PYTHON) -m venv $(VIRTUAL_ENV)
+	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
 
 .PHONY: wheel
 wheel: $(BIN)/wheel


stweil:
PR #164 and PR #166 seem to fix the parallel build issues. I still see an exception, but it looks like all processors are built nevertheless in less than 5 minutes:

    pip._vendor.pkg_resources.ContextualVersionConflict: (ocrd-utils 2.13.0 (/root/ocrd_all/venv/lib/python3.6/site-packages), Requirement.parse('ocrd-utils==2.13.1'), {'ocrd-models'})

Still nagging are the repeated tesseract builds (see PR #121), but they now take only 35 seconds with the parallel build.

Makefile
```diff
@@ -136,8 +136,8 @@ deinit:
 # Get Python modules.
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
-	$(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
+	$(SEM) $(PYTHON) -m venv $(VIRTUAL_ENV)
+	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
```
bertsky, line 140: Good idea to use `sem` for this as well. But we should not mix the semaphore for git submodule action with that for venv modifications. Also, sub-venvs should be different again. And we should not only protect the initial creation, but any `pip install` usage as much as we can.

```suggestion
	$(SEMPIP) $(PYTHON) -m venv $(VIRTUAL_ENV)
	. $(ACTIVATE_VENV) && $(SEMPIP) $(PIP) install --upgrade $(PIP_OPTIONS_E) pip setuptools
```

with `SEMPIP` defined in similar fashion to the `SEM` for git, only with `--id ocrd_all_$(VIRTUAL_ENV)`.

bertsky (changes requested):

Makefile
stweil: I considered that, but think that using new semaphores is overkill because the protected actions are rather short. Typically the module updates will be finished when sub-venvs are created.

stweil: > And we should not only protect the initial creation, but any pip install usage as much as we can.

I don't understand that part. Are there problems with parallel instances of `pip install`? I would have expected that `pip` can handle that. Upgrading `pip` is a very special case.

bertsky: > I considered that, but think that using new semaphores is overkill because the protected actions are rather short. Typically the module updates will be finished when sub-venvs are created.

Don't be so sure. Module checkouts can take a while for large/recursive repos. That's time wasted for others with large Python dependencies (Tensorflow etc). Especially since sub-venvs are completely independent of each other.

The cost of distinguishing `SEMGIT` and `SEMPIP` is minimal, and it maximizes parallel time available.

bertsky: > > And we should not only protect the initial creation, but any pip install usage as much as we can.
> 
> I don't understand that part. Are there problems with parallel instances of `pip install`? I would have expected that `pip` can handle that. Parallel installation is desired, because it can save much time. Upgrading `pip` is a very special case.

Yes, exactly. I don't think pip does file locking. You're right that pip updating itself is more sensitive, but I guess there are potential problems elsewhere, too (esp. after #142 makes pip intolerant of temporary inconsistencies). Have not tried/analysed it though. (And no, never saw `make all -j` fail like that either. But perhaps I have just not tried hard enough.)

stweil: > Typically the module updates will be finished when sub-venvs are created.

> Don't be so sure. Module checkouts can take a while for large/recursive repos. That's time wasted for others with large Python dependencies (Tensorflow etc).

See the new issue #169 for my answer. Yes, we can optimize. But currently introducing new semaphores does not result in faster builds. I just tested that. That might change when #169 is solved. I guess that new semaphores will reduce the build time by less than 5 seconds.

bertsky: Okay. So let's keep this in mind and check again. Yes, probably very unlikely that multiple modules want to install the same dependency at the same time. But even if so, as long as they picked the same version, it's probably going to work...

stweil:
@bertsky, can we merge this (and think about optimization later)?

bertsky:
> can we merge this (and think about optimization later)?

agreed.

PR #170, opened 2020-08-20
Updata submodule core from 2.13.0 to 2.13.1

stweil:
Parallel builds fail when the version in PyPI is newer than the local version.


stweil:
I managed somehow to have both `ocrd_modelfactory-2.13.0` (fully) and `ocrd_modelfactory-2.13.1` (partially) installed by a parallel build. Upgrading `core` here fixes that for now, although that is not the full solution for the future.

bertsky:
> I managed somehow to have both `ocrd_modelfactory-2.13.0` (fully) and `ocrd_modelfactory-2.13.1` (partially) installed by a parallel build. Upgrading `core` here fixes that for now, although that is not the full solution for the future.

Agreed. I already hinted at that in the other thread. This also broke our neck once when only the version here was correct, but had the same release number as the buggy version on PyPI.

It's unfortunate `pip` has no hold mechanism like `apt-mark`.

But we could emulate this if we added a `-c core-version.txt` to all `pip install` invocations, fuelling `core-version.txt` from our `core` module like so:
```make
core-version.txt: ocrd
	. $(ACTIVATE_VENV) && $(PIP) list --format freeze ocrd | grep ^ocrd== > $@

# example of having core as direct dependency to another module
$(call multirule,$(OCRD_WRAP)): ocrd_wrap ocrd
```

stweil:
Using `core-version.txt` sounds like a good idea. Just make sure that the virtual environment is available before that rule is executed.

stweil:
@kba, @bertsky, what about this pull request? Can it be merged, or should we directly update to core 2.13.2? The current situation (ocrd in PyPI newer than core in ocrd_all) must be fixed somehow.

kba:
> @kba, @bertsky, what about this pull request? Can it be merged, or should we directly update to core 2.13.2? The current situation (ocrd in PyPI newer than core in ocrd_all) must be fixed somehow.

I will make a big update with processors adapted to OCR-D/spec#164, if all goes well later today, with core updated to 2.13.2.. I see no harm in upgrading core now if it is urgent. But then upgrade to 2.13.2 not 2.13.1.

stweil:
@kba, if you update it today, that's sufficient for me. Just close this PR when done, please.

PR #176, opened 2020-08-21
Revert "circleci: run make with -j, #164"

kba:
This reverts commit 53e63fc1e1cb3a9e3c890756fad12fa7d31e3591.


stweil:
> And [GitHub suggests to postpone it](https://github.com/github/renaming#later-this-year-seamless-move-for-existing-repositories-), so we can wait and decide later this year.

GitHub no longer suggests to postpone renaming of the default branch. They offer excellent support for the renaming now. Therefore I suggest to do it.

PR #178, opened 2020-08-22
support Docker additional variants based on ocrd/core-cuda

kba:
Once we merge https://github.com/OCR-D/core/pull/454, there will be an additional `ocrd/core-cuda` base image based on [nvidia/cuda:11.0-base-ubuntu18.04](https://hub.docker.com/layers/nvidia/cuda/11.0-base-ubuntu18.04/images/sha256-d2ad87954a40fa4671e81df1fac5989448323bfa846ced05898058b1805b3723?context=explore). We can then create CUDA-capable variants of ocrd_all Docker images.

We'll need to investigate parallelization for the CI/CD because otherwise this will double the build time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,12 @@ jobs:
       - run: make docker-minimum-git
       - run: make docker-medium-git
       - run: make docker-maximum-git
+      - run: make docker-minimum-cuda
+      - run: make docker-medium-cuda
+      - run: make docker-maximum-cuda
+      - run: make docker-minimum-cuda-git
+      - run: make docker-medium-cuda-git
+      - run: make docker-maximum-cuda-git
       - run:
           name: Login to Docker Hub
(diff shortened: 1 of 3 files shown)


kba:
https://hub.docker.com/r/ocrd/core-cuda is now available from DockerHub

stweil (approved):

bertsky (approved):
No time to test this now, but (judging by my core-cuda experience) it _should_ work. Since this does not break anything, we can still revisit if we find any problems.

It would not hurt to _document_ the [extra steps](https://github.com/OCR-D/core/pull/454#issuecomment-678698247) needed to get the nvidia-docker installed and run with `--gpus ...` – both here in the README and in the setup guide.

stweil:
> We'll need to investigate parallelization for the CI/CD because otherwise this will double the build time.

I wonder whether there is a need for all variants of OCR-D docker configurations. One could argue that GPU users are power users which (nearly) always use the maximum variant. Is it possible to get download numbers for the variants without GPU? Maybe medium is also rarely used.

PR #185, opened 2020-09-07
update 2020-09-07

kba:


stweil:
ocrd_fileformat still needs an update, see PR there.

PR #187, opened 2020-09-08
Update core

kba:


stweil (approved):

stweil:
@kba, do we still need [all branches](https://github.com/OCR-D/ocrd_all/branches/all)?

kba:
> @kba, do we still need all branches?

No, I deleted the finished ones. But that list is pretty neat and harmless compared to other projects :)

@stweil @bertsky can we reorder the module processing order in such a way that core is always the first module to be installed to the "main" venv and the "local" venvs? That way, pip never would need to fetch ocrd from PyPI (which broke the CI because I forgot to update the core PyPI release)

bertsky:
> can we reorder the module processing order in such a way that core is always the first module to be installed to the "main" venv and the "local" venvs? That way, pip never would need to fetch ocrd from PyPI (which broke the CI because I forgot to update the core PyPI release)

Yes, it's possible IMO. Makefiles are declarative and thus order is not reliable when building in parallel, unless one introduces explicit dependencies. So we could add an extra prerequisite `$(BIN)/ocrd` to all modules, probably even as order-only (`|`). That should work because all our recipes already distinguish between first (`$<`) and non-first (`$^`) deps. It should even work in the recursive (nested venv) call.

PR #188, opened 2020-09-08
CircleCI deploy parallel push: terminate after last job

bertsky:
(I wonder why this could have ever worked before. I was only alerted to this by [a recent failure](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/371/workflows/1704573c-c4eb-4780-af4c-313d89a1930c/jobs/371), which should not have occured that way either...)

Also, I don't understand why CircleCI's "restart from failed" does not restart from the failed step, but repeats the whole job.

Anyway, be warned I cannot test this outside of the CI, so expect a total breakdown!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,30 +32,18 @@ jobs:
           name: Push images to Docker Hub in parallel
           command: |
             docker push ocrd/all:minimum &
-            jobs=($!)
             docker push ocrd/all:minimum-git &
-            jobs+=($!)
             docker push ocrd/all:medium &
-            jobs+=($!)
             docker push ocrd/all:medium-git &
-            jobs+=($!)
             docker push ocrd/all:maximum &
-            jobs+=($!)
             docker push ocrd/all:maximum-git &
-            jobs+=($!)
             docker push ocrd/all:minimum-cuda &
-            jobs=($!)
             docker push ocrd/all:minimum-cuda-git &
-            jobs+=($!)
             docker push ocrd/all:medium-cuda &
-            jobs+=($!)
             docker push ocrd/all:medium-cuda-git &
-            jobs+=($!)
             docker push ocrd/all:maximum-cuda &
-            jobs+=($!)
             docker push ocrd/all:maximum-cuda-git &
-            jobs+=($!)
-            while sleep 60; ps -p "${jobs[*]}"; do :; done
+            while jobs -pr | { read PIDS && echo $PIDS; }; do sleep 60; done
       - run:
           name: Create a date-versioned mirror of ocrd/all:maximum
           command: bash release.sh release-dockerhub


kba:
Only one way to find out, apparently :)

kba:
> Also, I don't understand why CircleCI's "restart from failed" does not restart from the failed step, but repeats the whole job.

I think for that to work, we'd need to restructure the CI config to not just have `deploy` and `build` but `deploy-minimum-git`, `deploy-maximum-cuda-git` etc. IIUC (and that's a big if) we could then also make use of up to 4 parallel jobs.

grafik

Due to the new CUDA builds, time to deploy is now ~4.5 hours. Thankfully, @stweil merged #187 yesterday, so it ran overnight and succeeded but that is too long and we should aim to reduce it by making the Circle CI config maximize the existing resources or switching CI provider, at least for deployment.

stweil:
I'd reduce the number of Docker builds for GPU because I doubt that all are needed for users of OCR-D.

bertsky:
> > Also, I don't understand why CircleCI's "restart from failed" does not restart from the failed step, but repeats the whole job.
> 
> I think for that to work, we'd need to restructure the CI config to not just have `deploy` and `build` but `deploy-minimum-git`, `deploy-maximum-cuda-git` etc. IIUC (and that's a big if) we could then also make use of up to 4 parallel jobs.

I see. If that's what it takes...

But instead of going parallel, we could also try to rewrite the Docker config in such a way that mini/medi/maxi is not built independent of, but on top of each other. We would avoid overall computation costs (but still have more latency than purely parallel). It would not be difficult, I believe. Maybe even the git variants could be re-used for the non-git builds. (cuda variants would still have to run independently, though.)

kba:
In light of https://github.com/OCR-D/core/issues/595 it's questionable whether an even reduced number of CUDA images are sensible to build at the moment, if they aren't currently usable because of CUDA/tensorflow incompatibilities.

I like the idea of stacking docker builds though, that would save us a lot of build time.

on that note: is there any disadvantage of git over non-git builds? I always use the former for ease of upgrades and testing branches.

stweil:
> is there any disadvantage of git over non-git builds?

I rarely use non-git builds. They have few advantages like using less resources for one-time builds compared with git. But as soon as git is used with `--depth 1`, even that advantages mostly disappear.

bertsky:
> > Anyway, be warned I cannot test this outside of the CI, so expect a total breakdown!
> Only one way to find out, apparently :)

Sorry to say it did in fact not work. Something strange is happening in CircleCI's command shells. Probably no job control or strange buffering rules. I don't feel like digging through lengthy documentation. Maybe we should just use `wait -f` instead of the job waiting/echoing loop, and set `no_output_timeout: 90m` or something...

> I like the idea of stacking docker builds though, that would save us a lot of build time.

I'll try to make a PR.

> on that note: is there any disadvantage of git over non-git builds? I always use the former for ease of upgrades and testing branches.
> I rarely use non-git builds. They have few advantages like using less resources for one-time builds compared with git. But as soon as git is used with `--depth 1`, even that advantages mostly disappear.

Same for me, and yes, size seems to be the only advantage of non-git builds. @stweil IIUC you build your Docker images locally and just strip all history? (That seems like a good idea – we should probably make that the default for our CI builds...)

So @kba are you trying to say we should drop the non-git builds (and just keep the name for compatibility)?

kba:
> Maybe we should just use wait -f instead of the job waiting/echoing loop, and set no_output_timeout: 90m or something...

Fine with me. I'm not keen either on wasting time debugging CircleCI idiosyncrasies. Feel free to adjust the circleci.yml temporarily in such a way that it's less time-consuming to test.

> So @kba are you trying to say we should drop the non-git builds (and just keep the name for compatibility)?

yes, I think image size shouldn't be a priority and `git clone --depth 1` would neutralize this advantage to an extent I see no reason for non-git builds.

PR #191, opened 2020-09-15
Update ocrd_segment and ocrd_tesserocr

stweil:


stweil:
This PR gets the latest fixes from both submodules which are required for our local OCR-D workflow.

Still open: https://github.com/OCR-D/core/issues/602.

bertsky:
We need at least the new core and ocrd_cis, too.

stweil:
> We need at least the new core and ocrd_cis, too.

Thanks. I now added those and also the latest tesseract, tesserocr and workflow-configuration.

bertsky:
> Still open: [OCR-D/core#602](https://github.com/OCR-D/core/issues/602).

@kba has already merged a fix now – please update again.

stweil:
Done.

kba (approved):

stweil:
I just noted that all current tags are not annotated tags. We should change release.sh to create annotated tags in the future.

kba:
> I just noted that all current tags are not annotated tags. We should change release.sh to create annotated tags in the future.

Why?

stweil:
`git describe` only shows annotated tags (unless you add `--tags`). Citing the documentation:

"Git supports two types of tags: lightweight and annotated."
"Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels."

So unless we consider the tags to be private or temporary, annotated tags would be the right choice.

PR #194, opened 2020-09-18
emulate file function in make<4.0

### LINKED ISSUE #192: Build on macOS fails to create delegator scripts for executables in sub-venvs
It tries to run the `chmod`:

    chmod: /Users/stweil/src/github/OCR-D/venv-20200917/bin/ocrd-segment-evaluate: No such file or directory
    [...]

bertsky:
Fix #192

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -498,6 +498,8 @@ endef
 #  the sub-venv does not have the
 #  executable)
 # TODO: variant for relay to Docker
+ifeq ($(firstword $(subst ., ,$(MAKE_VERSION))),4)
+# make >=4 has file function
 define delegator
 #!/bin/bash
 . $(ACTIVATE_VENV) && $(BIN)/$(notdir $(1)) "$$@"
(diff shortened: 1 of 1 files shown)


kba (commented):
It's a good solution but I wonder whether we actually need to support a 10+ year old GNU Make - it must be possible to use more modern versions of make, bash etc. on OSX.

But if we do need to support older GNU make, why not use the 3.8-delegator-workaround for all versions?

bertsky:
> But if we do need to support older GNU make, why not use the 3.8-delegator-workaround for all versions?

Right, we could avoid the solution base on file function entirely, and actually increase readability. But I am still not so sure we are not already relying on other `>=3.82` extensions or semantics elsewhere...

stweil (approved):
The new code works fine with the old make of macOS, thank you.

stweil:
> I am still not so sure we are not already relying on other >=3.82 extensions or semantics elsewhere.

I am sure we would note that - sooner or later.

PR #196, opened 2020-09-23
update-2020-09-23

kba:


stweil:
I get `CRITICAL` errors with that new code:

    18:08:21.706 CRITICAL root - getLogger was called before initLogging. Source of the call:
    18:08:21.706 CRITICAL root -   File "/home/stweil/src/github/OCR-D/venv-20200923/local/sub-venv/headless-tf22/lib/python3.7/site-packages/ocrd_anybaseocr/cli/ocrd_anybaseocr_cropping.py", line 68, in <module>

bertsky:
@stweil Thanks for testing this promptly!

@kba Looks like we'll have to fix all processors first, moving `getLogger` from module level to runtime scope to make sure the decorator's `initLogging` gets called first. Another PR orgy is due – bad timing. But after all, this is just a log message once per processor – we could live with that for now I suppose.

stweil (approved):
I suggest to merge this pull request despite the "critical" log message. My workflow tests pass nevertheless (better than before).

kba:
> Another PR orgy is due – bad timing.

I've sent PRs fixing this to all the python-based processor projects. Ideally, we should be able to merge this and create a new release without these warnings asap.

stweil:
Closed because replaced by PR #197.

kba:
I created a fresh PR so we don't get into trouble with inconsistent versions.

PR #198, opened 2020-09-25
Add $(BIN)/ocrd as order-only target to install first

kba:
<del>This should ensure that core is installed first for every module.</del>

It's ridiculously inefficient, not even sure if this ever finishes :fire:


bertsky:
The CI failure should go away if you update from master:
```
git clone https://github.com/OCR-D/spec "repo/spec"
fatal: destination path 'repo/spec' already exists and is not an empty directory.
```

But we need an _actual_ dependency, not order-only – because it's precisely a forced update which we are after.

bertsky (commented):
This seems to work. (You can see make installing core each time it is in a sub-venv.)

But why do these modules not get the core dependency yet:
- ocrd_im6convert
- ocrd_fileformat
- ocrd_olena
- dinglehopper
- sbb_textline_detector
?

bertsky (approved):

Issue #195, opened 2020-09-23
Build and install core from code in the local venvs


kba:
With the local-venv-mechanism, core is installed once in the "root" venv and once in each venv.

When doing `make all` after an update, only the root core installation is updated IIUC.

So in addition to [moving core to the front of the list of modules to install](https://github.com/OCR-D/ocrd_all/pull/187#issuecomment-688944306) we should make sure, that core is updated consistently across venvs.

bertsky:
> So in addition to [moving core to the front of the list of modules to install](https://github.com/OCR-D/ocrd_all/pull/187#issuecomment-688944306) we should make sure, that core is updated consistently across venvs.

The latter is entailed by the former BTW. (Once we make the dependency on core explicit, it will trigger recursively too.)

kba:
Workaround until this is fixed:

```
git pull
make all
# next steps not necessary once #195 is fixed
(source venv/local/sub-venv/headless-tf1/; cd core; make install)
(source venv/local/sub-venv/headless-tf22/; cd core; make install)
```

VolkerHartmann:
I did this:
git pull
make all
source venv/local/sub-venv/headless-tf1/bin/activate; cd core; make install
deactivate
cd ..
source venv/local/sub-venv/headless-tf22/bin/activate; cd core; make install
deactivate
cd ..
 . venv/bin/activate
-->
ocrd-segment-repair --version
Version 0.1.2, ocrd/core 2.16.3

bertsky:
> Workaround until this is fixed:
> 
> ```
> git pull
> make all
> # next steps not necessary once #195 is fixed
> (source venv/local/sub-venv/headless-tf1/; cd core; make install)
> (source venv/local/sub-venv/headless-tf22/; cd core; make install)
> ```

More generally, I think it is "enough" to force updating all modules:

    git submodule foreach 'touch $path'
    make all

stweil:
> More generally, I think it is "enough" to force updating all modules

No, that does not update core in sub venvs for me.

kba:
> No, that does not update core in sub venvs for me

You mean the workaround does not work for you? Why, it does for me?

stweil:
@kba workaround: works
@bertsky workaround: creates new empty files, does not work

bertsky:
> No, that does not update core in sub venvs for me. Using `git submodule foreach 'touch .'` also does not help.

Sorry – I should have said: `touch -c`

stweil:
[...]

PR #202, opened 2020-10-01
Build core in all sub venvs, too (fix issue #195)

stweil:


bertsky (changes requested):
I would strongly recommend against that pattern. It goes against our encapsulation and still installs core from PyPI in the venvs. It's up to `make` to decide which comes first, the `core` rule which updates the submodules' core in the right way or the submodules' rules which do it wrongly.

#198 was much closer, see [my comment](https://github.com/OCR-D/ocrd_all/pull/198#issuecomment-699044629).

stweil:
@bertsky, if it still installs core from PyPI in the sub venvs, then I expect that will happen in the primary venv, too, and we'd have one more problem.

stweil:
I close this PR because a different fix was implemented now.

PR #207, opened 2020-10-08
Docker prefix path

bertsky:
This removes the irritating situation that all executables from sub-venvs end up in /usr/local/bin whereas the others are in /usr/bin (which also makes it more difficult for dependent stages to add their scripts).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,9 @@ LABEL \
     org.label-schema.build-date=$BUILD_DATE
 
 
-ENV PREFIX=/usr/local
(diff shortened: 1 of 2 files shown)


stweil:
Please make separate pull requests for independent commits. That makes reviews and merging easier.

bertsky:
> Please make separate pull requests for independent commits. That makes reviews and merging easier.

This PR changes **3 lines in total**. That's not too much to ask. Why waste extra CI build time on such trivial, subtle differences like the names of venvs?

kba (approved):

PR #209, opened 2020-10-09
Don't use additional pip options for pip upgrade

stweil:
Such additional options might be unsupported by the old pip version.


stweil:
See https://github.com/OCR-D/ocrd_all/issues/142#issuecomment-705817824.

bertsky (approved):

PR #211, opened 2020-10-14
Make sure that pip always gets upgraded

stweil:
It was not upgraded for an existing venv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ deinit:
 
 # Get Python modules.
 
-$(VIRTUAL_ENV)/bin/pip3: $(ACTIVATE_VENV)
+$(VIRTUAL_ENV)/bin/$(PIP): $(ACTIVATE_VENV)
 	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --upgrade pip setuptools
 
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
@@ -542,7 +542,7 @@ $(SHARE):
 	@mkdir -p "$@"
 
 # At last, add venv dependency (must not become first):
-$(OCRD_EXECUTABLES) $(BIN)/wheel: | $(VIRTUAL_ENV)/bin/pip3
+$(OCRD_EXECUTABLES) $(BIN)/wheel: | $(VIRTUAL_ENV)/bin/$(PIP)
 $(OCRD_EXECUTABLES): | $(BIN)/wheel
 
 # At last, we know what all OCRD_EXECUTABLES are:
--- a/Makefile
+++ b/Makefile
@@ -137,9 +137,11 @@ deinit:
 
 # Get Python modules.
 
+$(VIRTUAL_ENV)/bin/pip3: $(ACTIVATE_VENV)
+	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --upgrade pip setuptools
+
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
 	$(SEM) $(PYTHON) -m venv $(VIRTUAL_ENV)
-	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --upgrade pip setuptools
 
 .PHONY: wheel
 wheel: $(BIN)/wheel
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -137,9 +137,11 @@ deinit:
 
 # Get Python modules.
 
+$(VIRTUAL_ENV)/bin/pip3: $(ACTIVATE_VENV)
```
bertsky, line 140:
```suggestion
$(VIRTUAL_ENV)/bin/$(PIP): $(ACTIVATE_VENV)
```

Makefile
```diff
@@ -540,7 +542,7 @@ $(SHARE):
 	@mkdir -p "$@"
 
 # At last, add venv dependency (must not become first):
-$(OCRD_EXECUTABLES) $(BIN)/wheel: | $(ACTIVATE_VENV)
+$(OCRD_EXECUTABLES) $(BIN)/wheel: | $(VIRTUAL_ENV)/bin/pip3
```
bertsky, line 545:
```suggestion
$(OCRD_EXECUTABLES) $(BIN)/wheel: | $(VIRTUAL_ENV)/bin/$(PIP)
```

bertsky (changes requested):
Why not. But I wonder what your use-case is. We never needed to update from the old 9-something default until recently. Does this have to do with the new dependency resolver that will arrive by default soon?

Makefile
stweil: That would not work if `PIP` includes a path prefix.

bertsky: It was never intended or documented that way. The README and help texts all state `name of the ... binary`, to allow differentiating between different Python versions.

_Not_ respecting that variable here would break those existing use cases.

stweil:
> We never needed to update from the old 9-something default until recently.

Maybe most builds did not use an existing venv but created the venv? The Makefile already upgraded pip when it created a new venv.

Makefile
stweil: A virtual environment does not provide different Python versions, and as far as I know it will always have a `pip3` executable. Which use cases exist which would fail? I never have set `PIP` manually in any of my builds.

bertsky:
> > We never needed to update from the old 9-something default until recently.
> 
> Maybe most builds did not use an existing venv but created the venv? The Makefile already upgraded pip when it created a new venv.

No, creating a venv does **not** update pip automatically. We only noticed this when the old default version (it's `9.0.1` under Ubuntu 18.04) both pre-built wheels in the `manylinux2014` format and fallback installation from source distributions using `pyproject.toml` stopped working.

stweil:
> No, creating a venv does not update pip automatically.

The current Makefile creates the venv in line 141. Line 142 updates pip. I don't understand your no.

bertsky:
> > > We never needed to update from the old 9-something default until recently.
> > 
> > No, creating a venv does not update pip automatically.
> 
> The current Makefile creates the venv in line 141. Line 142 updates pip. I don't understand your no.

You are taking this out of context. The question remains: what do we need this change for in the first place?

stweil:
> The question remains: what do we need this change for in the first place?

See the commit message: pip was not upgraded for an existing venv. So for example this sequence would run with an old version of pip and possibly fail:

    python3.7 -m venv myvenv
    source myvenv/bin/activate
    make all

Makefile
kba: I agree that this is somewhat academical but you might want to ensure it's the pip for the right minor version of Python, e.g. `make all PIP=pip3.7`.

Be that as it may, it's still sensible to adhere to the `$(PIP)` scheme throughout.

stweil: What about removing the `PIP` macro? I simply cannot imagine a real use case, as we always use virtual Python environments. I suggest to replace `$(PIP)` by `pip3` or even by `pip` because that avoids unused complexity.

stweil: Or do we have submodules which use the exported `PIP`?

stweil: @kba, `make all PIP=pip3.7` would work as expected, because the pip upgrade replaces `pip`, `pip3` and `pip3.7`.

kba: When we started OCR-D phase II we had to support Python 2.7, hence you'll find PIP and PYTHON parameterizable in most of the Makefiles I originally set up.

I agree that we can revisit that design choice. But a pull request should do one thing and while simplifying unused complexity is a good idea per se, we should make sure it is actually unused.

PR #213, opened 2020-10-21
update 2020-10-21

kba:


stweil:
ocrd-tesserocr-recognize might be broken with this pull request: Tesseract recently got a contribution to make it faster (tesseract-ocr/tesseract#2994). I have now run a local test. It was neary twice as fast as before, but did not produce usable OCR text. Further analysis has still to be done.

stweil (changes requested):
Please don't merge this pull request with the new tesseract because that breaks `ocrd-tesserocr-recognize`.

stweil (approved):
It's https://github.com/OCR-D/ocrd_tesserocr/pull/158 which had broken my test. PR #213 works for me. Sorry for the noise.

stweil:
@kba, the previous commit 609415b146dc89f24444ba4f594e57127ce4c82c was fine (I already had approved it). I suggest to merge that revision.

PR #214, opened 2020-10-22
add qurator-spk/sbb_binarization

kba:
> Pixelwise binarization with selectional auto-encoders in Keras

@jbarth-ubhd has an example of the result in https://github.com/OCR-D/ocrd-website/issues/172#issuecomment-708416972

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -82,3 +82,6 @@
 [submodule "ocrd_olahd_client"]
 	path = ocrd_olahd_client
 	url = https://github.com/OCR-D/ocrd_olahd_client.git
+[submodule "sbb_binarization"]
+	path = sbb_binarization
+	url = https://github.com/qurator-spk/sbb_binarization
--- a/.gitmodules
+++ b/.gitmodules
@@ -84,4 +84,4 @@
 	url = https://github.com/OCR-D/ocrd_olahd_client.git
 [submodule "sbb_binarization"]
 	path = sbb_binarization
-	url = https://github.com/qurator-spk/sbb_binarization
+	url = https://github.com/qurator-spk/sbb_binarization.git
--- a/Makefile
+++ b/Makefile
@@ -438,6 +438,19 @@ else
 endif
 endif
 
+ifneq ($(findstring sbb_binarization, $(OCRD_MODULES)),)
+OCRD_EXECUTABLES += $(SBB_BINARIZATION)
+SBB_BINARIZATION := $(BIN)/ocrd-sbb-binarize
+$(SBB_BINARIZATION): sbb_binarization $(BIN)/ocrd
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) -B -o $< $(notdir $(SBB_BINARIZATION))
+	$(call delegate_venv,$(SBB_BINARIZATION))
+$(SBB_BINARIZATION): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
+else
+	$(pip_install)
+endif
+endif
+
 ifneq ($(findstring sbb_textline_detector, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(SBB_LINE_DETECTOR)
 SBB_LINE_DETECTOR := $(BIN)/ocrd-sbb-textline-detector
--- a/Makefile
+++ b/Makefile
@@ -589,6 +589,7 @@ stripdir = $(patsubst %/,%,$(dir $(1)))
 install-models: \
 	install-models-tesseract \
 	install-models-ocropus \
+	install-models-sbb-binarization \
 	install-models-calamari
 
 UB_MANNHEIM_BACKUP_URL = https://ub-backup.bib.uni-mannheim.de/~stweil/ocrd-train/data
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -82,3 +82,6 @@
 [submodule "ocrd_olahd_client"]
 	path = ocrd_olahd_client
 	url = https://github.com/OCR-D/ocrd_olahd_client.git
+[submodule "sbb_binarization"]
+	path = sbb_binarization
+	url = https://github.com/qurator-spk/sbb_binarization
```
stweil, line 87:
```suggestion
	url = https://github.com/qurator-spk/sbb_binarization.git
```

Makefile
```diff
@@ -438,6 +438,19 @@ else
...
 endif
 
+ifneq ($(findstring sbb_binarization, $(OCRD_MODULES)),)
+OCRD_EXECUTABLES += $(SBB_BINARIZATION)
+SBB_BINARIZATION := $(BIN)/ocrd-sbb-binarize
+$(SBB_BINARIZATION): sbb_binarization $(BIN)/ocrd
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) -B -o $< $(notdir $(SBB_BINARIZATION))
+	$(call delegate_venv,$(SBB_BINARIZATION))
+$(SBB_BINARIZATION): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
```
stweil, line 448: @cneud, I think it would be good to use the latest TensorFlow for new software instead of old TensorFlow versions which are unsupported with recent version of Python3.

kba: @vahidrezanezhad probably knows best how much effort upgrading keras/tensorflow would be.

cneud: @stweil We had this tool available for a year already, only the OCRD-CLI was added now. We also still use TF1 for the `sbb-textline-detector` and `ocrd-calamari-recognize` in the @qurator-spk team, so I doubt we will be able to migrate quickly.

stweil (approved):
Looks good. Are there already models available for production? Then a rule `install-models-sbb-binarize` to download one or several models would be nice.

Makefile
bertsky: @stweil as discussed elsewhere already, this comes with the price tag of having to re-train all models. It's Google's fault, not the module developer's. And we on our part do have workarounds for this situation (sub-venvs, which could even be of different Python version).

bertsky (approved):
I second @stweil's comment about packaging models. Or can we do this via `package_data` and the upcoming [resource discovery](https://github.com/OCR-D/core/pull/559) feat?

Makefile
stweil: As far as I know the code migration is the smaller problem, migration of existing models is the larger problem. That's why I thought a code migration might be reasonable before putting much efforts in training of new models with old TensorFlow. Luckily we know now how to handle environments with mixed versions of TensorFlow. So it is possible to migrate the different software packages in smaller steps.

kba:
Pre-trained models are available from https://qurator-data.de/sbb_binarization/

I'll add an `install-models` rule for this as well.

As for bundling with the tool, I'll have to finish that in core first but then yes, certainly, they could be placed in one of the defined locations and `model` parameter could be a relative path. For now it must be absolute.

bertsky:
> As for bundling with the tool, I'll have to finish that in core first but then yes, certainly, they could be placed in one of the defined locations 

But the Python package on PyPI should not be polluted with such large models. We could use setuptools' `extra_requires` feature to build packages including them from our makefile.

> and `model` parameter could be a relative path. For now it must be absolute.

Couldn't we add an environment variable (say `SBB_BINARIZATION_MODELS`) in the meantime? Absolute paths make sharing makefile configurations across hosts impossible.

kba:
> Then a rule install-models-sbb-binarize to download one or several models would be nice.

https://github.com/OCR-D/ocrd_all/pull/214/commits/32285fad55d3a7011a560f6f66c4559ee7e5f8d0

> Couldn't we add an environment variable (say SBB_BINARIZATION_MODELS) in the meantime? Absolute paths make sharing makefile configurations across hosts impossible.

Yes but then I'd suggest to implement it in a processor-independent way, see https://github.com/OCR-D/spec/issues/176

bertsky:
> Yes but then I'd suggest to implement it in a processor-independent way, see [OCR-D/spec#176](https://github.com/OCR-D/spec/issues/176)

That's not the same thing, though. Yours is an environment variable for a specific parameter (which is entirely new AFAIK). My suggestion was to follow the example of `TESSDATA_PREFIX`, `OCROPUS_DATA`,  `CORASVANN_DATA` to allow passing a reference point in the search paths for relative filenames. (Not standardized either, of course – but usable already.)

kba:
>  (Not standardized either, of course – but usable already.)

I realize this is more specific but it would not be much harder to implement in core for all the processors to make use of.

I can implement (=steal from cor-asv-ann ;-)) `SBB_BINARIZATION_DATA` in the short term, I'll send a PR.

bertsky:
> I can implement (=steal from cor-asv-ann ;-)) `SBB_BINARIZATION_DATA` in the short term, I'll send a PR.

That would be great. In particular for derived Docker images :-)

kba:
> > I can implement (=steal from cor-asv-ann ;-)) `SBB_BINARIZATION_DATA` in the short term, I'll send a PR.

https://github.com/qurator-spk/sbb_binarization/pull/6

PR #215, opened 2020-10-27
update 2020-10-27

kba:


stweil (approved):

stweil:
The new Tesseract which is included in this pull request can give a performance boost of more than 10 % for `ocrd-tesserocr-recognize` because it tries a 2nd OCR with inverted line images less often.

PR #222, opened 2020-11-27
update 2020-11-27

kba:
<del>Notably, this defines (yet another) new venv for ocrd_calamari since it now requires tensorflow 2.4 and numpy >= 1.19</del>

ocrd_calamari doesn't require 2.4 yet, it uses 2.3 as does ocrd_pc_segmentation, so they can share `venv/local/sub-venv/headless-tf2` for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,105 @@
 # Changelog
 
+## [v2020-11-27](https://github.com/OCR-D/ocrd_all/releases/v2020-11-27)
+
+### [core](https://github.com/OCR-D/core) [079aace](https://github.com/OCR-D/core/commits/079aace)..[2298662](https://github.com/OCR-D/core/commits/2298662)
+
+> Release: [v2.21.0](https://github.com/OCR-D/core/releases/v2.21.0)
+
+  > * :package: v2.21.0
+  > * :memo: changelog
+  > * Merge branch 'profile-cpu-time'
+  > * Merge branch 'numpy-any-version'
+  > * :memo: changelog
+  > * Merge remote-tracking branch 'origin/bertsky-patch-task-sequence'
+  > * :memo: changelog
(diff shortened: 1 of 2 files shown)


bertsky:
> Notably, this defines (yet another) new venv for ocrd_calamari since it now requires tensorflow 2.14 and numpy >= 1.19

Is it really necessary to differentiate with `ocrd_pc_segmentation`'s `tf2` venv?

At least superficially, they could be compatible:

```
tensorflow >= 2.3.0rc2
```

```
tensorflow == 2.3.1
```

Of course, there might be other conflicts, I have not tried it. 

The motivation for this is that each new TF installation
- takes a lot of bandwidth to build
- bloats the Docker image by ~ 2GB

kba:
The problem here is numpy compatibility:

| tensorflow | numpy |
| --- | --- |
| 1.15.4 | `<1.19.0,>=1.16.0`|
| 2.1 | `<2.0,>=1.16.0` |
| 2.2 | `<2.0,>=1.16.0` |
| 2.3 | `<1.19.0,>=1.16.0` |
| 2.4.0rc3 | `~=1.19.2` |

If you see a way to group the venvs more efficiently, let me know.

kba:
oh wait, ocrd_calamari does indeed require 2.3, so yes we can reuse the venv! I'll adapt the PR.

bertsky:
> The problem here is numpy compatibility:
> 
> tensorflow	numpy
> 1.15.4	`<1.19.0,>=1.16.0`
> 2.1	`<2.0,>=1.16.0`
> 2.2	`<2.0,>=1.16.0`
> 2.3	`<1.19.0,>=1.16.0`
> 2.4.0rc3	`~=1.19.2`

Wow. Had no idea it's _that_ bad. There seems to be no limit to Tensorflow madness.

PR #225, opened 2020-12-10
make install-models-calamari: Install 1.0, not 0.3 models

kba:


stweil:
It looks like this change was already committed with 9a8f0cce1d64fd357a81be8905c828af5bd938ec.

Issue #228, opened 2020-12-17
VIRTUAL_ENV parameter not working correctly


andbue:
Hi, when installing ocrd for the first time I tried to follow the instructions in README.md. The following line
`make [PYTHON=python3] [VIRTUAL_ENV=./venv] [OCRD_MODULES="..."] [TARGET...]`
made me think that the parameter VIRTUAL_ENV was needed to specify the directory for the installation. It seems (at least that's what @kba told me) that the parameter is not really helpful when calling "make all", as it is passed to every single package, leading to conflicts between package versions. Maybe the readme could be more explicit in the description of the variables.

Interestingly enough, the installation somehow finished and was mostly working. The modules ocrd-sbb-binarize and ocrd-sbb-textline-detector needed fixed imports like
```python
import tensorflow.compat.v1 as tf
tf.disable_v2_behavior()
```
to run with the newer tensorflow and the GPU was not used by default (did not investigate further). Maybe it would be a good idea to upgrade the sbb- modules to tf2 to let users profit from improvements in performance, but that would call for some testing, I guess.

cneud:
> Maybe it would be a good idea to upgrade the sbb- modules to tf2

Just a quick comment on this - yes this would be nice, but also a lot of work. There are not only API breaking changes that need to be manually checked and adapted in the sources, but also models would have to be re-trained (and accordingly, performance with TF1 versions evaluated). I am not convinced we can and will pull this off during the remaining running time of the Qurator project (where the sbb_* tools are developed).

bertsky:
Not sure this is a documentation issue or a behavioural one. The square brackets universally denote optional information in shell call mnemonics. But what you report (ie. sub-venvs conflated into a single venv) should not happen.

I have no time to investigate now, but could you please post the contents of `$VIRTUAL_ENV/bin/ocrd-sbb-binarize`?

kba:
> It seems (at least that's what @kba told me) that the parameter is not really helpful when calling "make all", as it is passed to every single package, leading to conflicts between package versions. 

I have never tried passing `VIRTUAL_ENV` to the make call, we assume that if you have a preexisting venv you want to install into, you'd just have it activated when doing `make all`. But it shouldn't make a difference. This still doesn't explain how the recursive `make` calls would reuse the top-level `VIRTUAL_ENV` since we explicitly override them (e.g. `$(OCRD_CALAMARI): VIRTUAL_ENV := $(SUB_VENV)/headless-tf2`).

[...]

PR #229, opened 2020-12-17
Fix make VIRTUAL_ENV=PATH (issue #228)

stweil:
The VIRTUAL_ENV is also passed to each recursive make, so to get the right
sub venvs it must be changed explicitly for the recursive calls.

Signed-off-by: Stefan Weil <sw@weil.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,11 +219,9 @@ OCRD_COR_ASV_ANN += $(BIN)/cor-asv-ann-repl
 $(call multirule,$(OCRD_COR_ASV_ANN)): cor-asv-ann
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_ANN)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf1
-	$(call delegate_venv,$(OCRD_COR_ASV_ANN))
-$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
+	$(call delegate_venv,$(OCRD_COR_ASV_ANN),$(SUB_VENV)/headless-tf1)
 else
 	$(pip_install)
-$(OCRD_COR_ASV_ANN): $(BIN)/ocrd
 endif
 endif
 
@@ -235,12 +233,10 @@ OCRD_COR_ASV_FST += $(BIN)/cor-asv-fst-train
 $(call multirule,$(OCRD_COR_ASV_FST)): cor-asv-fst
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_COR_ASV_FST)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf1
-	$(call delegate_venv,$(OCRD_COR_ASV_FST))
-$(OCRD_COR_ASV_FST): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
+	$(call delegate_venv,$(OCRD_COR_ASV_FST),$(SUB_VENV)/headless-tf1)
 else
-	$(MAKE) -C $< deps
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
 	$(pip_install)
-$(OCRD_COR_ASV_FST): $(BIN)/ocrd
 endif
 endif
 
@@ -251,11 +247,9 @@ OCRD_KERASLM += $(BIN)/keraslm-rate
 $(call multirule,$(OCRD_KERASLM)): ocrd_keraslm
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_KERASLM)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf1
-	$(call delegate_venv,$(OCRD_KERASLM))
-$(OCRD_KERASLM): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
+	$(call delegate_venv,$(OCRD_KERASLM),$(SUB_VENV)/headless-tf1)
 else
 	$(pip_install)
-$(OCRD_KERASLM): $(BIN)/ocrd
 endif
 endif
 
@@ -318,11 +312,9 @@ OCRD_SEGMENT += $(BIN)/ocrd-segment-repair
 $(call multirule,$(OCRD_SEGMENT)): ocrd_segment
 ifeq (0,$(MAKELEVEL))
 	$(MAKE) -B -o $< $(notdir $(OCRD_SEGMENT)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf1
-	$(call delegate_venv,$(OCRD_SEGMENT))
-$(OCRD_SEGMENT): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1
+	$(call delegate_venv,$(OCRD_SEGMENT),$(SUB_VENV)/headless-tf1)
 else
 	$(pip_install)
-$(OCRD_SEGMENT): $(BIN)/ocrd
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


kba:
Thanks for the quick fix. But why is it necessary to pass `VIRTUAL_ENV` explicitly? Are the `$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1` obsolete? And does this only happen for `make all VIRTUAL_ENV=/path/to/venv` or also for `source /path/to/venv/bin/activate; make all`?

Maybe we should also add a smoke test to the "build-pull-request" CircleCI workflow to make sure that the processors are installed into their resp. venvs?

stweil:
Running `$(MAKE)` also passes `$(MAKEFLAGS)` which include `VIRTUAL_ENV=PATH` for this special use case. Not sure whether the mentioned constructs are still necessary – maybe they can be removed now.

stweil:
> Maybe we should also add a smoke test to the "build-pull-request" CircleCI workflow to make sure that the processors are installed into their resp. venvs?

I don't care where the processors are installed as long as they work. So the smoke test should run some simple tests using the processors.

By the way, what about dropping `headless-tf2`? I think the latest TensorFlow can simply be used without any conflicts in the main `VIRTUAL_ENV`. That would save build time and disk storage. As soon as all processors work with the current TensorFlow, there would not remain sub venvs for TensorFlow any longer.

kba:
> Running $(MAKE) also passes $(MAKEFLAGS) which include VIRTUAL_ENV=PATH for this special use case.

Since this only happens when passing `VIRTUAL_ENV` to the initial make call (as apparently nobody has done before @andbue) would it not be cleaner if we caught this early on, if $(MAKELEVEL) is 0 and `$(MAKEFLAGS)` contains VIRTUAL_ENV then abort with an error message?

> I don't care where the processors are installed as long as they work.

I share the sentiment, but for this particular problem, it is important where the executables are installed to.

> By the way, what about dropping headless-tf2? I think the latest TensorFlow can simply be used without any conflicts in the main VIRTUAL_ENV. That would save build time and disk storage

Would that really save a lot of disk storage? The expensive library is tensorflow, whether it's in a sub-venv or in the main venv wouldn't make much difference.

> As soon as all processors work with the current TensorFlow, there would not remain sub venvs for TensorFlow any longer.

Not sure whether that will ever be the case. E.g. the the QURATOR-developed tools will use 1.15.x for the foreseeable future.

stweil:
> Would that really save a lot of disk storage?

The difference is about 0.5 GB (no duplicate `numpy`, `scipy`, `cv2` and `skimage`):

    9,4G	../venv-old
    8,9G	../venv-new

bertsky:
> The VIRTUAL_ENV is also passed to each recursive make, so to get the right
> sub venvs it must be changed explicitly for the recursive calls.

Thanks for the initiative. I'm afraid this is not the right solution though. I'll go through your comments and explain bit by bit.

> Are the `$(OCRD_COR_ASV_ANN): VIRTUAL_ENV := $(SUB_VENV)/headless-tf1` constructs obsolete? 

In this PR, no, they are not. The target-specific value is still needed for the `delegate_venv` rules.

> And does this only happen for `make all VIRTUAL_ENV=/path/to/venv` or also for `source /path/to/venv/bin/activate; make all`?

The former. 

`VAR=something make target` is always weaker than `make target VAR=something` – the first is merely an environment variable imported from the shell (and not exported by default to sub-make, and overriden by all makefile's variable assignments except for `?=`), whereas the second is already a `make` variable assignment (and one with priority over all makefile assignments except for those with `override` directive).

That's also the reason why we did not see this earlier – nobody tried the second form since we introduced the sub-venv logic.

> Maybe we should also add a smoke test to the "build-pull-request" CircleCI workflow to make sure that the processors are installed into their resp. venvs?

Yes, as long as we manage them the way we do, I agree we should also validate them, somehow.

> Running `$(MAKE)` also passes `$(MAKEFLAGS)` which include `VIRTUAL_ENV=PATH` for this special use case. 

Sub-make passes all makefile-defined variables (including target-specific ones) plus all exported variables from the shell environment. The rules were fine, the problem here was merely the cmdline override.

NB: `MAKEFLAGS` has nothing to do with it, it's a different mechanism.

> Not sure whether the mentioned constructs are still necessary – maybe they can be removed now.

They cannot (see above).

> By the way, what about dropping `headless-tf2`? I think the latest TensorFlow can simply be used without any conflicts in the main `VIRTUAL_ENV`.

I am not very fond of that idea, too. Not only do we still have lots of TF1-based modules (from SBB, ocrd_segment, ocrd_keraslm, cor-asv-ann). But moving heavy dependencies like TF into sub-venvs makes it easier to install other task-specific stuff to the top-level venv ad-hoc flexibly. (I use this in my derived Docker images, for example.)

The size overhead is relatively small and will go away once we shift to slim Docker/network layers.

> Since this only happens when passing `VIRTUAL_ENV` to the initial make call (as apparently nobody has done before @andbue) would it not be cleaner if we caught this early on, if $(MAKELEVEL) is 0 and `$(MAKEFLAGS)` contains VIRTUAL_ENV then abort with an error message?

Not necessary. It suffices to qualify the target-specific `VIRTUAL_ENV := $(SUB_VENV)/...` by an additional `override` directive – that's exactly what it's there for!

I'll make a new PR for this. (I also found another problem: if you selectively install executables for sub-venvs before anything for the top venv, then the dependency is missing.)

bertsky:
> I also found another problem: if you selectively install executables for sub-venvs before anything for the top venv, then the dependency is missing

The reason being simply that (in the top level) the dependencies (including creation of the venv and installation of pip and wheel) are evaluated in make phase 1, but the recipes already get the target-specific value for `VIRTUAL_ENV` (which they need for `delegate_venv` but not for `ACTIVATE_VENV`). Ideas?

bertsky:
> > I also found another problem: if you selectively install executables for sub-venvs before anything for the top venv, then the dependency is missing
> 
> The reason being simply that (in the top level) the dependencies (including creation of the venv and installation of pip and wheel) are evaluated in make phase 1, but the recipes already get the target-specific value for `VIRTUAL_ENV` (which they need for `delegate_venv` but not for `ACTIVATE_VENV`). Ideas?

One solution would be to _keep_ @stweil's idea here, and get rid of the target-specific variable, but add a second parameter to `delegate_venv` with the sub-venv path to delegate to.

bertsky:
> One solution would be to _keep_ @stweil's idea here, and get rid of the target-specific variable, but add a second parameter to `delegate_venv` with the sub-venv path to delegate to.

This also makes rules slightly better readable. @stweil can I add to this branch?

stweil:
> can I add to this branch?

Sure.

bertsky:
> > Maybe we should also add a smoke test to the "build-pull-request" CircleCI workflow to make sure that the processors are installed into their resp. venvs?
> 
> Yes, as long as we manage them the way we do, I agree we should also validate them, somehow.

I wrote a `check` target that verifies all executables are runnable (and `--help` or `-h` succeeds) and the outer venv is ok with `pip check`, but it's harder to get into the sub-venvs without writing extra lines for each module (to run `pip check` there). I also found that not all our `OCRD_EXECUTABLES` do have a `--help` or `-h` option (`typegroups-classifier` does not).

Should I add this commit anyway?

stweil:
> Should I add this commit anyway?

I suggest to create a new PR for that.

bertsky:
> > > Maybe we should also add a smoke test to the "build-pull-request" CircleCI workflow to make sure that the processors are installed into their resp. venvs?
> > 
> > 
> > Yes, as long as we manage them the way we do, I agree we should also validate them, somehow.
> 
> I wrote a `check` target that verifies all executables are runnable (and `--help` or `-h` succeeds) and the outer venv is ok with `pip check`, but it's harder to get into the sub-venvs without writing extra lines for each module (to run `pip check` there). I also found that not all our `OCRD_EXECUTABLES` do have a `--help` or `-h` option (`typegroups-classifier` does not). And of course, whether we can produce a help text does not say whether all imports will work when actually running (e.g. TF might be the wrong version). So this is both too much already and still too little.
> 
> Should I add this commit anyway?

See #230

PR #231, opened 2020-12-28
:package: v2020-12-28

kba:


stweil:
Tesseract git master is currently unusable. We have to use the latest stable pre-release `5.0.0-alpha-20201224` which is luckily also tagged.

kba:
> We have to use the latest stable pre-release 5.0.0-alpha-20201224 which is luckily also tagged.

Yes! Tagged snapshots, that is a good decision.

I'll adapt tesseract in the PR.

PR #234, opened 2021-01-18
make install-models: delegate to ocrd resmgr

kba:
Once OCR-D/core#559 is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -397,7 +397,7 @@ ifneq ($(findstring ocrd_calamari, $(OCRD_MODULES)),)
 install-models: install-models-calamari
 .PHONY: install-models-calamari
 install-models-calamari: $(BIN)/ocrd
-	. $(ACTIVATE_VENV) && ocrd resmgr download ocrd-calamari-binarize '*'
+	. $(ACTIVATE_VENV) && ocrd resmgr download ocrd-calamari-recognize '*'
 OCRD_EXECUTABLES += $(OCRD_CALAMARI)
 OCRD_CALAMARI := $(BIN)/ocrd-calamari-recognize
 $(OCRD_CALAMARI): ocrd_calamari
--- a/Makefile
+++ b/Makefile
@@ -617,83 +617,23 @@ install-models: \
 	install-models-ocropus \
 	install-models-sbb-binarization \
 	install-models-calamari
+	$(warning "Deprecated: Use 'ocrd resmgr' instead of install-* ocrd_all targets")
 
-UB_MANNHEIM_BACKUP_URL = https://ub-backup.bib.uni-mannheim.de/~stweil/ocrd-train/data
 .PHONY: install-models-tesseract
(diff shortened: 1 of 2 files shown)


bertsky (changes requested):
I recommend a more radical change: Instead of unconditionally delegating `install-models` to `install-models-ocropus`, `install-models-calamari` etc, we should make these subtargets depend on whether the respective module is in the currently active `OCRD_MODULES`, and encapsulate each such rule near the other rules for the modules. 

Thus, instead of …
```make
ifneq ($(findstring ocrd_cis, $(OCRD_MODULES)),)
OCRD_EXECUTABLES += $(OCRD_CIS)
...
```

…we would have…
```make
ifneq ($(findstring ocrd_cis, $(OCRD_MODULES)),)
install-models: install-models-cis
install-models-cis: $(BIN)/ocrd
	. $(ACTIVATE_VENV) && ocrd resmgr download ocrd-cis-ocropy-recognize '*'
OCRD_EXECUTABLES += $(OCRD_CIS)
...
```

Also, we still need to adapt `Dockerfile`:
```Dockerfile
ENV XDG_DATA_HOME /usr/local/share
VOLUME $XDG_DATA_HOME/ocrd-resources
```

(perhaps also `HOME` itself)

kba:
Having thought about it some more, I agree that `install-models-*` should depend on `OCRD_MODULES` to ensure that only the models are installed for those processors the user requested (by having it in `OCRD_MODULES`)

kba:
> we should make these subtargets depend on whether the respective module is in the currently active OCRD_MODULES

fb4ee26

> Also, we still need to adapt Dockerfile:

2fdce0a

Makefile
```diff
@@ -384,6 +394,10 @@ $(OCRD_PAGETOPDF): ocrd_pagetopdf $(BIN)/ocrd
 endif
 
 ifneq ($(findstring ocrd_calamari, $(OCRD_MODULES)),)
+install-models: install-models-calamari
+.PHONY: install-models-calamari
+install-models-calamari: $(BIN)/ocrd
+	. $(ACTIVATE_VENV) && ocrd resmgr download ocrd-calamari-binarize '*'
```
bertsky, line 400: Shouldn't that be…
```suggestion
	. $(ACTIVATE_VENV) && ocrd resmgr download ocrd-calamari-recognize '*'
```
…actually?

bertsky (approved):
Splendid – thanks!

Perhaps we could offer more of these `install-models-*` rules in order to allow a quick installation without studying all processors and available models? (Currently, I would still have to `ocrd resmgr download PROCESSOR '*'` selectively for each processor I might want to use.)

kba:
> Perhaps we could offer more of these `install-models-*`

384ee6d I added targets for all projects that we have in core's `resource_list.yml`.

You could always do the installation and then `ocrd resmgr download '*'` to download all models at once later on.

PR #239, opened 2021-02-11
Dockerfile: disable volume for resources

bertsky:
@kba, https://github.com/OCR-D/ocrd_all/commit/2fdce0a872638d16e8afe9cc2d23ebc9168b0568#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R36 was actually detrimental to our idea. See comments in the change itself for exlanation.


kba (approved):
Wasn't aware that `VOLUME` breaks inheritance. Sure, let's drop it.

PR #240, opened 2021-02-12
adapt TESSDATA to resmgr default

bertsky:
Not sure if this is the right way to do it (`make` phase 1 cannot be used to check `ocrd bashlib constants XDG_DATA_HOME` etc, but `install` recipes already use `ocrd resmgr download`, so maybe we should _export_ `XDG_DATA_HOME` defined here...). 

But the current state is clearly broken: minimal models will only be installed into `venv/share/tessdata`, where models are usable for standalone Tesseract without `--tessdata-dir`, but not for our wrapper – it now expects them in `$XDG_DATA_HOME/ocrd-resources/ocrd-tesserocr-recognize` (and there is nothing besides setting the environment variable override `export TESSDATA_PREFIX=$VIRTUAL_ENV/share/tessdata` that could save that).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -642,7 +642,9 @@ CUSTOM_DEPS += g++ make automake libleptonica-dev
 # but since we are building statically, we need more (static) libs at build time
 CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
 
-TESSDATA := $(VIRTUAL_ENV)/share/tessdata
+XDG_DATA_HOME ?= $(if $(HOME),$(HOME)/.local/share,/usr/local/share)
+DEFAULT_RESLOC ?= $(XDG_DATA_HOME)/ocrd-resources
+TESSDATA ?= $(DEFAULT_RESLOC)/ocrd-tesserocr-recognize
 TESSDATA_URL := https://github.com/tesseract-ocr/tessdata_fast
 TESSERACT_TRAINEDDATA = $(ALL_TESSERACT_MODELS:%=$(TESSDATA)/%.traineddata)
 
@@ -683,7 +685,7 @@ tesseract/Makefile.in: tesseract
 TESSERACT_CONFIG ?= --disable-openmp --disable-shared CXXFLAGS="-g -O2 -fPIC"
 $(BIN)/tesseract: tesseract/Makefile.in
 	mkdir -p $(VIRTUAL_ENV)/build/tesseract
-	cd $(VIRTUAL_ENV)/build/tesseract && $(CURDIR)/tesseract/configure --prefix="$(VIRTUAL_ENV)" $(TESSERACT_CONFIG)
+	cd $(VIRTUAL_ENV)/build/tesseract && $(CURDIR)/tesseract/configure --disable-tessdata-prefix $(TESSERACT_CONFIG)
 	cd $(VIRTUAL_ENV)/build/tesseract && $(MAKE) install
 
 # Build and install Tesseract training tools.


Makefile
```diff
@@ -683,7 +685,7 @@ tesseract/Makefile.in: tesseract
 TESSERACT_CONFIG ?= --disable-openmp --disable-shared CXXFLAGS="-g -O2 -fPIC"
 $(BIN)/tesseract: tesseract/Makefile.in
 	mkdir -p $(VIRTUAL_ENV)/build/tesseract
-	cd $(VIRTUAL_ENV)/build/tesseract && $(CURDIR)/tesseract/configure --prefix="$(VIRTUAL_ENV)" $(TESSERACT_CONFIG)
+	cd $(VIRTUAL_ENV)/build/tesseract && $(CURDIR)/tesseract/configure --disable-tessdata-prefix $(TESSERACT_CONFIG)
```
kba, line 688: This will break the plain `tesseract` CLI in the sense that you'll have to provide the `TESSDATA_PREFIX` variable every call, right? Can we set it to the `$(TESSDATA)` var as well?

bertsky: > This will break the plain `tesseract` CLI in the sense that you'll have to provide the `TESSDATA_PREFIX` variable every call, right?

Yes, you'd have to use `--tessdata-dir` each time.

> Can we set it to the `$(TESSDATA)` var as well?

We should indeed.

bertsky: > > Can we set it to the `$(TESSDATA)` var as well?
> 
> We should indeed.

Sorry, no, it's not that easy. The problem is that the compiled-in data path always gets appended by `/tessdata`.

Makefile
```diff
@@ -642,7 +642,9 @@ CUSTOM_DEPS += g++ make automake libleptonica-dev
 # but since we are building statically, we need more (static) libs at build time
 CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
 
-TESSDATA := $(VIRTUAL_ENV)/share/tessdata
+XDG_DATA_HOME ?= $(if $(HOME),$(HOME)/.local/share,/usr/local/share)
```
bertsky, line 645: Note that we also depend on #239 to get `/usr/local/share` workable again.

bertsky:
We should also merge https://github.com/OCR-D/ocrd_tesserocr/pull/172 ASAP and adapt here.

Makefile
kba: Sounds reasonable. #239 just merged.

Makefile
kba: Then let's not bother with aligning tesseract and ocrd_tesserocr with regard to model path. I think it's reasonable to expect users to either provide `---tessdata-dir` for every call to the `tesseract` CLI or create an alias/wrapper executable that does it for them.

bertsky: Yes. Either that or they `export TESSDATA_PREFIX=$XDG_DATA_HOME/ocrd-resources/ocrd-tesserocr-recognize` in their shell, which would work too (because for envvars, no `/tessdata` gets appended).

But if they do that, they'll also circumvent OcrdResourceManager in ocrd_tesserocr now. Which is ok I guess, because it's the same result. Still, I get the feeling we are missing something here...

bertsky:
> We should also merge [OCR-D/ocrd_tesserocr#172](https://github.com/OCR-D/ocrd_tesserocr/pull/172) ASAP and adapt here.

I don't see anything to adapt anymore BTW. That was just an internal fix (in order to prevent module-level context for the resmgr lookup).

Makefile
kba: As long as `TESSDATA_PREFIX` is set in the way you propose, it should not make a difference 🤞

bertsky: We could add that export to our Dockerfile BTW. Not sure how to add something to the venv shell environment in native installations, though.

kba: PR https://github.com/OCR-D/ocrd_all/pull/244 Issue in ocrd-website https://github.com/OCR-D/ocrd-website/issues/200

bertsky: Oh, no! That was stupid of me to propose in the first place. We _still need_ `--prefix` for native installations – it's not the same as tessdata-prefix!

bertsky: I think I'll just re-instate `--prefix="$(VIRTUAL_ENV)"` here (so we don't attempt to break out of the venv) and then `cp -pr $(VIRTUAL_ENV)/share/tessdata/* $(TESSDATA)/` afterwards (so we at least get the config files in the right place). Step `install-models-tesseract` can stay the same.

But what if the user disabled `tesseract` as a module here, and wants the PPA/dpkg for `ocrd_tesserocr`? I don't think this could have worked lately... anyway, that's a separate issue.

@stweil what do you think?

stweil: > This will break the plain `tesseract` CLI in the sense that you'll have to provide the `TESSDATA_PREFIX` variable every call, right? Can we set it to the `$(TESSDATA)` var as well?

I just had that breakage. As far as I see removing `--disable-tessdata-prefix` would fix that without affecting users who set `TESSDATA_PREFIX` or use `--tessdata-dir`. Is that correct?

bertsky: > I just had that breakage. As far as I see removing `--disable-tessdata-prefix` would fix that without affecting users who set `TESSDATA_PREFIX` or use `--tessdata-dir`. Is that correct?

Yes. I think that's correct. This was an overambituous change driven by the urgent need to adapt to the resmgr change. (I think we could not reach you for review at the time.) A part of it (that mistakenly took out `--prefix` as well) was already undone in #244. 

The main problem was where minimal models (`eng+osd`) would be installed to during `install-tesseract` – the default under `$prefix/share/tessdata` would only serve the standalone CLI, but we could not make Tesseract take our new XDG path (at compile time), because it does not end on `/tessdata`. It's a dilemma. We decided to place them under XDG conventions (so at least our wrapper would work without extra download steps, which is important for the segmentation functionality – something users do not expect), rendering the standalone CLI unusable without `--tessdata-dir` anyway. We figured it would be more consistent that way – but I guess it is not. The users could still place models into the conventional tessdata path themselves, and then rightly expect the CLI to work as usual. What's more, we could even place a symlink there (if the FS supports that) ourselves at install time.

PR #244, opened 2021-03-05
set TESSDATA_PREFIX in Dockerfile

kba:


bertsky (changes requested):
Would be nice if for native installations we could also amend the venv with a shell export. Ideas? (Otherwise: document what the user need to do for using Tesseract CLI via ocrd_all.)

One thing I forgot about though: We should also ensure that the minimum required `{eng,osd}.traineddata` always gets installed, regardless of `install-models` or `ocrd resmgr` actions. Plus the `configs` and `tessconfigs` directories. Since we have no record of the latter as resources, we should probably just copy them from the prebuilt location. 

So far, in ocrd_all we had `install-tesseract` download the minimal models and its submodule `make install` the configs into the tessdata path, and when installing Tesseract from dpkg (i.e. in docker-mini or custom configs without `tesseract` as a module), we had it install the respective model dpkgs.

With `configure --disable-tessdata-prefix` these files are still installed by `make install` under the default prefix `/usr/local/share/tessdata`. Which reminds me that we *do need* `--prefix` for correct native installations under `VIRTUAL_ENV`...

(I'll come back to the eng/osd/configs issue in a second, let's first repair `--prefix` ...)

kba:
f5e7a68 is a good solution to the problem of having `{osd,eng}.traineddata` and the configs in the resmgr location.

Also appreciate the documentation update. Anything more to do for this PR?

bertsky:
> [f5e7a68](https://github.com/OCR-D/ocrd_all/commit/f5e7a686450bdab7c13c7448b8743032cc3972cb) is a good solution to the problem of having `{osd,eng}.traineddata` and the configs in the resmgr location.

It's actually just the configs. Tesseract's `make install` does not install model files, this comes via our wget rules (for `ALL_TESSERACT_MODELS`) on which `install-tesseract` (and in turn `all`) depends. If the user runs `make all` or `make install-tesseract` without thinking about data paths, but then decides to define `XDG_DATA_PATH` or `TESSDATA` and do some `make install-models` or `ocrd resmgr` action, we still loose. I would favour having `install-tesseract` _not_ depend on any models whatsoever, and move everything to `install-models-tesseract`. The problem is that currently the wget approach is more versatile w.r.t. all of Tesseract's stock models, most of which we don't have in our registration yet.

kba:
>  and no-one would like to memorize the full list to download

We could have a resource `configs` that downloads just the `configs` from a release tarball and unpacks them as `config` in `$XDG_DATA_HOME/ocrd-resources/ocrd-tesserocr-recognize`. Since these are basically what we call "parameter presets" I see no reason not to include them as a resource.

kba:
> The problem is that currently the wget approach is more versatile w.r.t. all of Tesseract's stock models, most of which we don't have in our registration yet.

It would probably be foolish to mirror the full tessdata_best repository but if you have a limited list of reasonably expected stock models, we could add those. and you can always use a URL for downloading arbitrary models, so I see no real benefit of `wget` over `ocrd resmgr download`.

kba:
>  I would favour having install-tesseract not depend on any models whatsoever, and move everything to install-models-tesseract.

Agreed, it is clearer like that.

bertsky:
> > The problem is that currently the wget approach is more versatile w.r.t. all of Tesseract's stock models, most of which we don't have in our registration yet.
> 
> It would probably be foolish to mirror the full tessdata_best repository but if you have a limited list of reasonably expected stock models, we could add those. and you can always use a URL for downloading arbitrary models, so I see no real benefit of `wget` over `ocrd resmgr download`.

True! (IIRC that was the whole point of implementing the downloader in core...)

A configs resource plus a set of frequently used languages/scripts sounds good. So we can remove the `cp $(VIRTUAL_ENV)/share/tessdata $(TESSDATA)` again?

kba:
> So we can remove the cp $(VIRTUAL_ENV)/share/tessdata $(TESSDATA) again?

Yes af10035, the `configs` are now included with `ocrd resmgr download ocrd-tesserocr-recognize '*'`.

bertsky (approved):

PR #256, opened 2021-05-19
update 2021-05-19

kba:

PR #257, opened 2021-05-20
update cor-asv-ann

bertsky:


bertsky:
uh-oh, running into the #256 problems, too (i.e. keras vs tf.keras in ocrd_anybaseocr)

kba:
I've merged this into #256

Issue #259, opened 2021-06-03
Unable to find a valid copy of libtoolize or glibtoolize in your PATH!


mikegerber:
On a rather pristine Debian host:

~~~
cd tesseract && ./autogen.sh
Unable to find a valid copy of libtoolize or glibtoolize in your PATH!
./autogen.sh: 59: ./autogen.sh: bail_out: not found
Running aclocal
./autogen.sh: 77: ./autogen.sh: aclocal: not found

  Something went wrong, bailing out!

make: *** [Makefile:698: tesseract/Makefile.in] Error 1
~~~

An `apt install build-essential libtool pkg-config` appears to fix the immediate problem. I wasn't sure where to put this in the README, so I decided to open this issue instead of doing a PR.

stweil:
Yes, the dependencies listed in `Makefile` did not match those which are [required for Tesseract builds](https://tesseract-ocr.github.io/tessdoc/Compiling-%E2%80%93-GitInstallation.html).

Pull request #260 tries to fix that.

stweil:
`build-essentials` are not required, but `libtool` and `pkg-config` are required.

kba:
Fixed in https://github.com/OCR-D/ocrd_all/pull/260/files

PR #260, opened 2021-06-03
Fix dependencies for Tesseract (fix issue #259)

stweil:
- Remove development packages for graphic libraries which are not required
- Add missing development packages
- Add required package for training tools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -658,9 +658,13 @@ ifneq ($(findstring tesseract, $(OCRD_MODULES)),)
 # Tesseract.
 
 # when not installing via PPA, we must cope without ocrd_tesserocr's deps-ubuntu-modules
-CUSTOM_DEPS += g++ make automake libleptonica-dev
-# but since we are building statically, we need more (static) libs at build time
-CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev libgif-dev libjpeg-dev libpng-dev libtiff-dev
+CUSTOM_DEPS += automake ca-certificates g++ libtool make pkg-config
+# Required library.
+CUSTOM_DEPS += libleptonica-dev
+# Optional libraries for enhanced functionality.
+CUSTOM_DEPS += libarchive-dev libcurl4-nss-dev
+# Optional library for training tools.
+CUSTOM_DEPS += libpango1.0-dev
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -658,9 +658,13 @@ ifneq ($(findstring tesseract, $(OCRD_MODULES)),)
 # Tesseract.
 
 # when not installing via PPA, we must cope without ocrd_tesserocr's deps-ubuntu-modules
-CUSTOM_DEPS += g++ make automake libleptonica-dev
-# but since we are building statically, we need more (static) libs at build time
```
stweil, line 662: We disable building a shared Tesseract library, but we don't build statically. All shared libraries for the supported graphic formats should be installed as dependencies of `libleptonica-dev`.

bertsky: > We disable building a shared Tesseract library, but we don't build statically. 

You are right: With `enable_shared=no enable_static=yes` I get …

    libtool: link: g++ -g -Og -O0 -DDEBUG -std=c++17 -o tesseract src/api/tesseract-tesseractmain.o  ./.libs/libtesseract.a -L/usr/local/lib /usr/local/lib/liblept.so -lpng16 -ljpeg -lgif -ltiff -lwebpmux -lwebp -lopenjp2 -lm -lz -larchive /usr/lib/x86_64-linux-gnu/libcurl-nss.so -lrt -lpthread -pthread

…which does link the binary dynamically.

But somehow I distinctly remember having to add `libgif-dev libjpeg-dev libpng-dev libtiff-dev` when we moved towards `--disable-shared` on your recommendation here.

Is it possible that in the meantime something changed in Tesseract build itself (`enable_shared_with_static_runtimes=no` perhaps)?

stweil: There was no such change on the Tesseract side. Tesseract builds for Windows require `libtiff-dev`, but not for Linux or MacOS builds.

bertsky: Ah, found it! This is a leftover of `LDFLAGS='-static'`, which has been abandoned.

https://github.com/OCR-D/ocrd_all/commit/07ab53c0f1701a0f7aa28568ee1e90c1522e8e76

bertsky (approved):

PR #264, opened 2021-06-18
update ocrd_pc_segmentation, move from tf21 to tf2

bertsky:
To prevent pc_segmentation from dragging in TF 2.5 and thus invalidating that venv for other modules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,10 +426,10 @@ OCRD_EXECUTABLES += $(OCRD_PC_SEGMENTATION)
 OCRD_PC_SEGMENTATION := $(BIN)/ocrd-pc-segmentation
 $(OCRD_PC_SEGMENTATION): ocrd_pc_segmentation
 ifeq (0,$(MAKELEVEL))
-	$(MAKE) -B -o $< $(notdir $(OCRD_PC_SEGMENTATION)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf21
-	$(call delegate_venv,$(OCRD_PC_SEGMENTATION),$(SUB_VENV)/headless-tf21)
+	$(MAKE) -B -o $< $(notdir $(OCRD_PC_SEGMENTATION)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
+	$(call delegate_venv,$(OCRD_PC_SEGMENTATION),$(SUB_VENV)/headless-tf2)
 ocrd_pc_segmentation-check:
-	$(MAKE) check OCRD_MODULES=ocrd_pc_segmentation VIRTUAL_ENV=$(SUB_VENV)/headless-tf21
+	$(MAKE) check OCRD_MODULES=ocrd_pc_segmentation VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
 else
 	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
 	$(pip_install)


bertsky:
See https://github.com/ocr-d-modul-2-segmentierung/page-segmentation/issues/4 for details.

bertsky:
Damn. Placing ocrd_pc_segmentation in sub-venv/headless-tf2 creates another conflict around h5py (ocrd_calamari wants `<3`, but ocr4all-pixel-classifier drags in 3.1).

bertsky:
> Placing ocrd_pc_segmentation in sub-venv/headless-tf2 creates another conflict around h5py (ocrd_calamari wants `<3`, but ocr4all-pixel-classifier drags in 3.1).

It's not ocrd_pc_segmentation's or ocr2all-pixel-classifier's fault: they simply allow running the most recent TF, which happens to be incompatible with current ocrd_calamari. Unfortunately, the latter does not advertise this dependency. (See fix for the latter [here](https://github.com/OCR-D/ocrd_calamari/pull/64) and analysis of the former [here](https://github.com/ocr-d-modul-2-segmentierung/page-segmentation/issues/4).)

I see no way around this other than using different venvs for ocrd_calamari (`headless-tf24`) and ocrd_pc_segmentation (`headless-tf2`) – which will cost us another 2GB in the maximum image. 

Unless anyone has a better idea?

kba:
To summarize:

  * With `h5py == 3.1` and `tf == 2.5`, [calamari still loads models](https://github.com/Calamari-OCR/calamari/pull/271#issuecomment-879855669) w/o `str/unicode` issues
  * consequently `ocrd_calamari` also does [not need `h5py < 3` or `tf < 2.5`](https://github.com/OCR-D/ocrd_calamari/commit/2fd6e536f6bf83975ba09353a83e04d75e680535#comments)
  * `ocrd_calamari` can be moved to `tf2` and the [`h5py` downgrade is redundant](https://github.com/OCR-D/ocrd_all/pull/264/commits/57fcac59c29ee6bc0450cf663fa9cee359d7b43a)

So, we need a new release of ocrd_calamari and the only remaining `tf21` processor would be `ocrd_anybaseocr`?

bertsky:
> To summarize:
> 
> * With `h5py == 3.1` and `tf == 2.5`, [calamari still loads models](https://github.com/Calamari-OCR/calamari/pull/271#issuecomment-879855669) w/o `str/unicode` issues

yes

> * consequently `ocrd_calamari` also does [not need `h5py < 3` or `tf < 2.5`](https://github.com/OCR-D/ocrd_calamari/commit/2fd6e536f6bf83975ba09353a83e04d75e680535#comments)

yes

> * `ocrd_calamari` can be moved to `tf2` 

no, ocrd_calamari is already in headless-tf2, but it can _stay_ there along with ocrd_pc_segmentation if ...

> and the [`h5py` downgrade is redundant](https://github.com/OCR-D/ocrd_all/pull/264/commits/57fcac59c29ee6bc0450cf663fa9cee359d7b43a)

yes, the combination h5py<3 and TF<2.5 is overly restrictive, and prevents sharing the venv with ocrd_pc_segmentation. (We would need a tf24 venv for no gain.)

> So, we need a new release of ocrd_calamari and the only remaining `tf21` processor would be `ocrd_anybaseocr`?

yes

bertsky:
Ok, I have removed the h5py hack for headless-tf2 and switched to https://github.com/OCR-D/ocrd_calamari/commit/76b34c50cb1dfa10d60e6acf67dc21ad9ed0468a now – let's see if it works.

bertsky:
> Ok, I have removed the h5py hack for headless-tf2 and switched to [OCR-D/ocrd_calamari@76b34c5](https://github.com/OCR-D/ocrd_calamari/commit/76b34c50cb1dfa10d60e6acf67dc21ad9ed0468a) now – let's see if it works.

No. Seems like ocrd_pc_segmentation needs another fix. Let's see...

PR #267, opened 2021-07-21
add --no-cache to venv init

### LINKED ISSUE #266: native installation on Ubuntu 18.04 fails due to broken system pip
I don't know why we did not detect this earlier, but on a standard Ubuntu 18.04, after all system dependencies have been met (`deps-ubuntu`), a simple `make all` fails:

```
  File "/local/ocr-d/ocrd_all.new/venv/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/local/ocr-d/ocrd_all.new/venv/lib/python3.6/site-packages/pip/commands/install.py", line 290, in run
    with self._build_session(options) as session:
  File "/local/ocr-d/ocrd_all.new/venv/lib/python3.6/site-packages/pip/basecommand.py", line 69, in _build_session
    if options.cache_dir else None
  File "/usr/lib/python3.6/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not int
Makefile:145: recipe for target '/local/ocr-d/ocrd_all.new/venv/bin/pip3' failed
```

This is of course pip version 9.0.1 preinstalled from the system. Since it is necessarily the first version that any venv will get set up with, and the _same error precludes even updating pip itself_, this is a pretty severe oversight both on Python's and on Ubuntu's side.

The only workaround I know is this:

    pip install --no-cache -U pip

I recommend adding this to our venv initialization recipe:

https://github.com/OCR-D/ocrd_all/blob/93c517b9c9ad79c78e6dfa37a7d0048f6393fa11/Makefile#L145

bertsky:
Fix #266

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,11 +141,9 @@ deinit:
 
 # Get Python modules.
 
-$(VIRTUAL_ENV)/bin/$(PIP): $(ACTIVATE_VENV)
-	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --no-cache --upgrade pip setuptools
-
 $(ACTIVATE_VENV) $(VIRTUAL_ENV):
 	$(SEM) $(PYTHON) -m venv $(VIRTUAL_ENV)
+	. $(ACTIVATE_VENV) && $(SEM) $(PIP) install --no-cache --upgrade pip setuptools
 
 .PHONY: wheel
 wheel: $(BIN)/wheel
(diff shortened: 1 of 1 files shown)


stweil (commented):
Please wait with this PR until it's clear that it is really needed.

kba (approved):
> Please wait with this PR until it's clear that it is really needed.

While it is not ideal if we don't know the exact cause of this, given the lengthy debugging in https://github.com/OCR-D/ocrd_all/issues/266#issuecomment-884908530, I'm in favor of this change. It doesn't hurt performance and fixes the problem.

stweil:
I don't like unexpected code / parameters without a comment which explains the reason. And my feeling is that we already have to much of such code.

bertsky:
> I don't like unexpected code / parameters without a comment which explains the reason. And my feeling is that we already have to much of such code.

See 4195e94. (This is a matter of taste, however. If you comment on every line of code, you will soon not be able to follow/read it. To get an idea of _why_ any piece of code came to be, it's usually best to use annotate/blame or search for linked issues.)

stweil:
> It doesn't hurt performance and fixes the problem.

It does not use the cache, so adds again a (small) overhead. I still vote against this patch.

Issue #263, opened 2021-06-16
multiple Tensorflow / CUDA versions again


bertsky:
To my knowledge, despite our efforts to work around the **Tensorflow dependency hell** (each TF version being tied closely to a narrow range of CUDA / Python / Numpy versions, and in turn CUDA being dependent on certain libcudnn / nvidia-driver), we have not yet tackled the problem of providing GPU access to multiple OCR-D processors relying on **different TF versions at the same time** yet.

However, for **native installations**, the solution is not far away: Since Nvidia put the version numbers into all the package names, it is in principle possible to install multiple versions of CUDA runtime and cuDNN at the same time – as long as they all can agree on a suitable `nvidia-driver` (which is usually the newest; luckily, this one appears to be largely backwards compatible). The problem is that TF loads the libcudart dynamically and to that end, needs the right version in the dynamic linker/loader's search path. But the CUDA packages seem to only activate the last installed CUDA toolkit in ld.so.conf. This is easily fixed, however:

    # get them all
    apt install cuda-10-0 cuda-10-1 cuda-10-2 cuda-11-0 cuda-11-1 libcudnn7 libcudnn8

```
# /etc/ld.so.conf.d/cuda.conf:
/usr/local/cuda-10.0/lib64
/usr/local/cuda-10.0/targets/x86_64-linux/lib
/usr/local/cuda-10.1/lib64
/usr/local/cuda-10.1/targets/x86_64-linux/lib
/usr/local/cuda-10.2/lib64
/usr/local/cuda-10.2/targets/x86_64-linux/lib
/usr/local/cuda-11.0/lib64
/usr/local/cuda-11.0/targets/x86_64-linux/lib
/usr/local/cuda-11.1/lib64
/usr/local/cuda-11.1/targets/x86_64-linux/lib
```

<details><summary><p>This _does_ work:

    for venv in venv/local/sub-venv/headless-tf*; do . $venv/bin/activate && python -c "import tensorflow as tf; print(tf.test.is_gpu_available())"; done

</p></summary><p>

```
2021-06-16 22:07:03.732233: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2021-06-16 22:07:03.755631: I tensorflow/core/platform/profile_utils/cpu_utils.cc:94] CPU Frequency: 3399905000 Hz
[...]

PR #270, opened 2021-07-23
Docker: multi-version CUDA

bertsky:
Implements #263

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,7 @@ RUN echo "apt-get -y autoremove && apt-get clean" >> docker.sh
 RUN echo "if [[ '${PIP_OPTIONS}' =~ -e|--editable ]]; then make -i clean-olena clean-tesseract; else rm -fr /build; fi" >> docker.sh
 # run the script in one layer/step (to minimise image size)
 # (and export all variables)
-RUN set -a; bash docker.sh
+RUN set -a; bash -x docker.sh
 # update ld.so cache for new libs in /usr/local
 RUN ldconfig
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,7 @@ RUN echo "apt-get -y autoremove && apt-get clean" >> docker.sh
 RUN echo "if [[ '${PIP_OPTIONS}' =~ -e|--editable ]]; then make -i clean-olena clean-tesseract; else rm -fr /build; fi" >> docker.sh
 # run the script in one layer/step (to minimise image size)
 # (and export all variables)
-RUN set -a; bash -x docker.sh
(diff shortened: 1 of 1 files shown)


bertsky:
Too bad: This currently yields `CUDA_ERROR_SYSTEM_DRIVER_MISMATCH` in Tensorflow. Should have checked earlier (in core-cuda)...

bertsky:
> Too bad: This currently yields `CUDA_ERROR_SYSTEM_DRIVER_MISMATCH` in Tensorflow. Should have checked earlier (in core-cuda)...

It seems that the choice `nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu18.04` as base image now requires at least nvidia-driver-470 on the host system. I have 440 and 465 on systems available to me, neither of them can work the image. But that means we are making a sacrifice here: to be able to support the newest Tensorflow/CUDA as well, we are forcing _all_ host systems to get a newer driver. (It just might be that upgrading the driver is easier than upgrading CUDA. But it's still quite _inconvenient_.)

bertsky:
If you have the [Nvidia repo source](http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64), you can just update `cuda-drivers-470` which will take care of all dependencies. (But a fresh installation might work, too.)

Anyway, this does work (based on a locally built `ocrd/core-cuda` from https://github.com/OCR-D/core/pull/704).

    for venv in /usr/local/sub-venv/headless-tf*; do . $venv/bin/activate && python -c "import tensorflow as tf; print(tf.test.is_gpu_available())"; done

– yields `True` 3x

bertsky:
> #### Conflicting files
> `core`

How are you supposed to keep PRs alive which involve subrepos then? I guess I'll have to update https://github.com/OCR-D/core/pull/704 each time core master changes, and then in turn update here.

bertsky:
So to sum up, we have two drawbacks here:
- the base image size for the `-cuda` variants becomes even larger (for `ocrd:core-cuda` it's already 12 GB)
- the host system needs a recent kernel driver to run the images (even for older CUDA)

But 
- considering what we gain here,
- and how urgent this is (with https://github.com/bertsky/ocrd_detectron2/issues/7 now even blocking our `ocrd/all:maximum-cuda` build),
- and that this can probably go away as soon as we build thin images,
- and that non-Docker and non-CUDA-Docker is not even affected,
- and that it also provides a solution for native installations (i.e. running `make cuda-ubuntu` or merely `make cuda-ldconfig` as fixup),
- and that dragging along these PRs with other changes (esp. if you want to combine them with other branches) is a lot of effort,

I'd say let's merge!

PR #275, opened 2021-09-14
Fix usage of default branches

stweil:
Makefile must use well defined releases instead of the latest code from a default branch.
If an update to that latest code is wanted for the submodules, using `HEAD` is better than using `master` because it handles default branches with any name.


stweil:
This PR is required because `tesseract-ocr` now uses `main` for the default branches. That breaks downloading model files.

Makefile
```diff
@@ -685,17 +686,11 @@ all: install-tesseract
 script/%.traineddata: $(TESSDATA)/script/%.traineddata
 	$(MAKE) $^
 
-# Special rule for equ.traineddata which is only available from tesseract-ocr/tessdata.
-$(TESSDATA)/equ.traineddata:
-	@mkdir -p $(dir $@)
-	$(call WGET,$@,https://github.com/tesseract-ocr/tessdata/raw/master/$(notdir $@)) || \
-		{ $(RM) $@; false; }
```
bertsky, line 692: What happened to that rule? You just deleted the only source where this model could be retrieved.

Makefile
```diff
@@ -669,7 +669,8 @@ CUSTOM_DEPS += libpango1.0-dev
 XDG_DATA_HOME ?= $(if $(HOME),$(HOME)/.local/share,/usr/local/share)
 DEFAULT_RESLOC ?= $(XDG_DATA_HOME)/ocrd-resources
 TESSDATA ?= $(DEFAULT_RESLOC)/ocrd-tesserocr-recognize
-TESSDATA_URL := https://github.com/tesseract-ocr/tessdata_fast
+TESSDATA_RELEASE = 4.1.0
+TESSDATA_URL := https://github.com/tesseract-ocr/tessdata_fast/raw/$(TESSDATA_RELEASE)
```
bertsky, line 673: I did not know _model files_ were tagged by releases now. Why can't we just get the newest version automatically?

release.sh
```diff
@@ -75,7 +75,7 @@ list_changed_submodules () {
 
 update_one_submodule () {
     local sm="$1"
-    local branch="master"
+    local branch="HEAD"
```
bertsky, line 78: I doubt that's the correct way to pull the default branch. IIUC `origin/HEAD` is only a symref and not guaranteed to be updated locally (after the first clone), and not guaranteed to resolve to the right branch (if a commit is shared by multiple branches) by git remote transfer protocols.

Makefile
stweil: It was a special rule which is now covered by the default rule, see my commit message.

Makefile
stweil: Sure, tagging of all files which are required for a Linux distribution or for other projects like OCR-D is essential. Therefore all model files have tagged releases since some time.

You can still get the newest git release (even with my patch by providing `TESSDATA_RELEASE=HEAD`. But that latest release might be buggy or incompatible, so you do that on your own risk. For example somebody who is still running Tesseract 3.05 will run into trouble with the latest release.

Makefile
bertsky: Ah, did not notice this has been copied to tessdata_fast now.

Makefile
bertsky: Understood. Could you document the new variable then?

stweil: I did not, because I think users should not use it. They should use a tagged release of `ocrd_all` which uses a tagged release of `tessdata_fast`. We also don't document some other variables, for example `TESSDATA_URL` which exists since 2 years.

release.sh
stweil: Maybe you are right. But "master" is also wrong for repositories which use "dev", "main" or some other name. What would you propose?

stweil: I'll make an additional pull request without that commit. As I always say: don't mix too many commits in the same pull request.

stweil:
The first commit of this PR is now in the separate PR #276.

release.sh
bertsky: > Maybe you are right. But "master" is also wrong for repositories which use "dev", "main" or some other name. What would you propose?

Yes, we've had ocrd_cis:dev for a while – it was an extra hickup for @kba each time as I recall. If indeed more and more repos play this game, then we need something robust that does work under all circumstances (or at least fails loudly).

Perhaps someone else with more git fu?

bertsky: > I'll make an additional pull request without that commit. As I always say: don't mix too many commits in the same pull request.

Please don't – these belong together.

stweil: The reason for splitting this in two PRs is quite simple: the first commit fixes things for everybody, the second fix (which obviously needs more thoughts) is required only for a few people like @kba.

Therefore I'd prefer to have the first commit (in PR #276) merged soon.

PR #277, opened 2021-09-22
update tesserocr (fixing ocrd_tesserocr)

bertsky:
This fixes a bug in ocrd-tesserocr-segment (and -recognize with segmentation enabled) where the binarized AlternativeImages are all-black because of a recent change in tesserocr's image serialization (from JPEG to BMP), see https://github.com/sirfz/tesserocr/issues/274.


bertsky:
CI failure is #276

PR #282, opened 2021-11-25
Update tesseract

stweil:


stweil:
Tesseract 5.0.0 will be released soon, and this is the latest code for that release.

PR #285, opened 2022-02-02
Update 2022 02 02

kba:
New core v2.30.0, faster dinglehopper, new OCR alignment processor `ocrd-cor-asv-ann-align`, new features for workflow-configuration and lots of fixes in general.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,6 +237,21 @@ else
 endif
 endif
 
+ifneq ($(findstring ocrd_detectron2, $(OCRD_MODULES)),)
+OCRD_EXECUTABLES += $(OCRD_DETECTRON2)
+OCRD_DETECTRON2 += $(BIN)/ocrd-detectron2-segment
+$(call multirule,$(OCRD_DETECTRON2)): ocrd_detectron2
+ifeq (0,$(MAKELEVEL))
+	$(MAKE) -B -o $< $(notdir $(OCRD_DETECTRON2)) VIRTUAL_ENV=$(SUB_VENV)/headless-torch14
+	$(call delegate_venv,$(OCRD_DETECTRON2),$(SUB_VENV)/headless-torch14)
+ocrd_detectron2-check:
+	$(MAKE) check OCRD_MODULES=ocrd_detectron2 VIRTUAL_ENV=$(SUB_VENV)/headless-torch14
+else
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
+	$(pip_install)
+endif
+endif
+
 ifneq ($(findstring cor-asv-fst, $(OCRD_MODULES)),)
 deps-ubuntu-modules: cor-asv-fst
 OCRD_EXECUTABLES += $(OCRD_COR_ASV_FST)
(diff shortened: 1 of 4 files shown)


bertsky:
I suggest adding ocrd_detectron2 and ocrd_doxa, too.

kba:
> I suggest adding ocrd_detectron2 and ocrd_doxa, too.

Sure, will do after the tech call.

bertsky:
> context deadline exceeded

That's the second time in a row. Maybe should look into the total time constraints per job...

bertsky:
> > context deadline exceeded
> 
> That's the second time in a row. Maybe should look into the total time constraints per job...

[Sounds](https://support.circleci.com/hc/en-us/articles/4410707277083-Context-deadline-exceeded-after-1-hour-Build-timed-out-Free-tier-only-) like we'd need to pay for longer than 1h per job.

Then let's simply switch to `make -j` and :crossed_fingers:

kba:
> [Sounds](https://support.circleci.com/hc/en-us/articles/4410707277083-Context-deadline-exceeded-after-1-hour-Build-timed-out-Free-tier-only-) like we'd need to pay for longer than 1h per job.

> This is part of an update to our new pricing plans. In particular, builds on the Free plan are limited to a maximum duration of 1 hour.

Looks like we need to switch CI providers :(

Makefile
```diff
@@ -479,6 +496,13 @@ else
 endif
 endif
 
+ifneq ($(findstring ocrd_doxa, $(OCRD_MODULES)),)
+OCRD_EXECUTABLES += $(OCRD_DOXA)
+OCRD_DOXA := $(BIN)/ocrd-doxa-binarize
+$(OCRD_DOXA): ocrd_doxa $(BIN)/ocrd
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< install
```
bertsky, line 503: nothin fancy here:
```suggestion
	$(pip_install)
```

stweil:
What about updating to tesseract 5.0.1, too.

Note that there is currently a broken build (assertion) for ocrd_pc_segmentation because latest pip breaks pip-tools (see https://github.com/jazzband/pip-tools/issues/1558), so a simply `make all` stops early.

kba:
> Note that there is currently a broken build (assertion) for ocrd_pc_segmentation because latest pip breaks pip-tools (see [jazzband/pip-tools#1558](https://github.com/jazzband/pip-tools/issues/1558)), so a simply `make all` stops early.

Thanks for the info, we also noticed this yesterday. A PR fixing this seems imminent https://github.com/jazzband/pip-tools/pull/1567

> What about updating to tesseract 5.0.1, too?

Sure, I'll add that.

stweil:
The latest Tesseract main now includes a fix for issue https://github.com/OCR-D/ocrd_tesserocr/issues/182. @bertsky, do you think that should be included as well, or would you prefer a fix in ocrd_tesserocr?

bertsky:
> The latest Tesseract main now includes a fix for issue [OCR-D/ocrd_tesserocr#182](https://github.com/OCR-D/ocrd_tesserocr/issues/182). @bertsky, do you think that should be included as well, or would you prefer a fix in ocrd_tesserocr?

@stweil I have not got around to test this. But since `shrink_polygons` is optional, I would not hasten this (I recall the problem I found with result iterators was quite involved)

bertsky:
Oh, since https://github.com/tesseract-ocr/tesseract/commit/443933a75a377b608acc0e9912e792e7a97919c9 is already merged, and catching a nullptr is always a good idea – I'd say yes, lease update Tesseract once more, regardless.

Makefile
```diff
@@ -237,6 +239,21 @@ else
 endif
 endif
 
+ifneq ($(findstring ocrd_detectron2, $(OCRD_MODULES)),)
+OCRD_EXECUTABLES += $(OCRD_DETECTRON2)
+OCRD_DETECTRON2 += $(BIN)/ocrd-detectron2-segment
+$(call multirule,$(OCRD_DETECTRON2)): ocrd_detectron2
```
stweil, line 245: Is there a special reason why multirule was used here?

Issue #286, opened 2022-02-05
CI/Docker build: faster or parallel


bertsky:
In `maximum-cuda`, we recently broke the 1h build time boundary for CircleCI free accounts. Hence the attempt to use `make -j` again to reduce it (the VMs are multi-core, so this should work in principle).

But it seems that we have a race condition which has not been detected yet:

```
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/sub-venv/headless-tf2/bin/pip' -> '/tmp/pip-0dn813kj-uninstall/usr/local/sub-venv/headless-tf2/bin/pip'
```

This looks like multiple `pip` calls on the same venv clash. Maybe we need to synchronize these calls in the same way we do for `git`  calls.

bertsky:
I'm not sure #287 actually does what we want. This looks like we still have races:
```
sem --will-cite --fg --id ocrd_all_pip/usr/local/sub-venv/headless-tf1 python3 -m venv /usr/local/sub-venv/headless-tf1
sem --will-cite --fg --id ocrd_all_pip/usr/local/sub-venv/headless-tf1 python3 -m venv /usr/local/sub-venv/headless-tf1
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/usr/local/sub-venv/headless-tf1/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']

Makefile:152: recipe for target '/usr/local/sub-venv/headless-tf1/bin/activate' failed
make[1]: *** [/usr/local/sub-venv/headless-tf1/bin/activate] Error 1
make[1]: Leaving directory '/build'
```
Or in another variant:
```
Successfully built ocrd-cor-asv-ann gast termcolor
Installing collected packages: numpy, protobuf, markdown, h5py, grpcio, absl-py, termcolor, tensorflow-estimator, tensorboard, scipy, python-dateutil, pyparsing, opt-einsum, kiwisolver, keras-preprocessing, keras-applications, google-pasta, gast, cycler, astor, tensorflow-gpu, matplotlib, keras, editdistance, ocrd-cor-asv-ann
  Attempting uninstall: numpy
    Found existing installation: numpy 1.19.5
    Uninstalling numpy-1.19.5:
[...]

PR #287, opened 2022-02-06
isolate pip with venv semaphores

bertsky:
Pursuant to #286, try using the same mechanism we already use for `git` to bring about isolation for `pip` as well – but allow different venvs to be modified concurrently.

Also contains an attempt at speeding up build by replacing the costly re-pip trick (forcing reinstallation to ensure up-to-date timestamps) with just direct timestamp update (but for the targets only).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,12 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
-define SEM
+define SEMGIT
 $(if $(shell sem --version 2>/dev/null),sem --will-cite --fg --id ocrd_all_git,$(error cannot find package GNU parallel))
+endef
+
+define SEMPIP
+$(if $(shell sem --version 2>/dev/null),sem --will-cite --fg --id ocrd_all_pip$(VIRTUAL_ENV),$(error cannot find package GNU parallel))
 endef
 
 define WGET
@@ -126,9 +130,9 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 ifneq ($(NO_UPDATE),1)
 $(OCRD_MODULES): always-update
-	$(SEM) git submodule sync $(GIT_RECURSIVE) $@
+	$(SEMGIT) git submodule sync $(GIT_RECURSIVE) $@
 	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
-		$(SEM) git submodule update --init $(GIT_RECURSIVE) $(GIT_DEPTH) $@ && \
+		$(SEMGIT) git submodule update --init $(GIT_RECURSIVE) $(GIT_DEPTH) $@ && \
 		touch $@; fi
 endif
 endif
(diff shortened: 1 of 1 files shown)


kba (approved):

bertsky:
Remaining CI failure seems to be only the stdout timeout for `docker save`. I have increased it on master.

Makefile
```diff
@@ -31,10 +31,14 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
-define SEM
+define SEMGIT
 $(if $(shell sem --version 2>/dev/null),sem --will-cite --fg --id ocrd_all_git,$(error cannot find package GNU parallel))
 endef
```
stweil, line 36: Is `shell`necessary here (and in the other places)?

bertsky:
> Remaining CI failure seems to be only the stdout timeout for `docker save`. I have increased it on master.

No idea why that update does not take effect in PRs immediately. (Anyway – should work, PR is ready to be merged AFAICS.)

bertsky:
This does not update the modules themselves, so we need not run release.sh – can I merge this, @kba?

bertsky:
CI failure: bad news – new upstream inconsistencies …

```
numpy>=1.20.0 (from imageio==2.16.0->scikit-image==0.17.2->ocr4all-pixel-classifier==0.6.5->-r requirements.in
numpy~=1.19.2 (from tensorflow==2.6.2->ocr4all-pixel-classifier==0.6.5->-r requirements.in
```
This came from ocrd_pc_segmentation – I suggest we try to fix that upstream.

bertsky:
> This came from ocrd_pc_segmentation – I suggest we try to fix that upstream.

Strange. [ocr4all-pixel-classifier requirements](https://gitlab2.informatik.uni-wuerzburg.de/ocr4all-page-segmentation/ocr4all-pixel-classifier/-/blob/master/requirements.in) does not even allow TF 2.6 – it requries <=2.5. And no restriction on scikit-image. But somehow that ends up in the above combination. It has to do with the mysterious `pip-compile` mechanism.

@chreul @maxnth @chaddy314 could you please assist?

maxnth:
@bertsky I just contacted the original author of the pixel-classifier and he's looking into fixing it upstream.

crater2150:
The problem seems to be, that `scikit-image` depends on `imageio>=2.4.1`. Note the missing upper bound on the version. With python packages, the only option to have a build that will still work in the future seems to be to hardcode all dependencies *including transitives* to a fixed version :(

I've added `imageio ~= 2.4.1` to ocr4all-pixel-classifier's dependencies as a workaround, and also re-added the full version-pinned `requirements.txt`. I also published a release 0.6.6 on PyPI.

bertsky:
> The problem seems to be, that `scikit-image` depends on `imageio>=2.4.1`. Note the missing upper bound on the version. With python packages, the only option to have a build that will still work in the future seems to be to hardcode all dependencies _including transitives_ to a fixed version :(

Indeed. Which will likely cause further conflicts as more packages are needed in the same venv.

Could anything be done on the TF version front? What was the reason for TF<=2.5 in the first place? (I have recently seen that the HDF5 model format quickly becomes incompatible with other minor releases of Python, because the custom layer code is serialised via marshal format. The better alternative is TF's [SavedModel](https://www.tensorflow.org/guide/keras/save_and_serialize) format, which is basically a directory with Protobuf file and some resource files. Conversion is as simple as an interactive load+save. Would that help to make ocr4all-pixel-classifier models run on the newer versions?

> I've added `imageio ~= 2.4.1` to ocr4all-pixel-classifier's dependencies as a workaround, and also re-added the full version-pinned `requirements.txt`. I also published a release 0.6.6 on PyPI.

Thanks @crater2150 – that helped!

crater2150:
> Could anything be done on the TF version front? What was the reason for TF<=2.5 in the first place?
I think, 2.5 was the most recent TF version, when that line was added, it simply hasn't been tested with newer versions.

I haven't had any problems with the HDF5 models between TF 2.0 and 2.5, so I'd say chances are good a newer TF version will just work. There isn't anyone actively developing the pixel classifier anymore, so larger changes to the codebase probably won't happen from our side.

PR #288, opened 2022-02-17
update ocrd_kraken

bertsky:
Even if still PR, the newer ocrd_kraken is already usable and therefore should be re-activated.

(Note: will probably work even better after applying the [recent suggestions](https://github.com/OCR-D/ocrd_kraken/pull/33#pullrequestreview-886469286) and updating here again.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
       - run:
           name: persist image
           command: docker image save ocrd/all:maximum-cuda > ocrd-all-maximum.tar
-          no_output_timeout: 30min
+          no_output_timeout: 30m
       # can be downloaded from CircleCI.com and imported via "docker image load"
       - store_artifacts:
           path: ocrd-all-maximum.tar
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "clstm"]
-	path = clstm
-	url = https://github.com/tmbdev/clstm.git
 [submodule "ocrd_keraslm"]
 	path = ocrd_keraslm
 	url = https://github.com/OCR-D/ocrd_keraslm.git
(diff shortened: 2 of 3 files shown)


bertsky:
> Did you merge the fix in master back to this branch? I did for your
> PR #289 and it's running again.

I didn't know CI config is taken from the very branch it applies to.

CI failure: see [here](https://github.com/OCR-D/ocrd_kraken/pull/33#pullrequestreview-900803910)

stweil:
What about https://github.com/OCR-D/ocrd_kraken/pull/33? I think we should require kraken 3.0.9 or even 4.0.

stweil:
Maybe PR #297 which also touches the Makefile rules for `ocrd_kraken` can be integrated here, too (or applied after this PR to avoid a merge conflict here).

bertsky:
@stweil,
> What about [OCR-D/ocrd_kraken#33](https://github.com/OCR-D/ocrd_kraken/pull/33)? I think we should require kraken 3.0.9 or even 4.0.

I don't understand your question. This PR sets ocrd_kraken to PR 33, which already requries `kraken>=3.0` in ocrd_all.

> Maybe PR #297 which also touches the Makefile rules for `ocrd_kraken` can be integrated here, too (or applied after this PR to avoid a merge conflict here).

No, it's the other way round. This PR is minimal and has been hanging here for one month already.

stweil:
> No, it's the other way round. This PR is minimal and has been hanging here for one month already.

To bring it to the point: this PR breaks some build configurations which no longer work when it is applied. I only suggested to fix that before the PR is merged instead of first breaking and then fixing.

bertsky:
> To bring it to the point: this PR breaks some build configurations which no longer work when it is applied. I only suggested to fix that before the PR is merged instead of first breaking and then fixing.

You mean Python 3.10? That can wait. Users should finally have ocrd_kraken back. Merging is on your side.

kba (approved):

PR #289, opened 2022-02-19
make Python 3.8 work

bertsky:
Depends on https://github.com/OCR-D/ocrd_anybaseocr/pull/89 (which in turn depends on https://github.com/OCR-D/ocropy/pull/2).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -603,10 +603,11 @@ define pip_install
 . $(ACTIVATE_VENV) && cd $< && $(SEMPIP) $(PIP) install $(PIP_OPTIONS_E) . && touch -c $@
 endef
 
-# Workaround for missing prebuilt versions of TF<2 for Python>3.6
+# Workaround for missing prebuilt versions of TF<2 for Python==3.8
+# todo: find another solution for 3.9, 3.10 etc
 # Nvidia has them, but under a different name, so let's rewrite that:
 define pip_install_tf1nvidia =
-. $(ACTIVATE_VENV) && if ! $(PIP) show -q tensorflow-gpu; then \
+. $(ACTIVATE_VENV) && if ! $(PYTHON) -c "import sys; sys.exit(sys.version_info.major==3 and sys.version_info.minor==8)" && ! $(PIP) show -q tensorflow-gpu; then \
 	$(PIP) install nvidia-pyindex && \
 	pushd $$(mktemp -d) && \
 	$(PIP) download --no-deps nvidia-tensorflow && \
--- a/Makefile
+++ b/Makefile
@@ -620,7 +620,8 @@ define pip_install_tf1nvidia =
 	sed -i s/nvidia_tensorflow/tensorflow_gpu/g $$name/tensorflow_core/tools/pip_package/setup.py && \
 	pushd $$name && for path in $$name*; do mv $$path $${path/$$name/$$newname}; done && popd && \
 	$(PYTHON) -m wheel pack $$name && \
-	$(PIP) install $$newname*.whl && popd && rm -fr $$OLDPWD; fi
+	$(PIP) install $$newname*.whl && popd && rm -fr $$OLDPWD; fi && \
+	$(PIP) install imageio==2.4.1 # preempt conflict over numpy between scikit-image and tensorflow
 endef
 
 # pattern for recursive make:
--- a/Makefile
+++ b/Makefile
@@ -478,10 +478,10 @@ OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-textline
 OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-layout-analysis
 $(call multirule,$(OCRD_ANYBASEOCR)): ocrd_anybaseocr
 ifeq (0,$(MAKELEVEL))
-	$(MAKE) -B -o $< $(notdir $(OCRD_ANYBASEOCR)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf21
-	$(call delegate_venv,$(OCRD_ANYBASEOCR),$(SUB_VENV)/headless-tf21)
+	$(MAKE) -B -o $< $(notdir $(OCRD_ANYBASEOCR)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
+	$(call delegate_venv,$(OCRD_ANYBASEOCR),$(SUB_VENV)/headless-tf2)
 ocrd_anybaseocr-check:
-	$(MAKE) check OCRD_MODULES=ocrd_anybaseocr VIRTUAL_ENV=$(SUB_VENV)/headless-tf21
+	$(MAKE) check OCRD_MODULES=ocrd_anybaseocr VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
 else
 	cd $< ; $(MAKE) patch-pix2pixhd
 	$(pip_install)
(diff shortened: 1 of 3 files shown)


bertsky:
CI failure: the same idiotic reason as in the [other PRs](https://github.com/OCR-D/ocrd_all/pull/288#issuecomment-1043619930). Anyone has a solution?

Makefile
```diff
@@ -593,6 +602,25 @@ define pip_install
...
+	pushd $$(mktemp -d) && \
+	$(PIP) download --no-deps nvidia-tensorflow && \
+	for name in nvidia_tensorflow-*.whl; do name=$${name%.whl}; done && \
+	$(PYTHON) -m wheel unpack $$name.whl && \
+	for name in nvidia_tensorflow-*/; do name=$${name%/}; done && \
+	newname=$${name/nvidia_tensorflow/tensorflow_gpu} &&\
+	sed -i s/nvidia_tensorflow/tensorflow_gpu/g $$name/$$name.dist-info/METADATA && \
+	sed -i s/nvidia_tensorflow/tensorflow_gpu/g $$name/$$name.dist-info/RECORD && \
+	sed -i s/nvidia_tensorflow/tensorflow_gpu/g $$name/tensorflow_core/tools/pip_package/setup.py && \
+	pushd $$name && for path in $$name*; do mv $$path $${path/$$name/$$newname}; done && popd && \
```
kba, line 619: wow

kba (approved):

Makefile
stweil: Do you expect that this will work for any architecture? I am afraid that it is limited to x86_64.

stweil (changes requested):
Building with Python 3.7 no longer works for me after I merged this PR. It fails for the `headless-tf1` parts.

I also tried Python 3.9 (the default for Debian stable) which failed similarly.

bertsky:
> Building with Python 3.7 no longer works for me after I merged this PR. It fails for the `headless-tf1` parts.

Then please share the error log so we can resolve this quickly.

> I also tried Python 3.9 (the default for Debian stable) which failed similarly.

Yeah sure, but that's true for master, too. (I am not claiming I can solve **all problems at once**.)

bertsky:
CI failure:
```
. /usr/local/sub-venv/headless-torch14/bin/activate && sem --will-cite --fg --id ocrd_all_pip/usr/local/sub-venv/headless-torch14 pip3 install --upgrade pip setuptools
Exception:
  File "/usr/local/sub-venv/headless-torch14/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/sub-venv/headless-torch14/lib/python3.6/site-packages/pip/commands/install.py", line 290, in run
    with self._build_session(options) as session:
  File "/usr/local/sub-venv/headless-torch14/lib/python3.6/site-packages/pip/basecommand.py", line 69, in _build_session
    if options.cache_dir else None
  File "/usr/lib/python3.6/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not int
```
I remember seeing this before, and solving be replacing/removing the pip cache dir.

stweil:
> Then please share the error log so we can resolve this quickly.

See https://github.com/stweil/ocrd_all/actions.

Before merging this pull request, Python 3.6 and Python 3.7 work fine while Python 3.8 fails as expected (https://github.com/stweil/ocrd_all/actions/runs/1937666710).

After merging this pull request, Python 3.7 fails (https://github.com/stweil/ocrd_all/actions/runs/1937673668).

bertsky:
> After merging this pull request, Python 3.6 and Python 3.8 work, but Python 3.7 fails (https://github.com/stweil/ocrd_all/actions/runs/1937673668).

Thanks! I updated the workaround to only activate in the case of Python 3.8. (AFAICS, on 3.7 this is not necessary, and we simply do not newer versions anyway.)

bertsky:
> I remember seeing this before, and solving be replacing/removing the pip cache dir.

Which is strange, because you would expect the CI container to start without a cache, and without `-j` there should be no outside interference/race against the cache...

Current CI failure is the 1h timeout again. :disappointed:

Makefile
stweil: Ping @bertsky.

stweil (changes requested):
It looks like handling of non Intel architectures is still missing.

stweil:
@bertsky, I think we can address the unhandled target architectures later and apply the pull request.

Maybe you can change the code which checks the Python version and introduce a new macro `PYTHON_VERSION` which will be needed for other rules, too: https://github.com/stweil/ocrd_all/commit/663544450eb9d51e96aea8619dcb94149637e7e7

Makefile
bertsky: > Do you expect that this will work for any architecture? I am afraid that it is limited to x86_64.

AFAIK Nvidia only ships wheels for Linux on x86_64. (See [release notes](https://docs.nvidia.com/deeplearning/frameworks/tensorflow-wheel-release-notes/tf-wheel-rel.html). Since they seem to be the only project willing to continue to support TF 1, that's as good as it gets.)

bertsky:
@stweil,
> I think we can address the unhandled target architectures later and apply the pull request.

Definitely. When and if...

> Maybe you can change the code which checks the Python version and introduce a new macro `PYTHON_VERSION` which will be needed for other rules, too: [stweil@6635444](https://github.com/stweil/ocrd_all/commit/663544450eb9d51e96aea8619dcb94149637e7e7)

There's a [recipe here](https://github.com/bertsky/ocrd_all/commit/c376a6bd79a092efe87a1286965f52f3a0b2ea3a) already which detects the major version. IIUC you want to expose that as a make variable for other rules as well. How is that related to this PR?

bertsky:
CI: 033b5dd does the trick for now. We should revisit the problems related to use of `make -j` later, but for now the time is just enough for the 1h+1m limit. (We can still disable the lengthy `docker image save` and artifact uploading later, if we exceed the limit but have not tackled the parallel build issues yet.)

Thus, ready for merging.

bertsky:
Now runs into the CircleCI timeout again. I'll disable saving artifacts.

Makefile
```diff
@@ -786,13 +820,20 @@ clean-tesseract:
 # install git and parallel first (which is required for the module updates)
 deps-ubuntu:
 	apt-get -y install git parallel
+ifneq ($(suffix $(PYTHON)),)
+# install specific Python version in system via PPA
+	apt-get install -y software-properties-common
+	add-apt-repository -y ppa:deadsnakes/ppa
+	apt-get update
+	apt-get install -y --no-install-recommends $(notdir $(PYTHON))-dev $(notdir $(PYTHON))-venv
+endif
```
stweil, line 829: Why was this block added? I don't think that it is a good idea to install a Python package from PPA just because the user added `PYTHON=python3.7` to the build options. It can either overwrite an existing installation or install an unneeded 2nd Python installation.

I noticed these instructions while building a Docker image with GitLab actions which already provides any desired Python version.

bertsky: Because our current reference platform, i.e. the one targeted by `deps-ubuntu`, is still Ubuntu 18.04. And that uses 3.6 by default for python3, and needs PPAs for more recent versions. (Same with our ocrd_tesserocr recipe.)

I agree it's not a good pattern, but we need some kind of dual version support during the grace period between Ubuntu 18 and 20 (Python 3.6 and 3.8). Think of it as something we can drop as soon as the shift to 20 is complete. (And after that we should find better mechanisms to encapsulate system dependencies.)

I think the natural way to avoid interference with tailored environments like CI is to _not_ set `PYTHON` there, but instead make it that environment's default version.

stweil: 18.04 does not need PPA for Python 3.7: https://packages.ubuntu.com/search?keywords=python3.7. And I don't think that the installation of Python should be part of the Makefile rules. Let it fail if someone says `make all PYTHON=python3.7` without providing that version.

PR #292, opened 2022-03-07
Add GitHub action for `make all`

stweil:
It is currently configured to be triggered manually either from the
GitHub web frontend or from the GitHub API.

`make all` can be run with a selected version of Ubuntu and Python,
or a matrix of versions.

### DIFF
--- a/.github/workflows/makeall-linux.yaml
+++ b/.github/workflows/makeall-linux.yaml
@@ -1,0 +1,51 @@
+# Reusable GitHub workflow for `make all` on Linux'.
+
+name: make all (Linux)
+
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+on:
+  workflow_call:
+    # Workflow triggered by other workflow.
+    inputs:
+      os:
+        description: 'Stringified JSON object with operating systems'
+        default: 'undefined'
+        required: false
+        type: string
+      python-version:
+        description: 'Python version'
+        default: 'undefined'
+        required: false
+        type: string
+
+jobs:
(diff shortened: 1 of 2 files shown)


stweil:
As soon as this PR was merged, we can run our first GitHub action for the full matrix of
[ubuntu-18.04, ubuntu-20.04] x [Python 3.6, 3.7, 3.8, 3.9, 3.10] (10 runs).

For both versions of Ubuntu, Python 3.6 and 3.7 (and 3.8 with PR #289) should build fine, while the other Python versions are expected to abort when trying to get Tensorflow 1.5.

stweil:
Example of GitHub action with full test matrix: https://github.com/stweil/ocrd_all/actions/runs/1947808212.

Even with PR #289, only Python 3.6 builds all targets. Python 3.7 and 3.8 fail when building `cor-asv-ann-check`.

bertsky:
> Even with PR #289, only Python 3.6 builds all targets. Python 3.7 and 3.8 fail when building `cor-asv-ann-check`.

That seems to be coming from new upstream conflicts between Tensorflow and Scikit-Image. We recently [encountered that](https://github.com/OCR-D/ocrd_all/pull/287#issuecomment-1048832299) in ocrd_pc_segmentation, but it seems to be wider. Needs to be investigated.

stweil:
I rebased and merged my latest pull requests and started now the first GitHub action for OCR-D/ocrd_all with the full build matrix: https://github.com/OCR-D/ocrd_all/actions/runs/1959782165.

PR #293, opened 2022-03-08
Update apt package list before trying to install packages

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -785,6 +785,7 @@ clean-tesseract:
 # (in case deps-ubuntu has been used with sudo and some modules were updated)
 # install git and parallel first (which is required for the module updates)
 deps-ubuntu:
+	apt-get update
 	apt-get -y install git parallel
 	$(MAKE) deps-ubuntu-modules
 	chown -R --reference=$(CURDIR) .git $(OCRD_MODULES)


stweil:
An outdated list of packages can result in a failure of `sudo make deps-ubuntu` like in this example: https://github.com/stweil/ocrd_all/actions/runs/1952943997.

Makefile
```diff
@@ -785,6 +785,7 @@ clean-tesseract:
 # (in case deps-ubuntu has been used with sudo and some modules were updated)
 # install git and parallel first (which is required for the module updates)
 deps-ubuntu:
+	apt-get update
```
bertsky, line 788: That does not belong in this rule IMO. The apt mechanism should be initialized / updated before and independent of the `deps-ubuntu` chain.

bertsky (commented):
@stweil, you let pass no more than a day without waiting for a single review. That's uncooperative. I always go out of my way to give everyone a chance to take a good look at changes, even if they are urgent. (And you usually take several weeks, if you review at all.)

Makefile
stweil: Building `ocrd_all` was currently broken, so users could expect a fast fix. It's your right to prefer and suggest a different solution, but the solution implemented by adding this single line does not harm and works fine.

stweil: @bertsky, how exactly would your preferred solution look like?

PR #295, opened 2022-03-10
[no ci] [RFC] Rework build process

stweil:
I currently try to rework the build process to support more modern Linux distributions and Python versions. Ideally the full test matrix of Ubuntu LTS versions and Python versions ranging from 3.6 to 3.10 should build fine.

My changes address several issues:

- Check early whether the required Tensorflow versions are available and skip those submodules which cannot be built because of missing Tensorflow.
- Use different rules for Python 3.6 and 3.10 when building `ocrd_kraken`.
- Disable `ocrd_cis` unconditionally because it requires an old `calamari_ocr` causing version conflicts.
- Address potential version conflicts for Python modules which are required by different OCR-D processors by using constraints.
- Reduce the number of sub_venvs by installing all modules which work with a recent Tensorflow in the main venv.
- Install Python module `wheel` early in all virtual environments, so it is no longer needed as a dependency.
- ... and other changes.

This PR is not for merging, but for discussion of the different aspects with the goal to find a consensus on the right solution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,9 +253,10 @@ endif
 
 ifneq ($(findstring ocrd_detectron2, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_DETECTRON2)
-OCRD_DETECTRON2 += $(BIN)/ocrd-detectron2-segment
-$(call multirule,$(OCRD_DETECTRON2)): ocrd_detectron2
-	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps
+OCRD_DETECTRON2 := $(BIN)/ocrd-detectron2-segment
+$(OCRD_DETECTRON2): ocrd_detectron2
+	# . $(ACTIVATE_VENV) && $(MAKE) -C $< deps
+	. $(ACTIVATE_VENV) && cd $< && pip install 'git+https://github.com/facebookresearch/detectron2.git' && pip install -r requirements.txt
 	$(pip_install)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -47,27 +47,31 @@ ifdef CHECK_SUBENVS
 
 # Create all required virtual environments.
 ifeq ($(wildcard $(VIRTUAL_ENV)),)
+ifeq ($(PYTHON_VERSION),3.6)
+create_venv := $(shell $(PYTHON) -m venv $(VIRTUAL_ENV) && bash -c "source $(VIRTUAL_ENV)/bin/activate && curl https://bootstrap.pypa.io/pip/3.6/get-pip.py | python - && pip install -U pip setuptools wheel"))
+else
 create_venv := $(shell $(PYTHON) -m venv $(VIRTUAL_ENV) && bash -c "source $(VIRTUAL_ENV)/bin/activate && pip install -U pip setuptools wheel"))
+endif
 endif
 ifeq ($(wildcard $(SUB_VENV)/headless-tf1),)
 create_venv := $(shell $(PYTHON) -m venv $(SUB_VENV)/headless-tf1 && bash -c "source $(SUB_VENV)/headless-tf1/bin/activate && pip install -U pip setuptools wheel")
 endif
-ifeq ($(wildcard $(SUB_VENV)/headless-tf21),)
-create_venv := $(shell $(PYTHON) -m venv $(SUB_VENV)/headless-tf21 && bash -c "source $(SUB_VENV)/headless-tf21/bin/activate && pip install -U pip setuptools wheel")
-endif
+#ifeq ($(wildcard $(SUB_VENV)/headless-tf21),)
+#create_venv := $(shell $(PYTHON) -m venv $(SUB_VENV)/headless-tf21 && bash -c "source $(SUB_VENV)/headless-tf21/bin/activate && pip install -U pip setuptools wheel")
+#endif
 
 # Try to install different versions of Tensorflow.
 ifneq (, $(shell bash -c "source $(SUB_VENV)/headless-tf1/bin/activate && pip install tensorflow-gpu==1.15" >/dev/null 2>&1 && echo true))
 TENSORFLOW_1 := 1
 endif
-ifneq (, $(shell bash -c "source $(SUB_VENV)/headless-tf21/bin/activate && pip install tensorflow==2.1" >/dev/null 2>&1 && echo true))
-TENSORFLOW_2_1 := 2.1
-endif
+#ifneq (, $(shell bash -c "source $(SUB_VENV)/headless-tf21/bin/activate && pip install tensorflow==2.1" >/dev/null 2>&1 && echo true))
+#TENSORFLOW_2_1 := 2.1
+#endif
 
 else
 
 TENSORFLOW_1 := 1
-TENSORFLOW_2_1 := 2.1
+#TENSORFLOW_2_1 := 2.1
 
 endif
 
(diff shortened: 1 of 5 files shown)


stweil:
I wonder whether we really need `export PIP ?= pip3` and would prefer to replace `$(PIP)` by a simple `pip`. @bertsky, you added that macro in commit 7807ae609458b847c609c3a133467370034e3363. Are there use cases where it is helpful? It is also exported. Are there submodules which need it? I found `core` and `cor-asv-ann` which use `PIP`, but it looks like they also don't need it.

kba:
> I wonder whether we really need export PIP ?= pip3 and would prefer to replace $(PIP) by a simple pip

We don't need it any more, we did need it when we still had to support 2.7 under certain circumstances. In fact, using anything but `pip/pip3` for `$(PIP)` could cause inconsistencies if people override it with a pip for a different minor version of python than the venv.

stweil:
Now 3 of 5 Python versions complete the build: https://github.com/stweil/ocrd_all/actions/runs/1993530591.

The remaining failures occur when building `ocrd_kraken` (see pull request #288) and might be fixed in the CI running now.

Makefile
```diff
@@ -25,12 +23,54 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
...
+ifeq (0, $(MAKELEVEL))
+SUB_VENV = $(VIRTUAL_ENV)/sub-venv
+endif
 
 BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+ifeq (0, $(MAKELEVEL))
+ifeq ($(MAKECMDGOALS), all)
```
stweil, line 35: TODO: support also builds with several targets, for example `make all check`.

Makefile
```diff
@@ -25,12 +23,54 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
...
+CHECK_SUBENVS := true
+endif
+ifeq ($(MAKECMDGOALS), check)
```
stweil, line 38: TODO: support also builds with several targets, for example `make all check`.

Makefile
```diff
@@ -25,12 +23,54 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
...
+create_venv := $(shell $(PYTHON) -m venv $(VIRTUAL_ENV) && bash -c "source $(VIRTUAL_ENV)/bin/activate && pip install -U pip setuptools wheel"))
+endif
+ifeq ($(wildcard $(SUB_VENV)/headless-tf1),)
+create_venv := $(shell $(PYTHON) -m venv $(SUB_VENV)/headless-tf1 && bash -c "source $(SUB_VENV)/headless-tf1/bin/activate && pip install -U pip setuptools wheel")
+endif
+ifeq ($(wildcard $(SUB_VENV)/headless-tf21),)
+create_venv := $(shell $(PYTHON) -m venv $(SUB_VENV)/headless-tf21 && bash -c "source $(SUB_VENV)/headless-tf21/bin/activate && pip install -U pip setuptools wheel")
+endif
+
+# Try to install different versions of Tensorflow.
```
stweil, line 56: TODO: This takes some time when the installation is really done (or tried), so a log message would be useful.

stweil:
Now [all builds pass](https://github.com/stweil/ocrd_all/actions/runs/2008097919).

stweil:
The updated branch still passes for all builds: https://github.com/stweil/ocrd_all/actions/runs/2027885009.

Makefile
```diff
@@ -488,31 +535,19 @@ OCRD_ANYBASEOCR += $(BIN)/ocrd-anybaseocr-tiseg
...
 $(call multirule,$(OCRD_ANYBASEOCR)): ocrd_anybaseocr
-ifeq (0,$(MAKELEVEL))
-	$(MAKE) -B -o $< $(notdir $(OCRD_ANYBASEOCR)) VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
-	$(call delegate_venv,$(OCRD_ANYBASEOCR),$(SUB_VENV)/headless-tf2)
-ocrd_anybaseocr-check:
-	$(MAKE) check OCRD_MODULES=ocrd_anybaseocr VIRTUAL_ENV=$(SUB_VENV)/headless-tf2
-else
 	cd $< ; $(MAKE) patch-pix2pixhd
+	# Hack for Python 3.10 which fails to install ocrd-fork-pylsd 0.0.4 from PyPI.
+	. $(ACTIVATE_VENV) && pip install git+https://github.com/kba/pylsd.git
```
stweil, line 540: @kba, @bertsky, could you have a look why `pip3.10 install ocrd-fork-pylsd` fails while `pip install git+https://github.com/kba/pylsd.git` works fine?

      gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Isource/include -I/home/stweil/src/github/OCR-D/venv3.10-20220323/include -I/usr/local/include/python3.10 -c source/src/lsd.cpp -o build/temp.linux-x86_64-3.10/source/src/lsd.o
      source/src/lsd.cpp:77:10: fatal error: lsd.h: No such file or directory
         77 | #include <lsd.h>
            |          ^~~~~~~
      compilation terminated.

kba: I tried to build a manylinux binary wheel for 3.10. Can you try

```
pip install --only-binary :all: ocrd-fork-pylsd
```

and see if that works for you?

stweil: That works, and `pip3.10 install ocrd-fork-pylsd` works now, too.

stweil:
Latest run passes all builds: https://github.com/stweil/ocrd_all/actions/runs/2029562843.

stweil:
@bertsky, @kba, I compared the build rules here with those from git master and wrote the complete `make all` results in two separate virtual environments. The result shows that my build rules reduce the disk usage significantly, especially by reducing the total number of required virtual environments from four to two (one main venv, one sub venv). That would also result in faster builds and smaller docker images and so fix the time problems in CircleCI where the image uploads contribute a significant part.

```
stweil@ocr-02:~/src/github/OCR-D$ du -msc venv3.7-20220331-*
11605	venv3.7-20220331-master
6876	venv3.7-20220331-tensor
18480	total
stweil@ocr-02:~/src/github/OCR-D$ du -msc venv3.7-20220331-*/*
74	venv3.7-20220331-master/bin
594	venv3.7-20220331-master/build
2	venv3.7-20220331-master/build.log
11	venv3.7-20220331-master/include
2985	venv3.7-20220331-master/lib
1	venv3.7-20220331-master/lib64
3	venv3.7-20220331-master/libexec
1	venv3.7-20220331-master/locale
1	venv3.7-20220331-master/pyvenv.cfg
1	venv3.7-20220331-master/requirements.txt
103	venv3.7-20220331-master/share
7837	venv3.7-20220331-master/sub-venv
74	venv3.7-20220331-tensor/bin
594	venv3.7-20220331-tensor/build
1	venv3.7-20220331-tensor/build.log
11	venv3.7-20220331-tensor/include
4006	venv3.7-20220331-tensor/lib
1	venv3.7-20220331-tensor/lib64
3	venv3.7-20220331-tensor/libexec
1	venv3.7-20220331-tensor/locale
1	venv3.7-20220331-tensor/pyvenv.cfg
1	venv3.7-20220331-tensor/requirements.txt
103	venv3.7-20220331-tensor/share
2087	venv3.7-20220331-tensor/sub-venv
18480	total
```

stweil:
Now I also have timing and size results for a local `make docker-maximum-cuda-git ` (single-threaded):

```
OCR-D:master
0.60user 0.55system 56:52.93elapsed 0%CPU (0avgtext+0avgdata 64456maxresident)k
3848inputs+3024outputs (39major+10597minor)pagefaults 0swaps
ocrd/all                  maximum-cuda-git   d737c47c1285   2 hours ago     32.8GB

stweil:tensorflow
0.71user 0.58system 45:54.85elapsed 0%CPU (0avgtext+0avgdata 67572maxresident)k
2280inputs+2712outputs (5major+11208minor)pagefaults 0swaps
ocrd/all                  maximum-cuda-git   a8fc92d9f811   2 minutes ago   23.7GB
```

stweil:
This proof of concept is no longer up to date, so it can be closed.

PR #297, opened 2022-03-16
Add new macro PYTHON_VERSION

stweil:
It is used for special handling of several versions of Python,
here to disable ocrd_kraken with Python 3.10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,13 @@ export PKG_CONFIG_PATH
 OCRD_EXECUTABLES = $(BIN)/ocrd # add more CLIs below
 CUSTOM_DEPS = unzip wget python3-venv parallel git less # add more packages for deps-ubuntu below (or modules as preqrequisites)
 
+ifeq ($(PYTHON_VERSION),3.10)
+# Python 3.10.x does not work with current kraken.
+DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy ocrd_kraken
+else
 DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy
+endif
+
 # Default to all submodules, but allow overriding by user
 # (and treat the empty value as if it was unset)
 # opencv-python is only needed for aarch64-linux-gnu and other less common platforms,
@@ -201,10 +207,6 @@ $(SHARE)/opencv-python: opencv-python/setup.py | $(ACTIVATE_VENV) $(SHARE) $(SHA
 $(BIN)/ocrd: $(SHARE)/opencv-python
 endif
 
-ifeq ($(PYTHON_VERSION),3.10)
-# Python 3.10.x does not work with current kraken.
-override OCRD_MODULES := $(filter-out ocrd_kraken, $(OCRD_MODULES))
-endif
 ifneq ($(findstring ocrd_kraken, $(OCRD_MODULES)),)
 OCRD_EXECUTABLES += $(OCRD_KRAKEN)
 install-models: install-models-kraken
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,9 @@ BIN = $(VIRTUAL_ENV)/bin
 SHARE = $(VIRTUAL_ENV)/share
 ACTIVATE_VENV = $(VIRTUAL_ENV)/bin/activate
 
+# Get Python major and minor versions for some conditional rules.
+PYTHON_VERSION := $(shell $(PYTHON) -c 'import sys; print("%u.%u" % (sys.version_info.major, sys.version_info.minor))')
+
 define SEMGIT
 $(if $(shell sem --version 2>/dev/null),sem --will-cite --fg --id ocrd_all_git,$(error cannot find package GNU parallel))
 endef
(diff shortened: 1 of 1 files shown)


stweil:
This is an extract from PR #295 and can be reused in PR #289.

stweil:
We could use `PYTHON_VERSION` also as a hack to disable submodules which require old Tensorflow versions for newer versions of Python which don't provide those versions. That's much faster than the (cleaner) solution to check whether the required Tensorflow versions are available.

kba (approved):

Makefile
```diff
@@ -198,6 +201,10 @@ $(SHARE)/opencv-python: opencv-python/setup.py | $(ACTIVATE_VENV) $(SHARE) $(SHA
 $(BIN)/ocrd: $(SHARE)/opencv-python
 endif
 
+ifeq ($(PYTHON_VERSION),3.10)
+# Python 3.10.x does not work with current kraken.
+override OCRD_MODULES := $(filter-out ocrd_kraken, $(OCRD_MODULES))
+endif
```
bertsky, line 207: This defies several layers of module selection mechanisms that are already in place.

If the user chose to install the ocrd_kraken module, then it is wrong to silently ignore it.

The proper way to remove a module for a certain Python version _by default_ would be to include it in `DISABLED_MODULES`.

bertsky (changes requested):
> This is an extract from PR https://github.com/OCR-D/ocrd_all/pull/295 and can be reused in PR https://github.com/OCR-D/ocrd_all/pull/289.

#289 already came first with that recipe BTW.

Makefile
kba: So

```make
DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy
ifeq ($(PYTHON_VERSION),3.10)                            
# Python 3.10.x does not work with current kraken.       
DISABLED_MODULES += ocrd_kraken                          
endif                                                    
```

would still allow overriding with `OCRD_MODULES` but disables it by default for 3.10?

bertsky: Not quite. That would still override any explicit user (or Docker) setting for `DISABLED_MODULES`. But we only want to override the actual default:

```make
ifeq ($(PYTHON_VERSION),3.10)
# Python 3.10.x does not work with current kraken.
DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy ocrd_kraken
else
DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy
endif
```

stweil: My patch disables `ocrd_kraken` unconditionally for Python 3.10 (which is necessary because that combination does not work). The proposed alternatives don't do that. Therefore I don't think they are equally good or even better.

stweil: We could discuss how to handle submodules which are disabled automatically because they don't work. Instead of silently disabling them, there could be a report (for example at the end of the build process) about those disabled submodules.

bertsky: > My patch disables `ocrd_kraken` unconditionally for Python 3.10

exactly.

> (which is necessary because that combination does not work)

thus, let it fail (loudly)!

> Therefore I don't think they are equally good or even better.

As argued above, they are better (you did not address the arguments).

> We could discuss how to handle submodules which are disabled automatically because they don't work. Instead of silently disabling them, there could be a report (for example at the end of the build process) about those disabled submodules.

That would only complicate matters further, and goes against the user expectations of running `make`. If a module does not work in your environment (whatever the cause or kind), then the simplest and uniform answer is that module's build failing (while others can continue if running with `-k`). Also, the build is incremental, so as soon as the user fixed it (by disabling the module or installing by hand), one can continue.

Makefile
```diff
@@ -53,7 +56,13 @@ export PKG_CONFIG_PATH
 OCRD_EXECUTABLES = $(BIN)/ocrd # add more CLIs below
 CUSTOM_DEPS = unzip wget python3-venv parallel git less # add more packages for deps-ubuntu below (or modules as preqrequisites)
 
-DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_kraken clstm ocrd_ocropy
+ifeq ($(PYTHON_VERSION),3.10)
+# Python 3.10.x does not work with current kraken.
+DISABLED_MODULES ?= cor-asv-fst opencv-python ocrd_ocropy ocrd_kraken
```
stweil, line 61: That pattern no longer works as soon as I want to disable other submodules for other Python versions!

bertsky: > That pattern no longer works as soon as I want to disable other submodules for other Python versions!

If and when that happens, `make` still offers plenty of other syntax to achieve the same thing.

stweil: Then I need your help. How would that look like?

bertsky: For example, something along the lines of:
```make
DISABLED_MODULES ?= $(DEFAULT_DISABLED_MODULES)
DEFAULT_DISABLED_MODULES = cor-asv-fst opencv-python ocrd_ocropy
ifeq ($(PYTHON_VERSION),3.10)
# Python 3.10.x does not work with current kraken.
DISABLED_MODULES += ocrd_kraken
endif
ifeq (...)
DISABLED_MODULES += ...
endif
```

stweil: That would change the user interface, right? Instead of providing `DISABLED_MODULES` they would have to use `DEFAULT_DISABLED_MODULES`. Of course all documentation would require an update, too.

Would that be acceptable and better than my original solution?

bertsky: No, it would not. The user does not need to know anything about `DEFAULT_DISABLED_MODULES`. In fact, that is the whole point (to have a dynamic default, but allowing explicit user-defined opt-outs as already documented).

bertsky (approved):

PR #300, opened 2022-03-19
Shorten path for SUB_VENV and define it only on the first make level

stweil:
There is no need for the additional directory "local" (which can
also be confused with "locale" on the same level).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,9 @@ ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
 # to nested venv in recursive calls for modules
 # that have known dependency clashes with others
 export VIRTUAL_ENV ?= $(CURDIR)/venv
-SUB_VENV = $(VIRTUAL_ENV)/local/sub-venv
+ifeq (0, $(MAKELEVEL))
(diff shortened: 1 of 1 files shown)


stweil:
This is also part of pull request #295.

bertsky (approved):
Fine with me.

kba (commented):
We should provide a migration script/makefile target, otherwise people with a local installation will have gigabytes of obsolete data in `/local/sub-venv`, won't they?

stweil:
Do you think that people reuse their local venvs? I personally don't do so because it does not really work good to run incremental builds. Typically I create fresh venvs with the date in the name and run fresh builds. And from time to time I remove the old venvs.

A notice in the release notes would be sufficient in my opinion.

bertsky:
> A notice in the release notes would be sufficient in my opinion.

I concur. (Migrating venvs is not possible BTW – FS paths are fixed.)

kba:
> Do you think that people reuse their local venvs?

They do, sometimes even with a weekly cronjob. Just yesterday I told @lena-hinrichsen that `git pull; make all` should be enough after an update. But we can change that of course and tell people to build from scratch for every new release.

PR #299, opened 2022-03-18
Replace PIP macro by simple pip

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

PR #302, opened 2022-03-23
core install: pass sem into every pip call

bertsky:


stweil:
This conflicts with #299.

bertsky:
> This conflicts with #299.

I don't care TBH – we are still trying to solve the deployment misery (see CI failures for CUDA variant and recent `-j` switching), which is urgent.

This is necessary to get a true separation of `pip` calls between (possibly parallel) venvs.

kba:
> This conflicts with #299.

I think it's a counterargument to #299. And it seems to fix the issue at hand (race condition while installing core). Let's merge and hope for the best. If CD still fails, I will push the images manually.

kba:
> > This conflicts with #299.
> 
> I think it's a counterargument to #299.

Wrong, sry. We still don't need $(PIP) in the ocrd_all Makefile, only in dependent packages.

stweil:
I fixed the merge conflict in #299 (now for the second time), so it is ready for merging again.

PR #304, opened 2022-03-24
rebase ocrd_anybaseocr on pix2pixhd as pkg

bertsky:
integrates https://github.com/OCR-D/ocrd_anybaseocr/pull/91 right away

PR #305, opened 2022-03-25
:package: v2022-03-25

kba:


stweil:
@kba, @bertsky, this release is broken because of the latest changes for `ocrd_anybaseocr` (which fails CI, too).

stweil:
See CI fix #306.

bertsky:
@stweil 
> this release is broken because of the latest changes for `ocrd_anybaseocr` (which fails CI, too).

yeah – it was meant to be integrated via #304, which seems to have slipped through. Your #306 is only half of the story.

bertsky:
> Your #306 is only half of the story.

correction: since I did publish `pix2pixhd` on PyPI in the end, we don't even have to checkout the submodule. So the current solution should be fine. (Therefore I have closed #304.)

(CircleCI still fails with the nagging pip race, though – `OSError` on missing pkg `METADATA` files. Perhaps we should downgrade to `pip<21.3.1` – as `pip>=22`  is only available for Python>=3.7? Otherwise I am out of ideas what to try – to make `-j` and therefore the time limit work again.)

stweil:
> Otherwise I am out of ideas what to try

Drop CircleCI in favour of GitHub actions?

stweil:
@bertsky, @kba, we still need https://github.com/OCR-D/ocrd_anybaseocr/pull/92 to fix the CI errors.

bertsky:
> Drop CircleCI in favour of GitHub actions?

Your GH actions workflow does not build/upload Docker images, which is the main purpose of CI here.

stweil:
I only wrote a first action. Of course another action would have to be written to build and upload Docker images. That's what I suggested.

PR #307, opened 2022-03-26
Use qualified names for Docker images

stweil:
This allows using podman instead of docker.

### DIFF
--- a/.github/workflows/makedocker.yml
+++ b/.github/workflows/makedocker.yml
@@ -1,0 +1,36 @@
+# GitHub workflow for `make docker-maximum-cuda`.
+
+name: make docker-maximum-cuda
+
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+on:
+  # Trigger workflow in GitHub web frontend or from API.
+  workflow_dispatch:
+    inputs:
+      os:
+        description: 'Operating system'
+        required: true
+        default: any
+        type: choice
+        options:
+          - ubuntu-18.04
+          - ubuntu-20.04
+
+jobs:
+  make:
+    runs-on: ${{ github.event.inputs.os }}
+
+    env:
+      PYTHON_VERSION: '3.7'
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ env.PYTHON_VERSION }}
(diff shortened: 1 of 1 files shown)


bertsky:
> This allows using podman instead of docker.

Interesting. Could you elaborate on that? (Here or in the README)

stweil:
Podman is a drop-in replacement for Docker and included in Debian / Ubuntu. The main difference is that it does not need a daemon and also can be run rootless. By default it does not assume that an image comes from docker.io. Using a qualified name tells it where those images come from (and also works with the normal Docker).

bertsky:
I did not know that Podman is directly compatible with Docker images, and can pull from Drockerhub.

What about GPU usage and `nvidia-docker` daemon?

Linking `podman` to `docker` sounds good: Could you please edit the README to explain that possibility?

stweil:
Before promoting podman it is necessary to test it in the OCR-D context. For example I simply don't know how good it can be used with GPU. Therefore I'd wait with mentioning it in the README.

stweil:
Can I merge this PR?

kba:
>> Before promoting podman it is necessary to test it in the OCR-D context
> Can I merge this PR?

Are we certain that this won't break anything with Docker / DockerHub deployment? I'd feel more comfortable with merging if we had running CI.

stweil:
> > > Before promoting podman it is necessary to test it in the OCR-D context
> > > Can I merge this PR?
> 
> Are we certain that this won't break anything with Docker / DockerHub deployment? I'd feel more comfortable with merging if we had running CI.

The change is only used in `Dockerfile`. Building a local Docker image is sufficient for a test.

stweil:
See https://github.com/stweil/ocrd_all/runs/5758316217 which shows that the patch does not break the Docker image build.

bertsky:
It's like a default namespace. Even without it, you can see docker.io top level when pushing or pulling against DH.

bertsky (approved):

bertsky:
> Can I merge this PR?

I suggest placing the merge and release cycle in one hand whenever possible. That would be @kba. See [here](https://github.com/OCR-D/ocrd_all/pull/309#issuecomment-1083258001)

kba:
Sry, this should have been part of the release, will fix right away.

PR #311, opened 2022-04-05
Update Dockerfile, Makefile and GitHub action for Docker images

stweil:
The new GitHub action currently allows maintainers to select git branch, Ubuntu version, Python version, docker image variant and upload to docker.io yes/no when it is run on demand.

By default it builds the `minimum` docker image on Ubuntu 18.04 with Python 3.6 and does not upload to docker.io.

### DIFF
--- a/.github/workflows/makedocker.yml
+++ b/.github/workflows/makedocker.yml
@@ -108,13 +108,13 @@ jobs:
            df -h /
     - name: Make Docker image
       run: make docker-${{ github.event.inputs.docker-image }}
-    - name: Alias Docker images
-      run: docker tag ocrd/all:${{ github.event.inputs.docker-image }} ubma/ocrd_all:${{ github.event.inputs.docker-image }}
+    # - name: Alias Docker images
+    #  run: docker tag ocrd/all:${{ github.event.inputs.docker-image }} ocrd/all:${{ github.event.inputs.docker-image }}
     - name: Show Docker images
       run: docker images
     - name: Login to Docker Hub and push images to Docker Hub
       run: |
            if ${{ github.event.inputs.upload-docker-image }}; then
-             echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
-             docker push ubma/ocrd_all:${{ github.event.inputs.docker-image }}
+             echo ${{ secrets.DOCKERHUB_PASS }} | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+             docker push ocrd/all:${{ github.event.inputs.docker-image }}
            fi
--- a/.github/workflows/makedocker.yml
+++ b/.github/workflows/makedocker.yml
@@ -115,6 +115,6 @@ jobs:
     - name: Login to Docker Hub and push images to Docker Hub
       run: |
            if ${{ github.event.inputs.upload-docker-image }}; then
-             echo ${{ secrets.DOCKERHUB_PASS }} | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+             echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
              docker push ocrd/all:${{ github.event.inputs.docker-image }}
            fi
--- a/.github/workflows/makedocker.yml
+++ b/.github/workflows/makedocker.yml
@@ -63,6 +63,49 @@ jobs:
         python-version: ${{ env.PYTHON_VERSION }}
     - name: Show Python3 version
       run: python3 --version
+    - name: Show disk usage of Homebrew, Android and .NET
+      run: sudo du -mscx /home/linuxbrew /usr/local/lib/android /usr/share/dotnet 2>/dev/null || true
+    - name: Remove Docker images
+      run: |
+           df -h
+           docker images
+           docker rmi alpine:3.12 alpine:3.13 alpine:3.14
+           docker rmi buildpack-deps:stretch buildpack-deps:buster buildpack-deps:bullseye
+           docker rmi debian:9 debian:10 debian:11
+           docker rmi moby/buildkit:latest
+           docker rmi node:12-alpine node:14-alpine node:16-alpine
+           docker rmi node:12 node:14 node:16
+           if false; then # don't remove Ubuntu images
+           docker rmi ubuntu:16.04 ubuntu:18.04 ubuntu:20.04
+           fi
+           docker images
+           df -h /
+    - name: Remove unneeded Debian packages
+      run: |
+           if false; then # skip time consuming package uninstall
+           sudo apt-get install -y deborphan
+           deborphan -a | sort
+           sudo apt-get purge -y $(deborphan -a|fgrep main/cli-mono|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/database|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/devel|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/httpd|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/php|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/vcs|while read dummy package; do echo $package; done)
+           deborphan | sort
+           sudo du -mscx /* 2>/dev/null || true
+           sudo du -mscx /opt/* 2>/dev/null || true
+           sudo du -mscx /usr/* 2>/dev/null || true
+           df -h /
+           fi
+    - name: Remove Homebrew, Android and .NET
+      run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/makedocker.yml
```diff
@@ -10,27 +10,74 @@ on:
...
+    - name: Show Docker images (test)
+      run: docker images
+    - name: Alias Docker images
+      run: docker tag ocrd/all:${{ github.event.inputs.docker-image }} ubma/ocrd_all:${{ github.event.inputs.docker-image }}
+    - name: Show Docker images
+      run: docker images
+    - name: Login to Docker Hub and push images to Docker Hub
+      run: |
+           if ${{ github.event.inputs.upload-docker-image }}; then
+             echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
```
stweil, line 81: Those secrets must be provided in the OCR-D/ocrd_all settings if pushing to docker.io is desired.

.github/workflows/makedocker.yml
```diff
@@ -10,27 +10,74 @@ on:
...
-      run: sudo make deps-ubuntu PYTHON=python${{ env.PYTHON_VERSION}}
-    - name: Make docker-maximum-cuda
-      run: make docker-maximum-cuda PYTHON=python${{ env.PYTHON_VERSION }}
+      run: sudo make deps-ubuntu
+    - name: Make Docker image
+      run: make docker-${{ github.event.inputs.docker-image }}
+    - name: Show Docker images (test)
+      run: docker images
+    - name: Alias Docker images
+      run: docker tag ocrd/all:${{ github.event.inputs.docker-image }} ubma/ocrd_all:${{ github.event.inputs.docker-image }}
```
stweil, line 75: Update here before merging. My branch pushes to https://hub.docker.com/r/ubma/ocrd_all/tags.

stweil:
GitHub actions has disk storage limitations: https://github.com/stweil/ocrd_all/runs/5840702456. 😢 
So that is another challenge.

stweil:
The build environment for GitHub actions is packed with [more or less useful things](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md), which results in a root partition filled to 61 %:

    /dev/sda1        84G   51G   33G  61% /

Now the GitHub action starts with cleaning and removes Homebrew, Android and .NET. Result:

    /dev/sda1        84G   29G   55G  35% /

Like that [all Docker images](https://hub.docker.com/r/ubma/ocrd_all/tags?page=1&ordering=-name) are built now. I noticed that they are much smaller than the [Docker images from CircleCI](https://hub.docker.com/r/ocrd/all/tags?page=1&ordering=-name&name=m).

.github/workflows/makedocker.yml
stweil: The secrets now use the same names like CircleCI, so maybe that works directly for OCR-D.

.github/workflows/makedocker.yml
stweil: I disabled tagging of the docker images for now, so instead of pushing the same image twice with different tags it is only pushed once to docker.io.

stweil:
@bertsky, @kba, from my side this PR works and is ready for merging.

If desired, more triggers for the build can be added, and the matrix of build parameters which is currently presented to maintainers can be enhanced or modified in other ways.

.github/workflows/makedocker.yml
```diff
@@ -10,27 +10,111 @@ on:
...
+           sudo rm -rf /usr/share/dotnet # will release about 20GB if you don't need .NET
+           sudo du -mscx /* 2>/dev/null || true
+           df -h /
+    - name: Make Docker image
+      run: make docker-${{ github.event.inputs.docker-image }}
+    # - name: Alias Docker images
+    #  run: docker tag ocrd/all:${{ github.event.inputs.docker-image }} ocrd/all:${{ github.event.inputs.docker-image }}
+    - name: Show Docker images
+      run: docker images
+    - name: Login to Docker Hub and push images to Docker Hub
```
kba, line 115: Looks good, the only thing missing AFAICS is to alias `maximum-git` to `latest`.

.github/workflows/makedocker.yml
kba: OK, I'll set up the environment vars.

kba: Set in the [Dockerhub Deployment environment](https://github.com/OCR-D/ocrd_all/settings/environments/457360666/edit)

.github/workflows/makedocker.yml
```diff
@@ -10,27 +10,111 @@ on:
...
+           sudo apt-get purge -y $(deborphan -a|fgrep main/httpd|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/php|while read dummy package; do echo $package; done)
+           sudo apt-get purge -y $(deborphan -a|fgrep main/vcs|while read dummy package; do echo $package; done)
+           deborphan | sort
+           sudo du -mscx /* 2>/dev/null || true
+           sudo du -mscx /opt/* 2>/dev/null || true
+           sudo du -mscx /usr/* 2>/dev/null || true
+           df -h /
+           fi
+    - name: Remove Homebrew, Android and .NET
```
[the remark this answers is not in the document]
stweil: GitHub actions provides only two "runners" for Linux, namely ubuntu 18.04 and 20.04, and both of them are rather fat. I don't know the answer to your question. It's a decision made by the GitHub owners. Maybe they wanted to make everybody happy by providing as much as possible.

.github/workflows/makedocker.yml
stweil: Currently CircleCI also makes xxx an alias of xxx-git, so for example maximum-git and maximum are identical. Should this be done here, too? If yes, I'd remove all build targets without `-git`.

.github/workflows/makedocker.yml
stweil: I know, but I disabled the tagging because I wanted to see the difference of `maximum` (smaller because `/build` is removed) and `maximum-git` (full sources included, so can be used for development). I simply don't know why the same Docker image was offered under two (or even three with `latest`) tags.

.github/workflows/makedocker.yml
stweil: That link is only accessible for people with more rights, not for me.

kba: OK, I set `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD` as repository secrets - can you try again pls?

stweil: The code currently uses `DOCKERHUB_PASS` (as in CircleCI). Should I change that on my side to `DOCKERHUB_PASSWORD`, or do you want to provide `DOCKERHUB_PASS`?

.github/workflows/makedocker.yml
kba: The alias from `*um-git` to `*um` is because we changed the default to contain the git repositories and did not want to break existing setups by removing the `*um-git` variants.

The alias `maximum-git -> latest` is just for convenience and consistency with other Docker images, so that people can do `docker run ocrd/all` without any tag (which defaults to `latest`).

.github/workflows/makedocker.yml
kba: `maximum-git` should be aliased to `latest`. And yes, it should always point to `maximum-git`.

.github/workflows/makedocker.yml
kba: I prefer the longer version for consistency, `_PASSWORD` and `_USERNAME` instead of `_PASS` and `_USER`.

stweil: So do I. The [action](https://github.com/OCR-D/ocrd_all/runs/5866458186) is running now.

Note that I added another commit which removes not only `/build` but also `/.cache` which removes the image size further.

stweil: The action was successful and uploaded a new image (684.26 MB now instead of 754.21 MB with `/.cache`).

.github/workflows/makedocker.yml
stweil: Okay. I can add a commit which makes the `latest` alias and uploads it either now of after the merge (which I'd prefer).

So do you want to drop support for the `*um` images (although they are much smaller and sufficient for people who don't develop or debug code) and provide them only as aliases? The I could add another commit for that, too.

stweil:
@bertsky, can we merge?

bertsky:
Please give me more time.

bertsky (approved):
Thanks @stweil!

bertsky:
@stweil, I cannot see runs of `make docker` for anything other than minimum here on GH Actions, and my local build of maximum fails...

bertsky:
> I cannot see runs of `make docker` for anything other than minimum here on GH Actions, and my local build of maximum fails...

Ah, found it on [your fork](https://github.com/stweil/ocrd_all/runs/5851425728) – I guess my local failure was due to `DOCKER_PARALLEL=-j2`...

bertsky:
So currently we still hope for the CircleCI to get through automatically, otherwise we now can trigger the GH Action manually?

stweil:
Right, but we also can add automatic triggers to the GH action if necessary.

The next step will be adding the docker image tag for latest.
After that, I plan to prepare a pull request for builds with only two venvs.

PR #315, opened 2022-04-29
Update clean target

stweil:


stweil:
The CI failure is unrelated to the change here.

PR #316, opened 2022-04-29
Update ocrd_cis

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

PR #317, opened 2022-05-03
Update 2022 05 03

kba:
contains #314 (but with ocrd_cis in the main venv), #315, #316 plus the regular updates to the processor projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+Changed:
+
+  * `make clean`: remove sub-venvs and all semaphores, #315
+
 ## [v2022-03-30](https://github.com/OCR-D/ocrd_all/releases/v2022-03-30)
 
 ### [core](https://github.com/OCR-D/core) [6d359e7](https://github.com/OCR-D/core/commits/6d359e7)..[f9c2b6c](https://github.com/OCR-D/core/commits/f9c2b6c)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,20 @@ Changed:
   > * updating model directory
   > * Merge pull request #70 from qurator-spk/cneud-readme
 
+### [ocrd_cis](https://github.com/cisocrgroup/ocrd_cis) [ca15800](https://github.com/cisocrgroup/ocrd_cis/commits/ca15800)..[2cdfa79 (rewind)](https://github.com/cisocrgroup/ocrd_cis/commits/2cdfa79 (rewind))
+
+> Release: [v0.1.5](https://github.com/cisocrgroup/ocrd_cis/releases/v0.1.5)
+
+  > join_polygons (alpha shape): make more robust
+  > remove Calamari dependency (not used, only CLI callout)
+  > segment: for more robust bg separator detection, combine criteria of gradient maximum and percentile
+  > segment: rewrite separator detection…
+  > resegment: if method=lineest, then annotate baselines, too
+  > resegment (ccomps/baseline): propagate/spread twice to catch diacritics/punctuation, too
+  > re/segment: alpha shape: smaller α to avoid holes
+  > segment: fix lines2regions non-continguous partitions
+  > segment: annotate baselines, too
+
 ### [ocrd_detectron2](https://github.com/bertsky/ocrd_detectron2) [5bb90b1](https://github.com/bertsky/ocrd_detectron2/commits/5bb90b1)..[855fc3b](https://github.com/bertsky/ocrd_detectron2/commits/855fc3b)
 
 > Release: [v0.1.1](https://github.com/bertsky/ocrd_detectron2/releases/v0.1.1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,126 @@
 
 ## Unreleased
 
+## [v2022-05-03](https://github.com/OCR-D/ocrd_all/releases/v2022-05-03)
+
 Changed:
 
   * `make clean`: remove sub-venvs and all semaphores, #315
+
+### [core](https://github.com/OCR-D/core) [f9c2b6c](https://github.com/OCR-D/core/commits/f9c2b6c)..[ecdb840](https://github.com/OCR-D/core/commits/ecdb840)
+
+> Release: [v2.33.0](https://github.com/OCR-D/core/releases/v2.33.0)
+
+  > * :package: v2.33.0
+  > * :memo: changelog
+  > * Merge remote-tracking branch 'stweil/master'
+  > * :memo: changelog
+  > * Merge remote-tracking branch 'mexthecat/issues_838'
+  > * :memo: changelog
+  > * Merge remote-tracking branch 'origin/remove-assets-server'
+  > * :memo: changelog
+  > * Merge remote-tracking branch 'bertsky/workspace-altimg-retrieve-existing'
+  > * :memo: changelog
+  > * workspace bulk_add: fix automatic file_id
+  > * Simplify Makefile
+  > * Remove version restrictions for pip (no longer needed)
+  > * Always use apt --no-install-recommend
+  > * Merge branch 'master' of https://github.com/OCR-D/core
+  > * :memo: changelog
+  > * workspace.remove_file_group: pass recursive
+
+### [eynollah](https://github.com/qurator-spk/eynollah) [10f1ace](https://github.com/qurator-spk/eynollah/commits/10f1ace)..[ae1990a](https://github.com/qurator-spk/eynollah/commits/ae1990a)
+
+> Release: [v0.0.11](https://github.com/qurator-spk/eynollah/releases/v0.0.11)
+
+  > * Merge pull request #74 from cneud/main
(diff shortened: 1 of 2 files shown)


stweil (approved):
Thank you.

stweil:
CircleCI fails again because https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64 is broken. That's unrelated to the changes here, and all we can do is wait and hope that it will work again soon. Maybe repeating the check is sufficient.

kba:
> CircleCI fails again because https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64 is broken. That's unrelated to the changes here, and all we can do is wait and hope that it will work again soon. Maybe repeating the check is sufficient.

Yeah, I am debugging this right now. The docker image for the CUDA docker images have been updated yesterday, I assume with a fix to the issue with the signatures. However, our `ocrd/core-cuda` image needs rebuilding, which currently fails

```
apt-get -y install --no-install-recommends cuda-runtime-10-0 cuda-runtime-10-1 cuda-runtime-10-2 cuda-runtime-11-0 cuda-runtime-11-1 cuda-runtime-11-3 libcudnn7
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package libcudnn7
```

stweil:
@kba, it is fixed by running `curl https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub | sudo apt-key add -`.

kba:
> `curl https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub | sudo apt-key add -`.

The problem is that we need to fix it in the `ocrd/core-cuda` image, which I am trying to rebuild. That is in turn based on the updated nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu18.04, which does not have the GPG key problem but a new one:

```
apt-get -y install --no-install-recommends cuda-runtime-10-0 cuda-runtime-10-1 cuda-runtime-10-2 cuda-runtime-11-0 cuda-runtime-11-1 cuda-runtime-11-3 libcudnn7             
Reading package lists...                                                                                                                                    
Building dependency tree...                                                                                                                                                  
Reading state information...                                                                                                                                                 
E: Unable to locate package libcudnn7  
```

kba:
> That problem is gone as soon as you have installed the key from https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub and run `apt update`.

As I said, that GPG key problem does not occur with the docker image, that has been fixed upstream AFAICT. However, this updated `nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu18.04` does not contain libcudnn7 anymore, only libcudnn8. Neither @bertsky nor myself were sure why we even had that in. So I removed it and restarted the build nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu18.04 and let's all cross fingers that this fixes the build and does not cause new problems 🤞

bertsky:
By updating sbb_binarization to its current head, you dragged in TF2 into headless-tf1.

bertsky:
@kba I therefore suggest moving sbb_binarization and eynollah into top level venv (and removing their tf1nvidia recipe line).

bertsky:
Alas, 
```
tensorflow 2.5.0 has requirement tensorflow-estimator<2.6.0,>=2.5.0rc0, but you have tensorflow-estimator 2.4.0.
```

The problem again comes from the fixed (freezed / `pip-tool` generated) requirements.txt in ocrd_pc_segmentation. TF 2.6 requires tensorflow-estimator 2.6, but ocr4all-pixel-classifier insists on TF 2.5.

We have a few choices here:
1. asking @crater2150 to update his `ocr4all-pixel-classifier/requirements.in` once again – in this case removing the upper end in
```
tensorflow >= 2.0.0, <= 2.5.0
```
2. moving ocrd_pc_segmentation to its own dedicated venv (which will cost at least 2 GB extra, all the benefits of the recent changes go to waste)
3. disabling ocrd_pc_segmentation for the time being, since AFAICT no-one has actually been using it in OCR-D (please correct me, @chreul @maxnth)

cneud:
My prefered solution would be (3) for the benefit of being able to flexibly make a new ocrd_all release quickly without depending on ocrd_pc_segmentation. We can re-activate it later if we see the need or a fix has been implemented. The drawback is that we then also have to make this clear in the documentation (it seems to have only [this](https://ocr-d.de/en/workflows#available-processors-8) mention though - and we have to remove sbb-textline-detector there also anyway).

bertsky:
Okay, so just to see how we would fare with that choice in CI, I added a commit disabling ocrd_pc_segmentation for now. (We can drop it from the PR if it does not help or there is no consensus.)

Besides the WF guide, there's also a markdown checkbox to be unchecked in the README here.

bertsky:
Ah, limitless joy – here comes another surprise from the TF trolls for us:
```
ImportError: cannot import name 'LayerNormalization'
```
(failing in ocrd-calamari-recognize, ocrd-anybaseocr-layout-analysis, ocrd-anybaseocr-tiseg, ocrd-eynollah-segment, ocrd-sbb-binarize)
I'll investigate ways out of this via SSH...

bertsky:
I can immediately see multiple problems here:
- eynollah and sbb_binarization depend on `tensorflow-gpu` (newest: v2.4.4), while the other TF2 modules use `tensorflow` (newest: v2.6.2). pip is still dumb-ass and allows both but cripples the variant that gets installed first.
- sbb_binarization still constraints `h5py < 3` which conflicts with h5py requirements for TF2 (and should only be relevant for TF1 IIRC)
- ocrd_anybaseocr still depends on `keras` (v2.6.0), which should not be necessary in TF2

stweil:
> * sbb_binarization still constraints `h5py < 3` which conflicts with h5py requirements for TF2 (and should only be relevant for TF1 IIRC)

That's the current show stopper, see my comment here: https://github.com/qurator-spk/sbb_binarization/pull/35/files#r864201260.

cneud:
> * eynollah and sbb_binarization depend on tensorflow-gpu (newest: v2.4.4), while the other TF2 modules use tensorflow (newest: v2.6.2). pip is still dumb-ass and allows both but cripples the variant that gets installed first.
> * sbb_binarization still constraints h5py < 3 which conflicts with h5py requirements for TF2 (and should only be relevant for TF1 IIRC)

Oops. We'll look into those asap.

stweil:
> eynollah and sbb_binarization depend on tensorflow-gpu, while the other TF2 modules use tensorflow

It seems that meanwhile it is possible to install both `tensorflow~=2.5.0` and `tensorflow-gpu~=2.5.0` from PyPI at the same time without any conflict, but I am not sure whether that is reasonable or makes a difference.

stweil:
Allthough it would be good if `ocrd_pc_segmentation` could relax the upper limit `tensorflow<=2.5.0`, we can currently build with `tensorflow==2.5.0`. So there is no need to remove it.

bertsky:
> > eynollah and sbb_binarization depend on tensorflow-gpu, while the other TF2 modules use tensorflow
> 
> It seems that meanwhile it is possible to install both `tensorflow` and `tensorflow-gpu` (with compatible version 2.x.y) from PyPI at the same time without any conflict, but I am not sure whether that is reasonable or makes a difference.
> Allthough it would be good if `ocrd_pc_segmentation` could relax the upper limit `tensorflow<=2.5.0`, we can currently build with `tensorflow==2.5.0`. So there is no need to remove it.

No, it **does not** – I just explained it above!

> I removed the commit which did that.

...and one step backwards, again – I'm out.

bertsky:
force-push removing while others are **already** at the job – @stweil!!!

chreul:
> [...] ocr4all-pixel-classifier [...]
>     3. disabling ocrd_pc_segmentation for the time being, since AFAICT no-one has actually been using it in OCR-D (please correct me, @chreul @maxnth)
ocr4all-pixel-classifier is not directly related to ocr4all. we do not use ocrd_pc_segmentation and do not intend to do so in the future.

stweil:
I don't object removing OCR-D processors which nobody uses. Technically there is [no need to remove `ocrd_pc_segmentation`](https://github.com/stweil/ocrd_all/runs/6284885691).

@kba, I suggest to merge this PR – either with or without that processor.

kba:
> I don't object removing OCR-D processors which nobody uses. Technically there is [no need to remove `ocrd_pc_segmentation`](https://github.com/stweil/ocrd_all/runs/6284885691).
> 
> @kba, I suggest to merge this PR – either with or without that processor.

On it.

bertsky:
> Technically there is [no need to remove `ocrd_pc_segmentation`](https://github.com/stweil/ocrd_all/runs/6284885691).

There [is](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/797/workflows/26857926-d2f0-4f52-9070-bd9aac788b79/jobs/979)!

See [above](https://github.com/OCR-D/ocrd_all/pull/317#issuecomment-1116401307).

stweil:
All Docker builds in CircleCI now passed again. The longest run took 52 minutes, so was well below one hour.

PR #318, opened 2022-05-06
Add build on Ubuntu 22.04 to GitHub action

stweil:

### DIFF
--- a/.github/workflows/makeall.yml
+++ b/.github/workflows/makeall.yml
@@ -24,6 +24,7 @@ on:
           - any
           - ubuntu-18.04
           - ubuntu-20.04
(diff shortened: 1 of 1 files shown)


stweil:
Although it is still not documented, GitHub action now provides a runner with the fresh Ubuntu 22.04. That runner works fine with `ocrd_all` and Python 3.7 (I did not test other Python versions). It currently does not provide a running Docker service, so it cannot be used to build Docker images without further changes. Nevertheless I think that these first results indicate that OCR-D should consider switching from Ubuntu 18.04 / Python 3.6 to Ubuntu 22.04 / Python 3.7.

PR #319, opened 2022-05-06
dockerfile: use var-defined venv path throughout

bertsky:


stweil (changes requested):
Still missing: line 102. Otherwise okay.

stweil (approved):
Thanks.

kba (approved):

PR #322, opened 2022-05-11
Remove useless commands in Makefile

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -777,8 +777,6 @@ endif
 	# deps-ubuntu is run as root, but not in a Docker build were
 	# it is not needed and costs a lot of time.
 	test "$$HOME" == "/" || chown -R --reference=$(CURDIR) .git $(OCRD_MODULES)
-# prevent the sem commands during above module updates from imposing sudo perms on HOME:
-	test "$$HOME" == "/" || chown -R --reference=$(HOME) $(HOME)/.parallel
 
 deps-ubuntu-modules:
 	set -e; for dir in $^; do $(MAKE) -C $$dir deps-ubuntu PYTHON=$(PYTHON); done


Makefile
```diff
@@ -777,8 +777,6 @@ endif
 	# deps-ubuntu is run as root, but not in a Docker build were
 	# it is not needed and costs a lot of time.
 	test "$$HOME" == "/" || chown -R --reference=$(CURDIR) .git $(OCRD_MODULES)
-# prevent the sem commands during above module updates from imposing sudo perms on HOME:
-	test "$$HOME" == "/" || chown -R --reference=$(HOME) $(HOME)/.parallel
 
 deps-ubuntu-modules:
```
stweil, line 781: Do we really need the separate target `deps-ubuntu-modules`? I think we could simplify the code and avoid a recursion level without it.

stweil:
@kba, @bertsky, can we merge this PR? That is required before a PR for "solution 2" can be made.

Makefile
bertsky: > Do we really need the separate target `deps-ubuntu-modules`? I think we could simplify the code and avoid a recursion level without it.

Yes, we really need that separate target. Just read the code.

bertsky (commented):
Please elaborate why you think this line can be removed.

stweil:
> Please elaborate why you think this line can be removed.

See my comment above: those commands run in sudo context and "restore" the ownership to root.

kba:
We change the permissions back because `deps-ubuntu` must be called with `sudo`.

With #323 we can be sure that the `git` calls and consequently the `$(SEMGIT)` calls are run as the user owning the repository.

That leaves the question whether `deps-ubuntu` in the submodules' Makefiles does things beyond `apt-get install`, right? or is there something else I am missing.

So, how about I merge this change into #323 and test in the context of that PR that for a fresh clone, `sudo make deps-ubuntu` leaves all files in the repo and subrepo with the right permissions?

bertsky:
@stweil that line is meant to restore the permissions to `~/.parallel` after the sudo actions. (We run `parallel` as root and might create configs there – at least on first use.)

stweil:
> @stweil that line is meant to restore the permissions to `~/.parallel` after the sudo actions. (We run `parallel` as root and might create configs there – at least on first use.)

Yes, we run it as `root`. Therefore `~` is `/root`, and `parallel` puts its files there.

stweil:
> We change the permissions back because `deps-ubuntu` must be called with `sudo`.

We don't change any permissions for files created by `parallel` "back". Either it runs as `root` and writes files owned by root in the root home, or it runs as normal user and writes files owned by that user in the user's home. The comment which I removed, too, is simply misleading because it is wrong.

Issue #321, opened 2022-05-10
Native installation - got exception during 'sudo make deps-ubuntu'


stefanCCS:
I have tried to setup a new system (native) on Ubuntu 22.04.
I have followed https://ocr-d.de/en/setup
There `sudo make deps-ubuntu` has thrown the following exception:
```
make deps-ubuntu-modules
make[1]: Entering directory '/home/ocrdadmin/ocrd_all'
fatal: unsafe repository ('/home/ocrdadmin/ocrd_all' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /home/ocrdadmin/ocrd_all
set -e; for dir in core; do make -C $dir deps-ubuntu PYTHON=python3; done
make[2]: Entering directory '/home/ocrdadmin/ocrd_all/core'
make[2]: *** No rule to make target 'deps-ubuntu'.  Stop.
make[2]: Leaving directory '/home/ocrdadmin/ocrd_all/core'
make[1]: *** [Makefile:784: deps-ubuntu-modules] Error 2
make[1]: Leaving directory '/home/ocrdadmin/ocrd_all'
make: *** [Makefile:768: deps-ubuntu] Error 2

```
--> please clarify

cneud:
I've recently run into this as well, and because of the submodules, only this worked for me
`git config --global --add safe.directory "*"` (see also [here](https://marc.info/?l=git&m=164989570902912&w=2))

kba:
Ah, right, the submodules. A little less drastic workaround than marking all directories as safe:

```
git config --global --add safe.directory '/path/to/ocrd_all
git config --global --add safe.directory '/path/to/monorepo/ocrd_all/*'
```

But really, the best solution will be to avoid calling `git submodule` as root.

cneud:
> git config --global --add safe.directory '/path/to/monorepo/ocrd_all/*'

Have you tried? Because that did not solve it for me, only the global "*" did :(

stweil:
Instead of requiring users to run `sudo make deps-ubuntu`, we could also add `sudo` only where necessary (`sudo apt-get install`). That would be needed in a number of submodules, so takes some time. Nevertheless I think that would be a good solution.

bertsky:
> Instead of requiring users to run `sudo make deps-ubuntu`, we could also add `sudo` only where necessary (`sudo apt-get install`). That would be needed in a number of submodules, so takes some time. Nevertheless I think that would be a good solution.

[...]

PR #323, opened 2022-05-13
Makefile: always run git as the user who owns the repo

kba:
This replaces calls to `git` with `sudo -u $(stat '%U' .) git` so git is always run as the user who owns the git repository, as discussed in https://github.com/OCR-D/ocrd_all/issues/321#issuecomment-1125485153.

@bertsky Will this interfere with the `$(SEMGIT)` semaphore?

Perhaps we should add the https://launchpad.net/~git-core/+archive/ubuntu/ppa to our CI apt sources, so we can detect any future breaking behavior for new git releases?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ GIT_RECURSIVE = # --recursive
 GIT_DEPTH = # --depth 1
 # Required and optional Tesseract models.
 ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
+# Always run git as the user who owns the ocrd_all repo
+GIT = sudo -u $$(stat -c '%U' .) git
 
 # directory for virtual Python environment
 # (but re-use if already active); overridden
@@ -73,7 +75,7 @@ DISABLED_MODULES ?= $(DEFAULT_DISABLED_MODULES)
 # opencv-python is only needed for aarch64-linux-gnu and other less common platforms,
 # so don't include it by default.
 ifeq ($(strip $(OCRD_MODULES)),)
-override OCRD_MODULES := $(filter-out $(DISABLED_MODULES),$(shell git submodule status | while read commit dir ref; do echo $$dir; done))
+override OCRD_MODULES := $(filter-out $(DISABLED_MODULES),$(shell $(GIT) submodule status | while read commit dir ref; do echo $$dir; done))
 endif
 
 # `all` is too much for a default, and `ocrd` is too little
@@ -143,9 +145,9 @@ modules: $(OCRD_MODULES)
 ifneq (,$(wildcard .git))
 ifneq ($(NO_UPDATE),1)
 $(OCRD_MODULES): always-update
-	$(SEMGIT) git submodule sync $(GIT_RECURSIVE) $@
-	if git submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
-		$(SEMGIT) git submodule update --init $(GIT_RECURSIVE) $(GIT_DEPTH) $@ && \
+	$(SEMGIT) $(GIT) submodule sync $(GIT_RECURSIVE) $@
+	if $(GIT) submodule status $(GIT_RECURSIVE) $@ | grep -qv '^ '; then \
+		$(SEMGIT) $(GIT) submodule update --init $(GIT_RECURSIVE) $(GIT_DEPTH) $@ && \
 		touch $@; fi
 endif
 endif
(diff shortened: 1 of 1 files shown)


stweil (commented):
This looks like a clever solution, but only works for users with sudo rights. All others would get an error:

`n.n. is not in the sudoers file.  This incident will be reported."

sudo is not clever enough to accept being run from a user who does not want to change the UID.

Makefile
```diff
@@ -17,6 +17,8 @@ GIT_RECURSIVE = # --recursive
 GIT_DEPTH = # --depth 1
 # Required and optional Tesseract models.
 ALL_TESSERACT_MODELS = eng equ osd $(TESSERACT_MODELS)
+# Always run git as the user who owns the ocrd_all repo
+GIT = sudo -u $$(stat -c '%U' .) git
```
stweil, line 21: Using `sudo` conditionally here (only when running as `root`) would fix this PR.

stweil (changes requested):

Makefile
stweil, line 21:
```suggestion
# root must run git as the user who owns the working directory.
ifeq ($(USER),root)
GIT = sudo -u $$(stat -c '%U' .) git
else
GIT = git
endif
```

stweil: `GIT = sudo -u "#$$(stat -c '%u' .)" git` is a little bit more efficient because it avoids translating of a user id to user name and back to user id.

bertsky: @stweil looks like you made a typo here – what's that about the hash sign?
@kba in my original proposal I used `%u` (mind the small caps) for that same reason

bertsky:
> Will this interfere with the `$(SEMGIT)` semaphore?

No, not at all.

Makefile
stweil: > looks like you made a typo here – what's that about the hash sign?

That's not a typo, but required for a numeric user id.

bertsky: > Using `sudo` conditionally here (only when running as `root`) would fix this PR.

Right. Superuser should always have access to `sudo`, and the others are unlikely to need it.

stweil:
> Another place that should be edited: the post-portem of the `deps-ubuntu` down below. (We should be able to get rid of the existing `chown` recipes.)

That's correct, but would need more review efforts or tests to see whether other commands still create files with root ownership. I suggest to address that separately.

stweil:
@kba, I don't think that a `SUDO` macro is needed. I'd prefer to remove the corresponding commit.

Makefile
kba: It's either `sudo -u '#$(stat -c '%u' .) git'` or `sudo -u '$(stat -c '%U' .) git`. I personally prefer the latter, i.e. the username. But if that is a problem in some cases, numerical is also fine.

stweil: Using the user's name isn't a problem in any normal environment (`/etc/passwd`must be readable), and using the numerical variant would only be a very minor optimisation. So any of both variants is fine for me.

Makefile
```diff
@@ -143,18 +150,18 @@ modules: $(OCRD_MODULES)
...
 endif
 endif
 
 deinit: clean
 .PHONY: deinit
 deinit:
-	git submodule deinit --all # --force
-	git submodule status | while read stat dir ver; do rmdir $$dir; done
+	$(GIT) submodule deinit --all # --force
+	$(GIT) submodule status | while read stat dir ver; do rmdir $$dir; done
```
stweil, line 164: I see no reason to change those lines and would not do it.

Makefile
```diff
@@ -820,7 +827,7 @@ docker%-cuda docker%-cuda-git: DOCKER_BASE_IMAGE = docker.io/ocrd/core-cuda
 docker%: Dockerfile $(DOCKER_MODULES)
 	docker build \
 	--build-arg BASE_IMAGE=$(DOCKER_BASE_IMAGE) \
-	--build-arg VCS_REF=$$(git rev-parse --short HEAD) \
+	--build-arg VCS_REF=$$($(GIT) rev-parse --short HEAD) \
```
stweil, line 830: Nor is it necessary or useful to change this line.

stweil, line 830: Nor is it necessary or useful to change this line.

stweil (changes requested):

Makefile
kba: cbdb5ae

Makefile
kba: cbdb5ae

kba:
Superseded by #324

PR #324, opened 2022-05-14
Force running git as sudo user in sudo make deps-ubuntu

stweil:


stweil:
This pull request includes #322 and replaces #323.

kba (approved):
Fine with me, if that is agreeable to all. @bertsky?

stweil:
Ping? We should prepare a new release which works with Ubuntu 22.04 out of the box.

Makefile
```diff
@@ -13,6 +13,13 @@ PIP_OPTIONS_E = $(filter-out -e,$(PIP_OPTIONS))
 NO_UPDATE ?= 0
 # Set to non-empty to try running all executables with --help / -h during make check
 CHECK_HELP ?=
+GIT = git
+ifdef SUDO_USER
+ifneq ($(USER),$(SUDO_USER))
```
bertsky, line 18: I don't think it's safe to use `USER` here (as a reference for the target user). The manual says:

>      USER             Set to the same value as LOGNAME, described above.

>     LOGNAME          Set to the login name of the target user when the -i option is specified, when the set_logname option is enabled
>                      in sudoers or when the env_reset option is enabled in sudoers (unless LOGNAME is present in the env_keep list).

This means that it may or may not be set. I recommend measuring directly:
```suggestion
ifneq ($(shell id -un),$(SUDO_USER))
```

bertsky (commented):
I like utilizing `SUDO_USER` much better than our previous attempts. But `sudo` does not guarantee setting `USER` for us.

Makefile
bertsky: So why use the `ifneq` at all?

stweil: It avoids using `sudo` when it is clearly not needed.

kba (approved):

PR #325, opened 2022-05-16
Disable modules which require old tensorflow-gpu unless it is available

stweil:


kba (approved):
Fine by me, but let's make sure @bertsky is on board since these are his projects.

stweil:
@bertsky. do you have any comments? Can I merge this pull request?

stweil:
Ping?

bertsky (approved):
So be it.

(I hope we can delegate to module-containers soon, anyway.)

PR #332, opened 2022-08-15
update 2022-08-15

kba:


bertsky:
CUDA 11.4 is out, so we should probably update the `docker-cuda` and `cuda-ubuntu` rules of core soon. But let's merge these long-awaited changes first.

PR #334, opened 2022-08-16
Update GitHub actions

stweil:


kba (approved):

stweil:
This commit should fix a warning from GitHub actions:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-python, actions/checkout
```

PR #335, opened 2022-10-24
:package: v2022-10-25

kba:
This is a fairly significant update because I let too much time pass since the last release.

However, the next release will be fairly soon because we have two potential new sets of processors and a bugfix release for core this week.

But first, we need a new release with up-to-date core and the released bug fixes.

I will prepare a summary of the changes and do the actual release tomorrow, Oct 25.


bertsky:
:crossed_fingers:

PR #336, opened 2022-10-24
Don't try running unavailable Python 3.6 on Ubuntu 22.04

stweil:

### DIFF
--- a/.github/workflows/makeall-linux.yaml
+++ b/.github/workflows/makeall-linux.yaml
@@ -20,6 +20,9 @@ on:
 
 jobs:
   make:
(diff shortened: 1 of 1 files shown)


stweil:
See [example of failing action](https://github.com/OCR-D/ocrd_all/actions/runs/3314546937/jobs/5473976678).

stweil:
The patch does not work as expected, therefore converted to draft.

stweil:
The pull request can be dropped as soon as Python 3.6 was removed from the code.

stweil:
This pull request is no longer needed because support for Python 3.6 was dropped.

PR #337, opened 2022-10-25
tessdata: only use prefix/module dir

bertsky:
This adapts to https://github.com/OCR-D/ocrd_tesserocr/pull/187, which will be integrated with https://github.com/OCR-D/ocrd_all/pull/335, on the data/model side: `install-tesseract` and `install-tesseract-models` used to be split between OCR-D resource location and Tesseract tessdata prefix. Since this has become one again, there is no need for elaborate duplication and customization. We simply let the processor and resmgr use the compile-time tessdata/module dir, which is based on the venv as prefix.

Also updates some of the executable lists of other modules.

Proposed course of action: merge this branch into #335 (after additional fixes to core have been merged there).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,6 @@ Variables:
 	PIP_OPTIONS: extra options for the `pip install` command like `-q` or `-v` or `-e`
 	TESSERACT_MODELS: list of additional models/languages to download for Tesseract. Default: "$(ALL_TESSERACT_MODELS)"
 	TESSERACT_CONFIG: command line options for Tesseract `configure`. Default: "$(TESSERACT_CONFIG)"
-	TESSDATA: directory path where to install Tesseract models. Default (based on XDG_DATA_HOME): "$(TESSDATA)"
 EOF
 endef
 export HELP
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -411,6 +412,7 @@ OCRD_TESSEROCR := $(BIN)/ocrd-tesserocr-binarize
 OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-crop
 OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-deskew
 OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-recognize
+OCRD_TESSEROCR += $(BIN)/ocrd-tesserocr-segment
```
kba, line 415: also appreciate these fixes.

README.md
```diff
@@ -333,14 +333,16 @@ This table lists which tag contains which module:
...
-| ocrd_olahd_client           | -         | ☑        | ☑         |
+| ocrd_olahd_client           | ☑         | ☑        | ☑         |
 | ocrd_olena                  | -         | ☑        | ☑         |
 | ocrd_segment                | -         | ☑        | ☑         |
 | tesseract                   | -         | ☑        | ☑         |
 | ocrd_anybaseocr             | -         | -        | ☑         |
-| ocrd_kraken                 | -         | -        | -         |
+| ocrd_detectron2             | -         | -        | ☑         |
+| ocrd_doxa                   | -         | -        | ☑         |
+| ocrd_kraken                 | -         | -        | ☑         |
```
kba, line 343: ... and these.

PR #355, opened 2023-03-16
update 2023-03-16

kba:

PR #356, opened 2023-03-16
update 2023-03-16

kba:
OK, now with cleaned up commit history, discussion is in #355

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,7 +7,8 @@ local.mk
 # not needed in Ubuntu:
 opencv-python
 # only needed for testing:
-*/assets
+**/assets/*
 # avoid interference with host-local builds:
-ocrd_olena/olena-git/build
-tesseract/build
+**/build/*
+*/.eggs/*
+*/*.egg-info
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,199 @@
 # Changelog
 
+## [v2023-03-16](https://github.com/OCR-D/ocrd_all/releases/v2023-03-16)
+
+### [core](https://github.com/OCR-D/core) [ee92cfc](https://github.com/OCR-D/core/commits/ee92cfc)..[cbe83ab](https://github.com/OCR-D/core/commits/cbe83ab)
+
+> Release: [v2.47.4](https://github.com/OCR-D/core/releases/v2.47.4)
+
+  > * :package: v2.47.4
+  > * :memo: changelog
+  > * Merge pull request #1011 from OCR-D/drop-fontgroup-resources
+  > * :package: v2.47.3
+  > * :memo: changelog
+  > * Dockerfile: reintroduce python3-pip so "pip install -U pip" succeeds again
+  > * :package: v2.47.2
+  > * :memo: changelog
+  > * Merge branch 'pr/986'
+  > * :package: v2.47.1
+  > * :memo: changelog
+  > * Merge pull request #1004 from stweil/fix-dockerfile
+  > * Merge pull request #1003 from OCR-D/fix-docker-venv
+  > * :package: v2.47.0
+  > * require importlib_resources for python <= 3.8, #996
+  > * :memo: changelog
+  > * resmgr: drop anybaseocr models from resource list
+  > * :memo: changelog
+  > * Merge branch 'pr/999'
+  > * :memo: changelog
+  > * ocrd_utils: adapt to newer importlib.resources API
+  > * Merge pull request #994 from OCR-D/fix-scrutinizer
+  > * :memo: changelog
+  > * Merge pull request #977 from OCR-D/fix-add-agent
+  > * :memo: changelog
+  > * Merge branch 'pr/993'
+  > * :memo: changelog
+  > * Merge pull request #991 from bertsky/fix-resmgr-download-mimetype
+  > * :memo: changelog
+  > * Merge pull request #985 from OCR-D/fix-917
+  > * :package: v2.46.0
+  > * :memo: changelog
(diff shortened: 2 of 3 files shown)


kba:
CI fails but that might be due to the test being built based on the `ocrd/core-cuda` image, which is now very outdated.

I'm building the images locally now and if the build succeeds, will push the non-cuda images manually. Not ideal but since I will be AFK for a few days and this is a substantial update, I want to update what I can today.

bertsky:
> ` [update ocrd_keraslm](/OCR-D/ocrd_all/pull/356/commits/682324b9e7c2d8ff612b4a2ac63268ca17a914a5) `

oops, that was not enough:

```
tensorflow-gpu 1.15.5 has requirement numpy<1.19.0,>=1.16.0, but you have numpy 1.19.5
```

bertsky:
> > `[update ocrd_keraslm](/OCR-D/ocrd_all/pull/356/commits/682324b9e7c2d8ff612b4a2ac63268ca17a914a5)`
> 
> oops, that was not enough:
> 
> ```
> tensorflow-gpu 1.15.5 has requirement numpy<1.19.0,>=1.16.0, but you have numpy 1.19.5
> ```

Damn. It's even worse. The above requirement comes from PyPI TF 1.15 (only Py36 and Py37). But on Nvidia TF 1.15 (only Py37 and Py38) we have:
```
   tensorflow-gpu 1.15.5+nv23.2 depends on numpy<1.24 and >=1.22.0; python_version >= "3.7"
```
But `>=1.22` itself was incompatible with the older h5py we need here...

kba:
OK, so, build is still running for medium and maximum, if they finish successfully, I'll push them tomorrow morning. minimum is updated.

bertsky:
> But `>=1.22` itself was incompatible with the older h5py we need here...

That turned out not true. There's a window of `<1.24 >=1.22` for `tensorflow-gpu==1.15.5+nv23.2`. But h5py seems to be fine (confirmed by processing). There was another glitch with Protobuf, but not it seems to work for Py37 and Py38. Please update to master (https://github.com/OCR-D/ocrd_keraslm/commit/9e3f5a06b8efb706f8f1ac1c172fa5809ad6bab9).

bertsky:
> There's a window of `<1.24 >=1.22` for `tensorflow-gpu==1.15.5+nv23.2`. But h5py seems to be fine (confirmed by processing).

The problem is that here in ocrd_all, numpy does get updated to 1.24 by other modules, which have no restriction on its version and therefore indiscriminately pull the newest version. For example via shapely:
```
Collecting numpy>=1.14
...
tensorflow-gpu 1.15.5+nv23.2 requires numpy<1.24,>=1.22.0; python_version >= "3.7", but you have numpy 1.24.2 which is incompatible.
```

So again, this is pip playing dumb. Not sure how to fix this, except with our typical posterior workaround:
https://github.com/OCR-D/ocrd_all/blob/d8830fb2daacc5c90347091c52f9e2ed574e41c9/Makefile#L615

bertsky:
Not sure if this is new:
```
ocrd-dinglehopper -h
  File "/usr/local/bin/ocrd-dinglehopper", line 5, in <module>
    from qurator.dinglehopper.ocrd_cli import ocrd_dinglehopper
ModuleNotFoundError: No module named 'qurator.dinglehopper'
```

Dinglehopper was installed with `pip install -e` ...

bertsky:
> Not sure if this is new:
> 
> ```
> ocrd-dinglehopper -h
> Traceback (most recent call last):
>   File "/usr/local/bin/ocrd-dinglehopper", line 5, in <module>
>     from qurator.dinglehopper.ocrd_cli import ocrd_dinglehopper
> ModuleNotFoundError: No module named 'qurator.dinglehopper'
> ```

@mikegerber This is caused by https://github.com/qurator-spk/dinglehopper/issues/76.

To get it working again, I have to revert both https://github.com/qurator-spk/dinglehopper/commit/b4ac24ac9dc3f95fbb719d9eac1c1eeb904435c0 and https://github.com/qurator-spk/dinglehopper/commit/833efa37da63dfff70dd9cebec7eb112d54f79b1.

Note that the [documentation](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) explicitly warns that every package in the namespace must follow the (same) convention.

(BTW, eynollah still uses the pkg_resources style namespace pkg.)

mikegerber:
> > Not sure if this is new:
> > ```
> > ocrd-dinglehopper -h
> > Traceback (most recent call last):
> >   File "/usr/local/bin/ocrd-dinglehopper", line 5, in <module>
> >     from qurator.dinglehopper.ocrd_cli import ocrd_dinglehopper
> > ModuleNotFoundError: No module named 'qurator.dinglehopper'
> > ```
> 
> @mikegerber This is caused by [qurator-spk/dinglehopper#76](https://github.com/qurator-spk/dinglehopper/issues/76).
> 
> To get it working again, I have to revert both [qurator-spk/dinglehopper@b4ac24a](https://github.com/qurator-spk/dinglehopper/commit/b4ac24ac9dc3f95fbb719d9eac1c1eeb904435c0) and [qurator-spk/dinglehopper@833efa3](https://github.com/qurator-spk/dinglehopper/commit/833efa37da63dfff70dd9cebec7eb112d54f79b1).
> 
> Note that the [documentation](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) explicitly warns that every package in the namespace must follow the (same) convention.
> 
> (BTW, eynollah still uses the pkg_resources style namespace pkg.)

This is why I called this CAJ (computer-aided Jenga)...

PR #358, opened 2023-03-16
Fix GitHub action for docker build

stweil:
- Remove the time consuming calculation of used disk space.
- Remove all container images (more robust than removing single images).


stweil:
@kba, please include this PR also in the next release.

stweil:
Removing all docker images would have been the simplest solution. And it works because the required Ubuntu image (and only that one) will simply be installed when required. That does not take much time.

bertsky:
> Removing all docker images would have been the simplest solution. And it works because

It does not work. I have tried that elsewhere. Also, as the names keep changing, opt-in is not predictable.

PR #361, opened 2023-03-28
enable running --help for make check

bertsky:
We should have activated this long ago. It's not a true smoke test, but at least we know whether the executable can be run at module level at all.


bertsky:
Do we still want that? We have `CHECK_HELP=1` in the Github Actions config now. So either we add it to CircleCI `build-pull-request` as well, or we adopt it everywhere (this PR).

kba:
Let's have the `--help` check everywhere, we want to know if that fails in all CI.

PR #362, opened 2023-03-28
add rule for ocrd-tool-all.json, reduce image size, fix+update modules, fix CUDA

bertsky:
in lieu of https://ocr-d.de/js/ocrd-all-tool.json, this generates the file dynamically

(to be used locally, or as part of CI – e.g. storing as artifact)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -594,12 +594,15 @@ endef
 
 # Workaround for missing prebuilt versions of TF<2 for Python==3.8
 # todo: find another solution for 3.9, 3.10 etc
+# https://docs.nvidia.com/deeplearning/frameworks/tensorflow-wheel-release-notes/tf-wheel-rel.html
 # Nvidia has them, but under a different name, so let's rewrite that:
+# (hold at nv22.12, because newer releases require CUDA 12, which is not supported by TF2,
+#  and therefore not in our ocrd/core-cuda base image yet)
 define pip_install_tf1nvidia =
 . $(ACTIVATE_VENV) && if test $(PYTHON_VERSION) = 3.8 && ! pip show -q tensorflow-gpu; then \
 	$(SEMPIP) pip install nvidia-pyindex && \
 	pushd $$(mktemp -d) && \
-	$(SEMPIP) pip download --no-deps "nvidia-tensorflow!=1.15.5+nv23.3" && \
+	$(SEMPIP) pip download --no-deps "nvidia-tensorflow==1.15.5+nv22.12" && \
 	for name in nvidia_tensorflow-*.whl; do name=$${name%.whl}; done && \
 	$(PYTHON) -m wheel unpack $$name.whl && \
 	for name in nvidia_tensorflow-*/; do name=$${name%/}; done && \
--- a/Makefile
+++ b/Makefile
@@ -110,24 +110,41 @@ cat <<"EOF"
 Rules to download and install all OCR-D module processors
 from their source repositories into a single virtualenv.
 
-Targets:
+Targets (general):
 	help: show this message
 	show: list the venv path and all executables (to be) installed
-	check: verify that all executables are runnable and the venv is consistent
-	testcuda: verify that CUDA is available for Tensorflow and Pytorch
+
+Targets (module management):
 	modules: download all submodules to the managed revision
-	ocrd-all-tool.json: generate union of ocrd-tool.json for all executables of all modules
+	deinit: clean, then deinit and rmdir all submodules
+	tidy: clean, then deinit opencv-python and git-clean all submodules
+	      (WARNING: potential data loss; if unsure, try with `make -n` and `git clean -n`)
+
+Targets (system dependencies, may need root privileges):
+	deps-ubuntu: install all system dependencies of all modules
+	deps-cuda: install CUDA toolkit and libraries (via micromamba and nvidia-pyindex)
+
+Targets (build and installation into venv):
 	all: install all executables of all modules
 	ocrd: only install the virtual environment and OCR-D/core packages
 	install-tesseract: only build and install Tesseract (with TESSERACT_MODELS)
-	install-tesseract-training: build and install Tesseract training tools
-	install-models: download commonly used models to appropriate locations
+	install-tesseract-training: also build and install Tesseract training tools
+	fix-cuda: workaround for non-conflicting CUDA libs after installation
 	clean: remove the virtual environment directory, and make clean-*
 	clean-tesseract: remove the build directory for tesseract
 	clean-olena: remove the build directory for ocrd_olena
-	tidy: clean, then deinit opencv-python and git-clean all submodules
-	      (WARNING: potential data loss; if unsure, try with `make -n` and `git clean -n`)
-	deinit: clean, then deinit and rmdir all submodules
+
+Targets (testing):
+	check: verify that all executables are runnable and the venv is consistent
+	test-core: verify ocrd via core module regression tests
+	test-cuda: verify that CUDA is available for Tensorflow and Pytorch
+	test-workflow: verify that most executables work correctly via test runs on test data
+
+Targets (auxiliary data):
+	ocrd-all-tool.json: generate union of ocrd-tool.json for all executables of all modules
+	install-models: download commonly used models to appropriate locations
+
+Targets (build of container images):
 	docker: (re)build a docker image including all executables
 	dockers: (re)build docker images for some pre-selected subsets of modules
 
@@ -141,6 +158,7 @@ Variables:
 	TMPDIR: path to use for temporary storage instead of the system default
 	PYTHON: name of the Python binary
 	PIP_OPTIONS: extra options for the `pip install` command like `-q` or `-v` or `-e`
+	CHECK_HELP: set to `1` to also check each executable can generate help output
 	TESSERACT_MODELS: list of additional models/languages to download for Tesseract. Default: "$(ALL_TESSERACT_MODELS)"
 	TESSERACT_CONFIG: command line options for Tesseract `configure`. Default: "$(TESSERACT_CONFIG)"
 EOF
@@ -211,6 +229,10 @@ CUSTOM_DEPS += python3 imagemagick libgeos-dev
 $(BIN)/ocrd: | $(ACTIVATE_VENV)
 	. $(ACTIVATE_VENV) && $(SEMPIP) pip install $(PIP_OPTIONS_E) ocrd ocrd_network
 endif
+
+.PHONY: test-core
+test-core: core $(BIN)/ocrd
+	. $(ACTIVATE_VENV) && $(MAKE) -C $< deps-test test
 
 # Convert the executable names (1) to a pattern rule,
 # so that the recipe will be used with single-recipe-
@@ -706,6 +728,7 @@ ifeq (0,$(MAKELEVEL))
 endif
 %-check: ;
 
+.PHONY: testcuda test-cuda test-assets test-workflow
 # ensure shapely#1598 workaround works
 # ensure CUDA works for Torch and TF
 testcuda test-cuda: $(ACTIVATE_VENV)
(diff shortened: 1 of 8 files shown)


bertsky:
sry for the noise – just wanted to rebase to master, so CI runs through

bertsky:
Note: In 3bc8d6a, I modified @stweil's Github Action workflow for Docker – see detailed commit msg.

I [triggered](https://github.com/bertsky/ocrd_all/actions/runs/4547695233) it for minimum (without Dockerhub or Github push, because that would not work from my fork anyway) to see if cleanup, Docker build and artifact uploading works.

bertsky:
Note: [failure of normal (Circle) CI](https://app.circleci.com/pipelines/github/bertsky/ocrd_all/12/workflows/fc97e34c-7c2d-45f4-84bc-acea4e53f018/jobs/33) seems to be an independent, very recent [problem coming from nvidia-tensorflow](https://github.com/NVIDIA/tensorflow/issues/85).

bertsky:
> I [triggered](https://github.com/bertsky/ocrd_all/actions/runs/4548662339) it for minimum (without Dockerhub or Github push, because that would not work from my fork anyway) to see if cleanup, Docker build and artifact uploading works.

It does give us an artifact [ocrd-all-tool.json](https://github.com/bertsky/ocrd_all/suites/11875223009/artifacts/621111484), but unfortunately, it's always zipped, and therefore cannot be linked directly. IIRC this is a restriction by Github Actions API.

I'll now [try to run](https://github.com/bertsky/ocrd_all/actions/runs/4548725892) the opposite of the spectrum – maximum-cuda-git.

bertsky:
> Note: [failure of normal (Circle) CI](https://app.circleci.com/pipelines/github/bertsky/ocrd_all/12/workflows/fc97e34c-7c2d-45f4-84bc-acea4e53f018/jobs/33) seems to be an independent, very recent [problem coming from nvidia-tensorflow](https://github.com/NVIDIA/tensorflow/issues/85).


> I'll now [try to run](https://github.com/bertsky/ocrd_all/actions/runs/4548725892) the opposite of the spectrum – maximum-cuda-git.

As I suspected: the nvidia-tensorflow is now spoiling all our builds.

bertsky:
Ok, so maximum-cuda-git seems impossible to build [on Github Actions](https://github.com/bertsky/ocrd_all/actions/runs/4553596275/jobs/8030348676):
```
no space left on device
```
That's despite our efforts to first wipe the VM clean of stuff we don't need (freeing 25 GB).

What now?

bertsky:
Building locally results in an image of 36 GB size. We should try to find the minimal set of CUDA runtimes we actually need for ocrd/core-cuda.

bertsky:
> Building locally results in an image of 36 GB size. We should try to find the minimal set of CUDA runtimes we actually need for ocrd/core-cuda.

Here's my analysis:

- concerning ocrd/core:
  - wrong base image; nvidia/cuda/*-runtime-cudnn-ubuntu* – the `cudnn` contains gigabytes of cublas and cudnn that are never actually used in our venvs, because pip needs to install newer/different versions anyway, we still miss `devel` for things like nvcc (and cudnn-devel actually means development files for cudnn, not for CUDA code)
  - multi-version CUDA runtimes; we probably don't need that if we get a correct CUDA toolkit (`devel`) and rebuild packages
  - apt-get autoremove is ill-conceived; first, it also removes packages indiscriminately that we do need like most of the things in `devel` (nvcc etc); second, due to layering it does not actually reduce the size; in the non-CUDA build, keeping the extra gcc costs merely ~100 MB
- concerning ocrd/all:
  - git branches, e.g. ~700 MB `gh-pages` in ocrd_detectron2; as soon as we started shipping complete repos (`-git` variants with `pip install -e`), it was wrong to allow `git submodule update --init`; instead, we should have created partial clones. Fortunately, `git>=2.26` allows using `git submodule update --init --single-branch`. Unfortunately, Ubuntu 20.04 still only ships v2.25 (and I don't know how to simply emulate this behaviour, short of doing explicit `git clone --single-branch` for each submodule)
  - `git submodule deinit` and `git clean` in the submodules within the Docker build; this has to happen outside, in the Docker build context, via proper module dependencies of the `docker%` target
  - [abandoning our single-layer principle](https://github.com/OCR-D/ocrd_all/commit/7a5ff455f2f8d0f49ea52a835d548244a162ed88), e.g. `apt-get -y install automake autoconf libtool pkg-config g++ && make deps-ubuntu && apt-get -y autoremove && apt-get clean` as an extra step; everything must be part of `docker.sh`, otherwise it does not save space at all
  - torch both in venv and sub-venv; we actually started this to accomodate ocrd_detectron2's more elaborate (CUDA-enabled) installation recipe with ocrd-typegroups-classifier (which installs whatever torch it can get); pip being dumb as it  is breaks this satisfiable conflict; but the version pulled from ocrd_detectron2 is always better, so it should be given preference – within the top-level venv

Perhaps there's more, but that should already yield a significant decrease in image size. Fighting for CUDA support in the various processors has just begun (again)...

bertsky:
> > * multi-version CUDA runtimes; we probably don't need that if we get a correct CUDA toolkit (`devel`) and rebuild packages
> 
> OK, you already implemented those in [OCR-D/core#1041](https://github.com/OCR-D/core/pull/1041) AFAICT

Yes, but it now looks like it's even more complicated. For TF with GPU support, you do need libcudnn8 from the OS. (Unlike Torch which uses a pip package nvidia-cudnn-cu11, so there will always be two copies of that library and libcublas and others, worth around 1 GB.)

We could either install this as a `FIXUP` in core-cuda, or via `deps-ubuntu` in ocrd_all. In ocrd_all we need some extra workaround anyway: TF now depends on CUDA>=11.8, but we wanted to keep 11.3 (for various reasons). So we need to hold it at `tensorflow<2.12`, which with stupid pip means _preinstalling_ it...

> > * Unfortunately, Ubuntu 20.04 still only ships v2.25 (and I don't know how to simply emulate this behaviour, short of doing explicit `git clone --single-branch` for each submodule)
> 
> Install a newer git from https://launchpad.net/~git-core/+archive/ubuntu/ppa?

I thought of that, but that in turn would require `software-properties-common` etc. I don't like dragging in hundreds of megabytes  worth of extras without being able to remove them afterwards (because we cannot just do `autoremove`, see above).

I now believe we already get close enough by using `--depth 1`. So for the outside (build context, everything in Dockerfile before `COPY`), we can use a newer Ubuntu and pass `GIT_DEPTH=--single-branch`. And for the inside (commands echoed into docker.sh) we can use `GIT_DEPTH='--depth 1'`. I hope...

> OK, so we would clean up the git repos before the `docker build` call?

Yes, one could do e.g. 

    git submodule foreach 'for ref in $(git for-each-ref --no-contains=HEAD --format="%(refname)" refs/remotes/ | sed s,^refs/remotes/,,); do git branch -d -r $ref; done' && git gc

But for the CI build, we don't need that – as long as we never initially clone more than needed anyway (hence `--single-branch` or `--depth 1`).

> OK, so we revert [7a5ff45](https://github.com/OCR-D/ocrd_all/commit/7a5ff455f2f8d0f49ea52a835d548244a162ed88) and replace `apt-get ...` with `echo "apt-get ..." >> docker.sh`?
> 

Something like that, yes. (But we cannot use `autoremove`, too.)

> If it's really just about ocrd_detectron2 and ocrd_typegroups_classifier, can't we align their torch requirement to always install the same version?

oh, sure we can. I just added an order-only dependency between the two.

bertsky:
So, regarding all the CUDA woes, duplicate libraries, and Torch vs TF dependency conflicts, see [this comprehensive explanation](https://github.com/OCR-D/core/pull/1041#issuecomment-1573439760). There's a subissue specific to TF1, but that has been addressed via 5b1ce3e (for now). Also, since pip is so dumb and we have to install modules one by one, and they can ignore previous dependencies, I had to add fc35815 to enforce/reinstate our preinstallation compromise on cuDNN versions.

Regarding recursive cloning with single branches: on the CI, I did make use of `--single-branch` option in c06998d. But with 2d7b4c7 I decided to directly address the root cause by not updating/cloning the git modules within the Docker build anymore. So the (outer) Docker build context is the only relevant state now. Also, with 15e8d77 you can now do `make tidy` _before_ `make docker...` to avoid bloating your locally built images with unnecessary branches.

The various module updates are required for several reasons, but mostly bugfixing, too.

Makefile
```diff
@@ -123,6 +125,7 @@ Targets:
 	clean: remove the virtual environment directory, and make clean-*
 	clean-tesseract: remove the build directory for tesseract
 	clean-olena: remove the build directory for ocrd_olena
+	tidy: clean, then deinit opencv-python and git-clean all submodules
```
kba, line 128: Maybe we should add a `WARNING: potential data loss` note or similar, since `git clean` will remove data. Doesn't really concern most users - and is actually a one-command way to ensure a clean environment for debugging - but better safe than sorry.

Makefile
```diff
@@ -164,20 +169,30 @@ $(OCRD_MODULES): always-update
...
+# if you already have a clone with too many refs, consider the following recipe:
+#git submodule foreach 'for ref in $(git for-each-ref --no-contains=HEAD --format="%(refname)" refs/remotes/ | sed s,^refs/remotes/,,); do git branch -d -r $ref; done'
+	git gc
+
+
 # Get Python modules.
 
 $(BIN)/pip: $(ACTIVATE_VENV)
 	. $(ACTIVATE_VENV) && $(SEMPIP) pip install --upgrade pip setuptools
+	hash -r
```
kba, line 190: Isn't this a no-op since every line is executed by `$SHELL` separately?

Makefile
```diff
@@ -691,6 +706,30 @@ check: $(OCRD_EXECUTABLES:%=%-check) $(OCRD_MODULES:%=%-check)
...
+
+define tool-jsons-code =
+import json
+import sys
+all = dict()
+for path in sys.argv[1:]:
+    all.update(json.load(open(path))['tools'])
+print(json.dumps(all, indent=2))
+endef
+tool-jsons-file != mktemp -u
```
kba, line 727: Been using `make` for years and never knew `!=` could replace `:= $(shell ...)`.

kba (approved):
LGTM, pending verification that OCR-D/core#1041 works, ready to ship IMHO.

Makefile
bertsky: Yes, good idea.

bertsky:
now includes #365 and depends on https://github.com/OCR-D/core/pull/1055

bertsky:
CI now also [runs successfully](https://github.com/bertsky/ocrd_all/actions/runs/5261994836).

Please merge and release!

Makefile
```diff
@@ -470,37 +507,41 @@ OCRD_CIS += $(BIN)/ocrd-cis-ocropy-segment
 #OCRD_CIS += $(BIN)/ocrd-cis-ocropy-train
 OCRD_CIS += $(BIN)/ocrd-cis-postcorrect
 $(call multirule,$(OCRD_CIS)): ocrd_cis $(BIN)/ocrd
+	@# workaround against breaking changes in Numpy and OpenCV
+	. $(ACTIVATE_VENV) && $(SEMPIP) pip install "numpy<1.24" "opencv-python-headless<4.5"
```
stweil, line 511: @bertsky, could you please document which breaking changes required the old versions of numpy and opencv-python-headless? Those old versions don't work with Python 3.11. So in the long run it will be necessary to work with recent package versions.

bertsky: Sorry, I don't remember. At the time we had multiple modules which had not migrated to the new APIs of these modules, but I also remember hammering these into quite a few modules before the PR was finished.

Since we now have a `test-workflow` backing the deployment, which does cover lots of critical modules, and the quiver diachronic view as a complementary check  that can also be run locally, in advance, I actually recommend trying out dropping this – in a new PR.

PR #372, opened 2023-06-16
Remove old code for Python < 3.7

stweil:


Makefile
```diff
@@ -70,14 +70,12 @@ CUSTOM_DEPS = unzip wget parallel git less # add more packages for deps-ubuntu b
 
 DEFAULT_DISABLED_MODULES = cor-asv-fst opencv-python ocrd_ocropy ocrd_pc_segmentation
 ifeq ($(filter docker-%,$(MAKECMDGOALS)),)
-ifneq ($(PYTHON_VERSION),3.6)
```
bertsky, line 73: Why remove 3.6 here? AFAIK cor-asv-ann and ocrd_keralsm do still work with that.

stweil: Do you expect that the Makefile will still be used with Python 3.6? As far as I know we dropped support for that, so any related lines are no longer useful.

bertsky: Yes, I expect there to be users who have not migrated their installation yet. And not supporting is not the same as actively breaking things.

stweil: Hm. I would not call it "actively breaking things" if two modules are disabled by default when someone gets the latest ocrd_all and uses it with an unsupported Python version.

I don't think that users who have not migrated their installation yet will update ocrd_all unless they plan to migrate their installation.

@kba, what do you think?

bertsky: > I would not call it "actively breaking things" if two modules are disabled by default when someone gets the latest ocrd_all and uses it with an unsupported Python version.

Users mostly don't know about `DISABLED_MODULES`, let alone `DEFAULT_DISABLED_MODULES`. Your change will break these two modules in the sense that they will not be available anymore – for no reason. Again, not supporting is different from breaking a version of Python.

> I don't think that users who have not migrated their installation yet will update ocrd_all unless they plan to migrate their installation.

IMO this kind of assumption about our users' intentions is unjustified and unnecessary.

bertsky: They can still try using `--ignore-requires-python` or selectively updating only some of the submodules.

bertsky: Ok, I surrender.

bertsky (approved):

PR #379, opened 2023-06-20
docker: build ocrd-all-tool.json

bertsky:
This will provide /build/ocrd-all-tool.json as part of every prebuilt Docker image (so it won't have to be downloaded).

Still unclear to me exactly how we are going to integrate this in core (to save time looking for processor executables and dump their jsons, or for the Processing Server). But let's start right here.


bertsky:
Perhaps we should also do a `--dump-module-dir` for each executable and store a dict from executable to module location somewhere?

kba (approved):
LGTM, I will work on the core resmgr implementation asap.

PR #383, opened 2023-07-01
Repair broken GitHub action and extend it to support Python 3.11

stweil:


stweil:
These fixes / workarounds were necessary:

- The GitHub action was syntactically broken and had to be fixed.
- `make deps` tried to install Python although that was already provided by the GitHub action.
- Builds with Python 3.8 aborted with "disk full" (maybe caused by additional downloads for Tensorflow). Therefore all Docker images (which are unused in this GitHub action) are now deleted to free disk space.
- Builds with Python 3.10 and 3.11 cannot build an old version of Numpy which was used as a workaround in the build process, so that workaround was removed for those Python versions.

Now all combinations of operating system and Python version [can be built](https://github.com/stweil/ocrd_all/actions/runs/5431220207) again.

Makefile
```diff
@@ -878,13 +882,15 @@ clean-tesseract:
 # install git and parallel first (which is required for the module updates)
 deps-ubuntu:
 	apt-get -y install git parallel
+ifneq ($(PYTHON),python)
```
bertsky, line 885: It's a good solution, but please document it in the `help` target (under variables) and README, too.

Makefile
```diff
@@ -515,8 +515,12 @@ OCRD_CIS += $(BIN)/ocrd-cis-ocropy-segment
 #OCRD_CIS += $(BIN)/ocrd-cis-ocropy-train
 OCRD_CIS += $(BIN)/ocrd-cis-postcorrect
 $(call multirule,$(OCRD_CIS)): ocrd_cis $(BIN)/ocrd
+ifneq ($(PYTHON_VERSION),3.10)
+ifneq ($(PYTHON_VERSION),3.11)
 	@# workaround against breaking changes in Numpy and OpenCV
 	. $(ACTIVATE_VENV) && $(SEMPIP) pip install "numpy<1.24" "opencv-python-headless<4.5"
+endif
+endif
 	$(pip_install)
```
bertsky, line 524: Actually, these workarounds are not needed anymore – even for Python 3.8:

```suggestion
	$(pip_install)
```

.github/workflows/makeall-linux.yaml
```diff
@@ -42,12 +42,17 @@ jobs:
     - name: Show inputs.os
       run: echo ${{ inputs.os }}
     - name: Show environment
-      run: echo OS=${{ env.OS }}, PYTHON_VERSION=${{ env.PYTHON_VERSION }}
+      run: echo "OS=${{ env.OS }}, PYTHON_VERSION=${{ env.PYTHON_VERSION }} ($(python --version))"
+    - name: Remove Docker images
+      run: |
+           df -h
+           docker rmi $(docker images -q)
+           df -h
```
bertsky, line 50: We could also add the [other recipes from makedocker](https://github.com/OCR-D/ocrd_all/blob/f6dca7db1a3c0585d8452b91eda3226c14c4d150/.github/workflows/makedocker.yml#L59-L79) here to free even more space.

bertsky (commented):
LGTM.

Could you please also remove the `-git` variants from the [Docker build workflow](https://github.com/OCR-D/ocrd_all/blob/f6dca7db1a3c0585d8452b91eda3226c14c4d150/.github/workflows/makedocker.yml#L15-L27), which are redundant now?

.github/workflows/makeall-linux.yaml
stweil: Thanks, that is currently not necessary (and would cost additional build time), but can be done when needed.

stweil:
> LGTM.
> 
> Could you please also remove the `-git` variants from the [Docker build workflow](https://github.com/OCR-D/ocrd_all/blob/f6dca7db1a3c0585d8452b91eda3226c14c4d150/.github/workflows/makedocker.yml#L15-L27), which are redundant now?

Sure, that's now implemented in pull request #384.

Makefile
stweil: Done in commit 204e9cad9bf527195a090bd671703959cbb34b06.

README.md
```diff
@@ -247,6 +247,8 @@ prior to a commit.)
 #### _PYTHON_
 
 Name of the Python binary to use (at least python3 required).
+It is also used for the target `deps-ubuntu` to select the required packages
+unless it was set to `python`.
```
bertsky, line 251:
```suggestion

If set to just `python`, then for the target `deps-ubuntu` it is assumed that Python is already installed.
```

bertsky (approved):
Thanks!

PR #386, opened 2023-07-07
Draft for slim containers

joschrew:
Description:
All processors run in their own docker-container as a processing-worker. Also there are containers: processing-server, rabbitmq and mongodb running. Executables for processors are delegators to `ocrd network client`.

Related issue:
https://github.com/OCR-D/ocrd_all/issues/69#issuecomment-1600309509

Example usage:

Clone this repo:
`git clone git@github.com:joschrew/ocrd_all.git`

Change to repository
`cd ocrd_all`

Core, ocrd_cis and ocrd_tesserocr are needed for the example-run:
`git submodule update --init core/ ocrd_cis/ ocrd_tesserocr/`

Create the venv and docker-compose.yaml:
`make -f Makefile-slim slim-venv`

Create datadir (necessary to get the workspaces to the containers):
`mkdir data`

Start the containers:
`docker-compose up -d`

Get a workspace for testing:
```
curl "https://raw.githubusercontent.com/OCR-D/ocrd-webapi-implementation/main/things/example_ws.ocrd.zip" --output foo.zip
unzip foo.zip "data/*"
```

Activate the venv:
`. venv2/bin/activate`

Run a processor on the workspace:
`ocrd-cis-ocropy-binarize -I OCR-D-IMG -O OCR-D-BIN -m mets.xml`


Makefile-slim
```diff
@@ -0,0 +1,41 @@
...
+
+docker-compose.yaml:
+	@cat slim-containers-files/docker-compose.template.yaml > docker-compose.yaml
+	@$(foreach mod,$(OCRD_MODULES),$(foreach proc,$($(mod)),$(call add_proc,$(proc),$(mod))))
+
+.env:
+	@rm -rf .env
+	@echo OCRD_PS_PORT=8000 >> .env
+	@echo OCRD_PS_MTU=1300 >> .env
+	@echo MONGODB_URL=mongodb://ocrd-mongodb:27017 >> .env
```
bertsky, line 39: There should be some login credentials in here, right?
```suggestion
	@echo MONGODB_USER=admin >> .env
	@echo MONGODB_PASS=admin >> .env
	@echo 'MONGODB_URL=mongodb://$${MONGODB_USER}:$${MONGODB_PASS}@ocrd-mongodb:27017' >> .env
```

Makefile-slim
```diff
@@ -0,0 +1,41 @@
...
+	@echo RABBITMQ_URL=amqp://admin:admin@ocrd-rabbitmq:5672 >> .env
```
bertsky, line 40: Must be reusable for clients **and** Compose, hence factor out credentials as indepenent variable:
```suggestion
	@echo RABBITMQ_USER=admin >> .env
	@echo RABBITMQ_PASS=admin >> .env
	@echo 'RABBITMQ_URL=amqp://$${RABBITMQ_USER}:$${RABBITMQ_PASS}@ocrd-rabbitmq:5672' >> .env
```

slim-containers-files/ps-config.yaml
```diff
@@ -0,0 +1,12 @@
+process_queue:
+  address: ocrd-rabbitmq
+  port: 5672
+  skip_deployment: true
+  credentials:
+    username: admin
+    password: admin
```
bertsky, line 7: IIUC these will be used by the PS even if the deployment is done by Compose. So we need a mechanism to align these settings with the ones in .env, right? (So either we parse the config file when generating .env, or we write to the config file on startup...)

slim-containers-files/ps-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+database:
+  address: ocrd-mongodb
+  port: 27017
+  skip_deployment: true
```
bertsky, line 11: Same here – there should be a `credentials` key with `username` and `password` – either parsed by make or generated dynamically.

slim-containers-files/Dummy-Core-Dockerfile
```diff
@@ -0,0 +1,13 @@
+# I need this because i need the network-for-slim-branch and this contains a comment how to make
+# pudb run.
+FROM ocrd/core:latest AS base
+WORKDIR /build-ocrd
+RUN apt install vim-tiny -y
+RUN git clone https://github.com/ocr-d/core.git && \
+	cd core && \
+	git checkout network-for-slim-prep && \
+	#sed -i "290 i \        from pudb.remote import set_trace; set_trace(term_size=(160, 40), host='0.0.0.0', port=6900)" ocrd_network/ocrd_network/processing_server.py && \
```
kba, line 9: What exactly does this do, ie. what and how do you debug with `pudb`?

Makefile-slim
```diff
@@ -0,0 +1,41 @@
...
+BIN = $(VIRTUAL_ENV)/bin
+ACTIVATE_VENV = $(BIN)/activate
+OCRD_MODULES = OCRD_CIS OCRD_TESSEROCR
+OCRD_CIS = ocrd-cis-ocropy-binarize ocrd-cis-ocropy-dewarp
+OCRD_TESSEROCR = ocrd-tesserocr-recognize ocrd-tesserocr-segment-region
+PROCESSORS = $(foreach mod,$(OCRD_MODULES),$(foreach proc,$($(mod)), $(proc) ))
+DELEGATORS = $(foreach proc,$(PROCESSORS),$(BIN)/$(proc))
+
+slim-venv: docker-compose.yaml .env $(DELEGATORS) | $(VIRTUAL_ENV)
+
```
bertsky, line 12: None of this would be needed if you added to the existing `Makefile` directly. We already have (sensible+configurable) definitions for `VIRTUAL_ENV`, `OCRD_MODULES` (not a variable but the true submodule name) and `OCRD_EXECUTABLES` (your `PROCESSORS`). There is even an existing delegator mechanism (used for sub-venvs on some modules).

slim-containers-files/delegator_template.py
```diff
@@ -0,0 +1,24 @@
...
+import sys
+from pathlib import Path
+import subprocess
+
+# Later the address (or rather the port) should be dynamic
+processing_server_address = "http://localhost:8000"
+processor_name = "{{ processor_name }}"
+
+args = list(sys.argv)
+if "-m" in args:
```
kba, line 11: This does not handle the full complexity of `--mets/--working-dir` though. What is the purpose here, prepending `/data` to the METS path?

Makefile-slim
```diff
@@ -0,0 +1,41 @@
...
+$(VIRTUAL_ENV): $(ACTIVATE_VENV)
+	. $(ACTIVATE_VENV) && $(MAKE) -C core install
+
+%/bin/activate:
+	$(PYTHON) -m venv $(subst /bin/activate,,$@)
+	. $@ && pip install --upgrade pip setuptools wheel
+
+# append the service to docker-compose for a processor
+add_proc = sed -e "s/{{\s*processor_name\s*}}/$1/" -e "s/{{\s*processor_group_name\s*}}/\L$2/" \
+            slim-containers-files/docker-compose.processor.template.yaml >> docker-compose.yaml;
```
bertsky, line 29: We can use `define` + `endef` directive and write everything as a pure make rule here – using `subst` instead of `sed` etc. Also, IMO a unique identifier like `PROCESSOR_NAME` and `MODULE_NAME` would be better than the double-brace notation.

bertsky (commented):
Early stage, dry review – still trying to make it run for me.

slim-containers-files/ocrd_tesserocr/Dockerfile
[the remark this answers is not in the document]
kba: I think it's just for proof-of-concept, so @joschrew does not need to keep multiple PR in sync.

Makefile-slim
```diff
@@ -0,0 +1,41 @@
...
+
+docker-compose.yaml:
+	@cat slim-containers-files/docker-compose.template.yaml > docker-compose.yaml
+	@$(foreach mod,$(OCRD_MODULES),$(foreach proc,$($(mod)),$(call add_proc,$(proc),$(mod))))
+
+.env:
+	@rm -rf .env
```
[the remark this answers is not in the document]
kba: Unless you make it `.PHONY`. But I'm confused by `.env` anyway, why is this generated from the Makefile, why not just a static file that can be adapted with different ports etc. as needed?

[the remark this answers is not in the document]
bertsky: Because if it is controlled in git you cannot simply adapt it (or then have to be careful with git actions).

I do think the make-generated .env is a good mechanism.

joschrew:
I have started the process of addressing all comments but this will take a while. I think I'll add them step by step. I will report back if that is finished.

slim-containers-files/Dummy-Core-Dockerfile
```diff
@@ -0,0 +1,13 @@
+# I need this because i need the network-for-slim-branch and this contains a comment how to make
+# pudb run.
+FROM ocrd/core:latest AS base
+WORKDIR /build-ocrd
+RUN apt install vim-tiny -y
```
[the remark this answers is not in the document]
joschrew: Removed the Dummy-Dockerfile completely and using core/Dockerfile now. Set core-submodule to my branch "[network-for-slim-prep](https://github.com/OCR-D/core/tree/network-for-slim-prep)" to have needed code changes.

slim-containers-files/Dummy-Core-Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN git clone https://github.com/ocr-d/core.git && \
+	cd core && \
+	git checkout network-for-slim-prep && \
```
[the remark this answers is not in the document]
joschrew: see https://github.com/OCR-D/ocrd_all/pull/386#discussion_r1261331573

slim-containers-files/Dummy-Core-Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+	#sed -i "290 i \        from pudb.remote import set_trace; set_trace(term_size=(160, 40), host='0.0.0.0', port=6900)" ocrd_network/ocrd_network/processing_server.py && \
+	make install-dev && \
+	pip install pudb
+EXPOSE 6900
```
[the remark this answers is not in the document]
joschrew: see https://github.com/OCR-D/ocrd_all/pull/386#discussion_r1261331573

slim-containers-files/delegator_template.py
```diff
@@ -0,0 +1,24 @@
...
+        print("absolute path is not supported")
+        exit(1)
+    args[idx + 1] = f"/data/{metspath}"
+
+
+cmd = [
+    "ocrd", "network", "client", "processing", "processor",
+    processor_name, "--address", processing_server_address
+]
+subprocess.run(cmd + args[1:])
```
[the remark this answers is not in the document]
bertsky: It could be as simple as passing `"--callback-url", "http://localhost:" + PORT + "/"` here. The `PORT` should be generated before by an internal background task like so:
```python
PORT = None
async def serve_callback():
    nonlocal PORT
    # set up web server
    def get(path):
        raise Exception("done")
    ...
    # run web server on arbitrary available port
    PORT = ...

server = asyncio.create_task(serve_callback)
```
Then after the `subprocess.run`, you can simply block by doing `await server`. The whole thing must be in a function that you `asyncio.run(...)`.

slim-containers-files/delegator_template.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+import sys
+from pathlib import Path
+import subprocess
+
+# Later the address (or rather the port) should be dynamic
+processing_server_address = "http://localhost:8000"
```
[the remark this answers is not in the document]
joschrew: After trying some things I think I cannot solve this properly. First the hostname cannot be the service name because the delegator is run from the host. And I cannot read the port from .env because I don't know the working dir where the delegator is executed from. So I decided to set the port dynamic in the Makefile like the processor-name. The proposed suggestion at least does not work. Commit: https://github.com/OCR-D/ocrd_all/pull/386/commits/7736969458dc24281ff87d189167d426209a02be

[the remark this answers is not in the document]
bertsky: > First the hostname cannot be the service name because the delegator is run from the host.

Oh, right! Sorry, weak thinking on my side.

> And I cannot read the port from .env because I don't know the working dir where the delegator is executed from.

Ok, good point. However, we could write all the .env values into the venv's `activate` by make, **dynamically** (i.e. setting the exact values used for the servers):
```Makefile
%/bin/activate:
	$(PYTHON) -m venv $(subst /bin/activate,,$@)
	. $@ && pip install --upgrade pip setuptools wheel
	@echo ". $(CURDIR)/.env" >> $@
```

IMO the .env should be the central source for configuration, so you should be able to modify it to your needs after it was generated.

slim-containers-files/delegator_template.py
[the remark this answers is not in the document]
joschrew: [I have added a waiting mechanism with the python build-in `HTTPServer`](https://github.com/OCR-D/ocrd_all/pull/386/commits/e86611765afc7828a7ff0d44aa4a6c266e7b0d79). I tried using your proposed async approach but discarded that before I could completely finish. Probably async is possible but it is way more complicated (that's why I finally opted out) and I think a "non-async" http server has the same effect and the waiting-costs for running a "normal" http server vs. a async http server is neglectable. 
I cannot say I have much more than an idea of how it internally works but I think a python-process with an async server gets scheduled by the operating system as well anyway. The async-stuff with polling and so on is "only" python internal. It would gain time if we had another async process running alongside which could then be run when the server is waiting.
So I don't think async is helpful here anyway, that's why I use the synchrony server.

slim-containers-files/docker-compose.processor.template.yaml
```diff
@@ -0,0 +1,14 @@
...
+  {{ processor_name }}:
+    extends:
+      file: slim-containers-files/{{ processor_group_name}}/docker-compose.yaml
+      service: {{ processor_name }}
+    command: ocrd network processing-worker --database $MONGODB_URL --queue $RABBITMQ_URL --create-queue {{ processor_name }}
+    depends_on:
+      - ocrd-processing-server
+      - ocrd-mongodb
+      - ocrd-rabbitmq
+    # restart: The worker creates its queue but rabbitmq needs a few seconds to be available
```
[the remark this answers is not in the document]
joschrew: The underlining problem is the following: The container for the queue is started and running, but it needs 1-3 seconds that queue creation is possible. But the processing worker tries to create it's queue right away.

This suggestion (`service_started`) is not working because the queue-service is considered started from docker-compose but it is in reality not ready to be used right away although it is considered started. I have a similar issue in another project and already tried a few things solving this. I think the only solution to this problem from the docker side would be to implement a (manual) health-check for the rabbitmq container. But therefore I'd have to extend the rabbit-mq image which I do not want.

For this PR to function some extension to core is needed anyway. There I want to add an optional queue-creation-timeout to the worker startup so that it waits a few seconds with adding its queue or to try again a few times. But this restart-fix was the fastest way to do that that's why it is here and I agree that it should be removed finally. I will remark this as solved as soon as the needed changes to core are made (which need one change to this PR as well).

slim-containers-files/Dummy-Core-Dockerfile
joschrew: I used that for debugging with pudb. Therefore I installed pudb in the container and inserted a breakpoint (with sed in line 290). Pudb is able to run as a server I used that to connect to the debugger from the host.
This should of course not be in this PR but I kept it at first to be able to use it again if needed. It is deleted already

slim-containers-files/delegator_template.py
joschrew: I think I didn't fully take into account what --mets and --working-dir actually are doing. My first idea was that the user can use the slim containers exactly like an ocrd_all venv. Regarding the workspaces this will not work because they have to be available to the docker containers. For example arbitrary absolute path can be provided to the venv via --mets. This of course will not work with docker. But I didn't think this through obviously.

However, I tried to somehow work around this but as bertsky already addressed this does not make much sense so I already removed that.

Makefile-slim
joschrew: I tried a lot of stuff regarding this and I kind of got this to work but I think my solution only made it worse (using file function and putting the template in a variable and so on). So I cannot solve this. And I prefer the double brace notation to putting the variables in upper case, otherwise it is harder to distinguish for me between env-variables and things the makefile replaces.

slim-containers-files/ps-config.yaml
joschrew: [processing server config now defined in docker-compose and it reads from .env] (https://github.com/OCR-D/ocrd_all/pull/386/commits/c6e67d6a9c1844a398c57b43745c73e7353391bf)

slim-containers-files/ps-config.yaml
joschrew: [processing server config now defined in docker-compose and it reads from .env](https://github.com/OCR-D/ocrd_all/pull/386/commits/c6e67d6a9c1844a398c57b43745c73e7353391bf)

slim-containers-files/docker-compose.template.yaml
```diff
@@ -0,0 +1,34 @@
...
+  ocrd-processing-server:
+    build:
+      # later real ocrd-core image should be referenced here
+      dockerfile: slim-containers-files/Dummy-Core-Dockerfile
+      args:
+        BASE_IMAGE: ubuntu:20.04
+    ports:
+      - ${OCRD_PS_PORT}:8000
+    volumes:
+      - "./slim-containers-files/ps-config.yaml:/ocrd-processing-server-config.yaml"
```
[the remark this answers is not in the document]
joschrew: As a side effect of [this](https://github.com/OCR-D/ocrd_all/pull/386/commits/c6e67d6a9c1844a398c57b43745c73e7353391bf) this is resolved, no longer using volume mount

slim-containers-files/delegator_template.py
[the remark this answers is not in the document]
joschrew: I see a problem with that approach, because it only works with the venv activated in bash (not csh or fish) and invoking executables directly would not work with this approach  (e.g.: `venv/bin/ocrd-some-processor ...`).
But I think it is a good idea to use the .env for the config so I have a [slightly modified approach](https://github.com/OCR-D/ocrd_all/pull/386/commits/cfd1a79f1ff47252dea8fe23956470a8a3182c0b): the path to the .env is put into the delegator-template when generated in the makefile. I think this solution is better then sourcing .env in the venv-activator-script (btw in the example code this does not work: `@echo ". $(CURDIR)/.env" >> $@`, exporting is needed to be available to a started python process e.g. the processor).

slim-containers-files/delegator_template.py
```diff
@@ -3,8 +3,17 @@
 import subprocess
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from os.path import basename
+import re
+from pathlib import Path
 
-processing_server_address = "http://localhost:{{ OCRD_PS_PORT }}"
+env_path = "{{ OCRD_SLIM_ENV_PATH }}"
```
kba, line 9: Maybe use https://github.com/theskumar/python-dotenv for that, so you don't have to parse `sh` variable assignments?

slim-containers-files/delegator_template.py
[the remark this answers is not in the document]
joschrew: Have reverted the last commit regarding this and decided to use the [proposed approach](https://github.com/OCR-D/ocrd_all/pull/386/commits/a7c1328d93d26ad0fdda5a834b051f8be400363c)

slim-containers-files/ocrd_tesserocr/docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
+services:
+  ocrd-tesserocr-recognize:
+    build:
+      context: ../../ocrd_tesserocr
+      dockerfile: ../slim-containers-files/ocrd_tesserocr/Dockerfile
+    command:
+      ocrd network processing-worker ocrd-tesseroc-recognize --database $MONGODB_URL --queue $RABBITMQ_URL --create-queue
```
kba, line 7:
```suggestion
      ocrd network processing-worker ocrd-tesserocr-recognize --database $MONGODB_URL --queue $RABBITMQ_URL --create-queue
```

slim-containers-files/ocrd_tesserocr/docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+  ocrd-tesserocr-segment-region:
+    build:
+      context: ../../ocrd_tesserocr
+      dockerfile: ../slim-containers-files/ocrd_tesserocr/Dockerfile
+    command:
+      ocrd network processing-worker ocrd-tesserocr-segment-region --database $MONGODB_URL --queue $RABBITMQ_URL --create-queue
```
kba, line 14: We should now use the `ocrd-tesserocr-segment-region worker` syntax instead to get instance caching, right?

bertsky:
superseded by #449

PR #387, opened 2023-09-15
Build ocrd_kraken also for latest Python versions

stweil:
Python 3.10 and 3.11 now are supported by current kraken.


stweil:
Please merge.

PR #395, opened 2023-10-18
update 2023-10-18

kba:
Small hotfix release for core, we've fixed a bug breaking the ocrd_tesserocr worker logging and another bug for bashlib processors and updated the logging configuration (that is the default in the docker containers) to log everything `DEBUG` and up and to attach a fileHandler to all loggers (no more empty `ocrd.log`).

Since OCR-D/core#1118 we have tagged Dockerhub releases again. To make the build consistent, instead of relying on `latest`, the docker containers are now built based on the tag derived from the git tag of core. This also ensures that if something went wrong with building the most recent core build, we'll notice it right away instead of silently (and likely inadvertently) using an older core base image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Changelog
 
+## Unreleased
+
 ## [v2023-10-18](https://github.com/OCR-D/ocrd_all/releases/v2023-10-18)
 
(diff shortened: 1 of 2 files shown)


stweil:
Maybe you want to fix the commit summary. s/asisgnment/assignment/

kba:
> Maybe you want to fix the commit summary. s/asisgnment/assignment/

I'll force-push that commit out anyway, once https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/981/workflows/0c20fe72-da04-4846-b329-3adf8e149729/jobs/1324 finishes.

PR #399, opened 2023-10-23
Update ocrd_neat

stweil:


kba:
Thanks, was included in v2023-12-05.

PR #410, opened 2024-02-23
Support Python 3.12

stweil:


stweil:
If desired I can also add CI support for Python 3.12.

PR #418, opened 2024-03-13
dockerignore tesseract build dir

kba:
should be regenerated from `./configure' within docker build context

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,6 +16,7 @@ opencv-python
 # recursive submodules, too deep
 ocrd_tesserocr/repo/tesseract/test/*
 ocrd_tesserocr/repo/tesseract/unittest/third_party/googletest/*
+ocrd_tesserocr/build_tesseract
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -16,6 +16,7 @@ opencv-python
 # recursive submodules, too deep
 ocrd_tesserocr/repo/tesseract/test/*
 ocrd_tesserocr/repo/tesseract/unittest/third_party/googletest/*
+ocrd_tesserocr/build_tesseract
```
stweil, line 19: Wouldn't it be better to rename the build directory in ocrd_tesserocr, either to ocrd_tesserocr/build or to ocrd_tesserocr/build/tesseract?

stweil: Yes, and naming build directories `build` is very common.

stweil:
This PR is obsolete with latest ocrd_tesserocr. Therefore I close it.

PR #423, opened 2024-03-15
Add GitHub action for macOS builds

stweil:


stweil:
This pull request depends on pull request #410 and currently ignored errors from `make check` because of issue https://github.com/OCR-D/ocrd_froc/issues/10.

It must be triggered manually, but could also be run for example weekly in the future to detect any regressions earlier.

stweil:
I tried two different variants for getting the submodules:

1. action/checkout can get all submodules recursively. By default it uses depth 1, but as it gets more submodules than required, it takes 189 s for the checkout step.
2. Get submodules during the build process (`make all GIT_DEPTH="--depth 1"`). That costs about 45 s during  the build step, but reduces the time for the checkout step to 1 s.

So the. 2nd variant is much better, at least with `GIT_DEPTH="--depth 1"` which should be used by all CI and Docker builds. Currently those builds don't use it.

stweil:
Example result for this action: https://github.com/stweil/ocrd_all/actions/runs/8294000625/job/22698144574.

.github/workflows/makeall-macos.yml
```diff
@@ -0,0 +1,78 @@
...
+      run: ls -l /usr/bin/python* /usr/local/bin/p*
+    - uses: actions/checkout@v4
+    - name: Install dependencies
+      # curl and icu4c are already installed
+      run: brew install automake bash cabextract geos leptonica libarchive libtool pango parallel
+    - name: Make all
+      if: runner.os == 'macOS'
+      run: LC_ALL=C PKG_CONFIG_PATH=$HOMEBREW_PREFIX/opt/icu4c/lib/pkgconfig:$HOMEBREW_PREFIX/opt/libarchive/lib/pkgconfig make all GIT_DEPTH="--depth 1"
+    - name: Make check
+      run: make check || echo "make check failed"
```
bertsky, line 78:
```suggestion
      run: make check
```

.github/workflows/makeall-macos.yml
```diff
@@ -0,0 +1,78 @@
...
+          - 'macos-13'
+          - 'macos-14'
+      python-version:
+        description: 'Python version'
+        required: true
+        default: 'default'
+        type: choice
+        options:
+          - 'default'
+          - '3.7'
```
bertsky, line 36:
```suggestion
```

.github/workflows/makeall-macos.yml
stweil: Done. Note that the check will now fail until ocrd_froc was updated.

.github/workflows/makeall-macos.yml
stweil: Done, thank you for your review.

PR #429, opened 2024-04-24
Clean network logging in dockerimage

### LINKED ISSUE #426: Docker: Logfile permissions problems
When I run the ocrd_all docker-image as another user than root `initLogging` causes an error when trying to init the processing-server logfile. Because the logfile is already created during the docker image build process it is owned by root. When I start the ocrd_all container as another user, initLogging might be called and raises an error like this:
```
  File "/usr/local/bin/ocrd", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1685, in invoke
    super().invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/ocrd/cli/network.py", line 25, in network_cli
    initLogging()
  File "/usr/local/lib/python3.8/site-packages/ocrd_utils/logging.py", line 181, in initLogging
    logging.config.fileConfig(config_file)
  File "/usr/lib/python3.8/logging/config.py", line 79, in fileConfig
    handlers = _install_handlers(cp, formatters)
  File "/usr/lib/python3.8/logging/config.py", line 145, in _install_handlers
    h = klass(*args, **kwargs)
  File "/usr/lib/python3.8/logging/__init__.py", line 1147, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/lib/python3.8/logging/__init__.py", line 1176, in _open
    return open(self.baseFilename, self.mode, encoding=self.encoding)
PermissionError: [Errno 13] Permission denied: '/tmp/ocrd_processing_server.log'

joschrew:
Fixes #426

This pr removes the possibly created (while creating the docker image) log-file and folders so that when an docker container with ocrd is created no logfiles and folders exist and thus are created with permissions of the user starting the docker container.

Alternative to https://github.com/OCR-D/core/pull/1214

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -152,6 +152,9 @@ RUN sed -i '/width\|height/s/value="16KP"/value="64KP"/' /etc/ImageMagick-6/poli
 # cannot just set ENV, because internal bashrc will override it anyway
 RUN echo "PS1='\w\$ '" >> /etc/bash.bashrc
 
+# clean possibly created log-files & -folders of ocrd_network logger to prevent permission problems
+RUN rm -rf /tmp/ocrd_*
+
 # reset to interactive
 ENV DEBIAN_FRONTEND teletype
 


MehmedGIT (approved):

kba:
Is this superseded by OCR-D/core#1214?

stweil:
> Is this superseded by [OCR-D/core#1214](https://github.com/OCR-D/core/pull/1214)?

I don't think that one is a valid solution because of security concerns.

joschrew:
I don't think 774 and 664 solve the problem. Idea is the following: The log-files are created by root, e.g. when in the docker-image creation ocrd-resources are downloaded or any other ocrd command is executed which involves logging. Later the image is used by another user with different permissions. Then ocrd is used within this container, the logfiles exist and the ocrd-process with the user's permissions tries to write to them which will in this case not work with 664 for example.

The permissions are only set for the logfiles and its directories which are created by ocrd and only if they do not exist previously. As far as I see it, in the worst case the logfiles could be lost. But maybe I there are other problem with the permissions I am not aware of? What are the possible problems with these permissions?

joschrew:
> Is this superseded by [OCR-D/core#1214](https://github.com/OCR-D/core/pull/1214)?

When I thought about the issue I came up with this two (both mentioned pr's) possible solutions. I think both possibly fixes the described problem. The "problem" with this PR is, it only fixes the problem with the ocrd-all-docker-image, and not if similar things happen with other images using ocrd like ocrd-processors.

PR #431, opened 2024-04-29
Replace Tesseract model name frk by the new name deu_latf

stweil:


stweil:
This PR depends on an update of ocrd_tesserocr which includes https://github.com/OCR-D/ocrd_tesserocr/pull/205.

PR #441, opened 2024-06-07
update 2024-06-08

kba:

PR #436, opened 2024-05-24
Docker build staged

bertsky:
fixes #416 
fixes #417 
fixes #434 
fixes #427

PR #442, opened 2024-06-26
Dockerfile: remove /build/core/.git directory to overwrite with submodule link

kba:
Otherwise build fails with

```
#11 ERROR: cannot replace to directory /.../build/core/.git with file
```

I am not sure how we introduced this and this workaround is not pretty (`rm -rf /build/core/.git`) so if somebody has a better solution, I'd be happy to change it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,7 +95,7 @@ RUN rm -rf /build/core/.git
 COPY . .
 
 # verify we got the right version of core
-RUN bash -c 'CORE_VERSION=`git -C /build/core describe --tags --abbrev=0`; if [[ "$BASE_IMAGE" != *":$CORE_VERSION" ]]; then echo $BASE_IMAGE inconsistent with core version $CORE_VERSION; exit 1;fi'
+RUN bash -c 'CORE_VERSION=`git -C /build/core describe --tags`; if [[ "$BASE_IMAGE" != *":$CORE_VERSION" ]]; then echo $BASE_IMAGE inconsistent with core version $CORE_VERSION; exit 1;fi'
 
 # make apt system functional
 RUN apt-get -y update && apt-get install -y apt-utils
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,6 +89,7 @@ RUN rm $VIRTUAL_ENV/bin/pip* && apt-get purge -y python3-pip && python3 -m venv 
 # copy (sub)module directories to build
 # (ADD/COPY cannot copy a list of directories,
 #  so we must rely on .dockerignore here)
+RUN rm -rf /build/core/.git
 COPY . .
 
 # make apt system functional
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,11 +86,16 @@ WORKDIR /build
 # create virtual environment
 RUN rm $VIRTUAL_ENV/bin/pip* && apt-get purge -y python3-pip && python3 -m venv $VIRTUAL_ENV && python3 -m pip install --force pip
 
+# from-stage already contains a clone clashing with build context
+RUN rm -rf /build/core/.git
+
 # copy (sub)module directories to build
 # (ADD/COPY cannot copy a list of directories,
 #  so we must rely on .dockerignore here)
-RUN rm -rf /build/core/.git
 COPY . .
+
+# verify we got the right version of core
+RUN bash -c 'CORE_VERSION=`git -C /build/core describe --tags --abbrev=0`; if [[ "$BASE_IMAGE" != *":$CORE_VERSION" ]]; then echo $BASE_IMAGE inconsistent with core version $CORE_VERSION; exit 1;fi'
 
 # make apt system functional
 RUN apt-get -y update && apt-get install -y apt-utils
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,8 +94,9 @@ RUN rm -rf /build/core/.git
 #  so we must rely on .dockerignore here)
 COPY . .
 
+# TODO(kba)
 # verify we got the right version of core
-RUN bash -c 'CORE_VERSION=`git -C /build/core describe --tags`; if [[ "$BASE_IMAGE" != *":$CORE_VERSION" ]]; then echo $BASE_IMAGE inconsistent with core version $CORE_VERSION; exit 1;fi'
+# RUN bash -c 'CORE_VERSION=`git -C /build/core describe --tags`; if [[ "$BASE_IMAGE" != *":$CORE_VERSION" ]]; then echo $BASE_IMAGE inconsistent with core version $CORE_VERSION; exit 1;fi'
 
 # make apt system functional
 RUN apt-get -y update && apt-get install -y apt-utils
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ jobs:
   build:
     docker:
       - image: cimg/base:current-22.04
+    resource_class: large
     steps:
       - checkout
       - setup_remote_docker      # https://circleci.com/docs/2.0/building-docker-images/


Dockerfile
```diff
@@ -89,6 +89,7 @@ RUN rm $VIRTUAL_ENV/bin/pip* && apt-get purge -y python3-pip && python3 -m venv
 # copy (sub)module directories to build
 # (ADD/COPY cannot copy a list of directories,
 #  so we must rely on .dockerignore here)
+RUN rm -rf /build/core/.git
```
bertsky, line 92: How can that help at all? The submodules are copied in the follow-up line.

bertsky (changes requested):
I suppose you are trying to fix [CI of #441](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/1075/workflows/11f4fb25-cdf0-4ce9-bc2b-f5aa72ad3c50/jobs/1484?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary).

This used to work before. Something else is going on here. I don't think we should resort to workarounds just yet.

I'll try to debug via SSH...

Dockerfile
kba: The problem is that the `COPY . .` in the following line fails because the `/build/core/.git` that comes with the base image is a directory but since `core` is a git submodule in ocrd_all, `/build/core/.git` is a file pointing to the parent git dir (`gitdir: ../.git/modules/core`). And `COPY` (as well as `cp` and `mv`) will abort when trying to overwrite a directory with a non-directory.

I get the same behavior locally, so it is not just in the CD.

bertsky: Oh, right! I forgot about that (base stage already yields an independent clone).

So this also begs the question (again) what is supposed to happen if the submodule version of core deviates from the base stage version of core. For now we settled with the approach of determining core's version from the submodule and passing that as tag for the base stage in a Docker build. Perhaps we should put an assertion into the Dockerfile like so:
```Dockerfile
# from-stage already contains a clone clashing with build context
RUN rm -rf /build/core/.git
COPY . .
# verify we got the right version of core
RUN COREV=`git -C /build/core describe --tags --abbrev=0` && [[ "$BASE_IMAGE" =~ :$COREV ]]
```

bertsky: At any rate, I suggested adding a dedicated comment for the `RUN rm` step and moving it a few lines up because the existing comment still applies to `COPY` only.

kba: > So this also begs the question (again) what is supposed to happen if the submodule version of core deviates from the base stage version of core.

Considering that we derive `CORE_VERSION` from `git describe` and derive `BASE_IMAGE` from `CORE_VERSION in the Makefile, I don't see where this might actually happen. If there is no versioned Docker image matching the version in the `core` submodule (ie. the CD of core failed and we did not notice), then the build will fail at the `FROM $BASE_IMAGE` line. The only scenario I can think of is that the `core` submodule contains unreleased commits, in which case there could be a mismatch, e.g. it is not at version `vX` but at `vX-commitsha`, ie. we should not use the `--abbrev=0` option, so only a core submodule at the exact commit of a tag should be acceptable.

Hence, I've adapted your proposed check to fail the build in this case with a message like

```
ocrd/core:v2.66.1 inconsistent with core version v2.66.1-1-gd960be4e
```

Considering this is now a requisite for #441, I'll merge once CI is happy.

kba:
CI [failed as expected](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/1078/workflows/46661fe9-14c5-4eda-a549-959c7d1da24d/jobs/1489?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) with

```
ocrd/core-cuda:v2.65.0 inconsistent with core version v2.65.0-1-gcc6ea575
```

which is good, but I update core in this PR as well to make sure it won't if all is set up properly.

kba:
> CI [failed as expected](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/1078/workflows/46661fe9-14c5-4eda-a549-959c7d1da24d/jobs/1489?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) with
> 
> ```
> ocrd/core-cuda:v2.65.0 inconsistent with core version v2.65.0-1-gcc6ea575
> ```
> 
> which is good, but I update core in this PR as well to make sure it won't if all is set up properly.

Ah, but of course [there is a catch](https://app.circleci.com/pipelines/github/OCR-D/ocrd_all/1079/workflows/f0e0b372-4a3a-46a5-a368-234f098a9251/jobs/1490?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary). The check won't work if we derive from a non-versioned Docker base image such as `ocrd/core:minimum-cuda`.

I still think the check is a good idea and will open an issue for it but comment it out for now because we really need a new release urgently.

kba:
It looks like `make docker-maximum-cuda` does not just build that image:

```
> make -n docker-maximum-cuda NO_UPDATE=1
docker build \
--progress=plain \
--build-arg BASE_IMAGE=ocrd/core-cuda:v2.66.1 \
--build-arg VCS_REF=$(git rev-parse --short HEAD) \
--build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
--build-arg OCRD_MODULES="ocrd_cis ocrd_fileformat ocrd_im6convert ocrd_pagetopdf ocrd_repair_inconsistencies ocrd_tesserocr ocrd_wrap workflow-configuration ocrd_olahd_client" \
--build-arg PIP_OPTIONS="-e" \
--build-arg PARALLEL="" \
--build-arg PYTHON="python3" \
--network=host \
-t ocrd/all:minimum-cuda .
docker build \
--progress=plain \
--build-arg BASE_IMAGE=ocrd/all:minimum-cuda \
--build-arg VCS_REF=$(git rev-parse --short HEAD) \
--build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
--build-arg OCRD_MODULES="cor-asv-ann dinglehopper docstruct format-converters nmalign ocrd_calamari ocrd_cis ocrd_fileformat ocrd_im6convert ocrd_keraslm ocrd_olahd_client ocrd_olena ocrd_pagetopdf ocrd_repair_in
consistencies ocrd_segment ocrd_tesserocr ocrd_wrap workflow-configuration" \
--build-arg PIP_OPTIONS="-e" \
--build-arg PARALLEL="" \
--build-arg PYTHON="python3" \
--network=host \
-t ocrd/all:medium-cuda .
docker build \
--progress=plain \
--build-arg BASE_IMAGE=ocrd/all:medium-cuda \
--build-arg VCS_REF=$(git rev-parse --short HEAD) \
--build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
--build-arg OCRD_MODULES="cor-asv-ann core dinglehopper docstruct eynollah format-converters nmalign ocrd_anybaseocr ocrd_calamari ocrd_cis ocrd_detectron2 ocrd_doxa ocrd_fileformat ocrd_froc ocrd_im6convert ocrd_
keraslm ocrd_kraken ocrd_olahd_client ocrd_olena ocrd_pagetopdf ocrd_repair_inconsistencies ocrd_segment ocrd_tesserocr ocrd_wrap sbb_binarization workflow-configuration" \
--build-arg PIP_OPTIONS="-e" \
--build-arg PARALLEL="" \
--build-arg PYTHON="python3" \
--network=host \
-t ocrd/all:maximum-cuda .
```

Which leads to exceeding the 1h time limit...

Dockerfile
bertsky: Perfect!

kba:
``

> It looks like `make docker-maximum-cuda` does not just build that image:
> 
> ```
> > make -n docker-maximum-cuda NO_UPDATE=1
> docker build \
> --progress=plain \
> --build-arg BASE_IMAGE=ocrd/core-cuda:v2.66.1 \
> --build-arg VCS_REF=$(git rev-parse --short HEAD) \
> --build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
> --build-arg OCRD_MODULES="ocrd_cis ocrd_fileformat ocrd_im6convert ocrd_pagetopdf ocrd_repair_inconsistencies ocrd_tesserocr ocrd_wrap workflow-configuration ocrd_olahd_client" \
> --build-arg PIP_OPTIONS="-e" \
> --build-arg PARALLEL="" \
> --build-arg PYTHON="python3" \
> --network=host \
> -t ocrd/all:minimum-cuda .
> docker build \
> --progress=plain \
> --build-arg BASE_IMAGE=ocrd/all:minimum-cuda \
> --build-arg VCS_REF=$(git rev-parse --short HEAD) \
> --build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
> --build-arg OCRD_MODULES="cor-asv-ann dinglehopper docstruct format-converters nmalign ocrd_calamari ocrd_cis ocrd_fileformat ocrd_im6convert ocrd_keraslm ocrd_olahd_client ocrd_olena ocrd_pagetopdf ocrd_repair_in
> consistencies ocrd_segment ocrd_tesserocr ocrd_wrap workflow-configuration" \
> --build-arg PIP_OPTIONS="-e" \
> --build-arg PARALLEL="" \
> --build-arg PYTHON="python3" \
> --network=host \
> -t ocrd/all:medium-cuda .
> docker build \
> --progress=plain \
> --build-arg BASE_IMAGE=ocrd/all:medium-cuda \
> --build-arg VCS_REF=$(git rev-parse --short HEAD) \
> --build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
> --build-arg OCRD_MODULES="cor-asv-ann core dinglehopper docstruct eynollah format-converters nmalign ocrd_anybaseocr ocrd_calamari ocrd_cis ocrd_detectron2 ocrd_doxa ocrd_fileformat ocrd_froc ocrd_im6convert ocrd_
> keraslm ocrd_kraken ocrd_olahd_client ocrd_olena ocrd_pagetopdf ocrd_repair_inconsistencies ocrd_segment ocrd_tesserocr ocrd_wrap sbb_binarization workflow-configuration" \
> --build-arg PIP_OPTIONS="-e" \
> --build-arg PARALLEL="" \
> --build-arg PYTHON="python3" \
> --network=host \
> -t ocrd/all:maximum-cuda .
> ```
> 
> Which leads to exceeding the 1h time limit...

That is the correct behavior as of #436. Upping the `resource_class` from the implied `medium` to `large` fixes the build time. Merged into #441.

bertsky:
> Which leads to exceeding the 1h time limit...

That's odd. We used to be well below 1h with the new multi-stage build (core → minimum → medium → maximum), which we switched to recently in #436. In fact, it was [just 47min](https://github.com/OCR-D/ocrd_all/runs/25450020458). It now times out during layer export – perhaps in this case, merely the network side happened to be slow?

Or is `resource_class: large` somehow having an opposite effect here?

bertsky:
> Upping the `resource_class` from the implied `medium` to `large` fixes the build time. Merged into #441.

Ah, I did not notice this was the most recent change. Thanks!

PR #445, opened 2024-07-15
Update 2024 07 15

kba:

PR #446, opened 2024-07-17
Update ocrd_fileformat

stweil:


kba:
I don't find the difference in state of ocrd_fileformat here from the one in #445 - AFAICT ocrd_fileformat is at the same commit ff46bd7. But if that were the case shouldn't Git(Hub) say "already up-to-date"?

stweil:
You are right, commit ab0ae664359ca9bff5d65a5254b9f1629dd50905 updated ocrd_fileformat.

stweil:
> But if that were the case shouldn't Git(Hub) say "already up-to-date"?

Maybe it did not do so because my PR was made before your PR was merged.

PR #447, opened 2024-08-01
Dockerfile: default for OCRD_MODULES was outdated

kba:
This only happens if users are building with `docker build` directly instead of through `make docker*` so we only noticed now. Perhaps it would be wise to change the Dockerfile to not have a default for `OCRD_MODULES` to encourage users to use the Makefile (which among other things makes sure that submodules are checked out).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ ENV TF_FORCE_GPU_ALLOW_GROWTH=true
 
 # allow passing build-time parameter for list of tools to be installed
 # (defaults to medium, which also requires those modules to be present)
-ARG OCRD_MODULES="core dinglehopper format-converters ocrd_calamari ocrd_cis ocrd_im6convert ocrd_keraslm ocrd_olena ocrd_segment ocrd_tesserocr tesseract tesserocr cor-asv-ann workflow-configuration"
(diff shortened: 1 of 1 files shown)


bertsky (approved):
right – thanks!

bertsky:
CI failure seems to come from:
```
#12 280.1 rm -f -rf /build/ocrd_tesserocr/build/tesseract
#12 280.1 cd repo/tesseract; make distclean
#12 280.1 make[2]: Entering directory '/build/ocrd_tesserocr/repo/tesseract'
#12 280.1 make[2]: Leaving directory '/build/ocrd_tesserocr/repo/tesseract'
#12 280.1 make[2]: *** No rule to make target 'distclean'.  Stop.
#12 280.1 make[1]: [Makefile:184: clean-tesseract] Error 2 (ignored)
```

So the `clean-tesseract` rule in ocrd_tesserocr seems to be wrong:
https://github.com/OCR-D/ocrd_tesserocr/blob/d23992b0068e6fa9612cddcfe106628461538442/Makefile#L182-L184

bertsky:
Fixed in ocrd_tesserocr master – please update that submodule here!

bertsky:
> Fixed in ocrd_tesserocr master – please update that submodule here!

wait – **not** fixed yet!

PR #448, opened 2024-08-17
Disable ocrd_anybaseocr and sbb_binarization for Python 3.11

stweil:


kba (commented):
I would prefer to first do the refactoring of ocrd_anybaseocr. sbb_binarization will be gone once we have the binarization made available from eynollah. I am just swamped with other work but hope to get back to ocrd_all and core next week. But let's discuss in the tech call later.

stweil:
Sorry, I could not attend the meeting today. Both modules are disabled for Python 3.11 because they break the build today. So there are not many alternatives to this pull request: neither forbidding Python 3.11 nor requiring local patches would be good solutions. If things change in the future, a module can be added again or be removed of course.

bertsky:
ocrd_anybaseocr has been refactored and updated, sbb_binarization removed

PR #449, opened 2024-08-20
Run ocrd network sample

joschrew:
This PR showcases the usage of ocrd network.

Step by step guide to run the example
- get o copy of this pr / repo and cd into it
- the docker images ocrd/all:maximum and ocrd/segment are currently used. Later more processor specific containers like ocrd/tesseocr should be used, but currently many of the images are not ready yet. Because of that ocrd/all is mainly used for now
- python or python3 with python-click is needed (e.g.: `sudo apt install python3-click`)
- Create a directory for the workspaces, which will be mounted to the containers: `mkdir /tmp/mydata`. This can be configured later. The folder must be owned by the current user
- create directory for some assets needed for now: `mkdir /tmp/path-to-my-assets`
- To run this example a custom logging-conf is needed due to current shortcomings of the ocrd/all docker image. Currently in the images the logfiles are present, owned by root which would cause errors if not using the custom logging conf. Therefore execute `cp run-network/my_ocrd_logging.conf /tmp/path-to-my-assets/`
- to run the tesserocr-recognize the Fraktur-data is needed which must be available, like the logging-conf: `curl -L "https://github.com/tesseract-ocr/tessdata_best/raw/main/script/Fraktur.traineddata" --output /tmp/path-to-my-assets/Fraktur.traineddata`
- run `make run-network` to start the setup
- copy a workflow to process to /tmp/mydata. The workspace I use is finally available here /tmp/mydata/vd18test/mets.xml together with its images in the DEFAULT filegroup. If this name differs, the workflow-script in the next step has to be adjusted
- create the workflow script in file workflow.txt:
```
cis-ocropy-binarize      -I DEFAULT -O O-1
anybaseocr-crop          -I O-1     -O O-2
cis-ocropy-denoise       -I O-2     -O O-3  -P dpi 300 -P level-of-operation page -P noise_maxsize 3.0
cis-ocropy-deskew        -I O-3     -O O-4  -P level-of-operation page
tesserocr-segment-region -I O-4     -O O-5  -P padding 5 -P find_tables false -P dpi 300
segment-repair           -I O-5     -O O-6  -P plausibilize true -P plausibilize_merge_min_overlap 0.7
cis-ocropy-clip          -I O-6     -O O-7
cis-ocropy-segment       -I O-7     -O O-8  -P spread 2.4 -P dpi 300
cis-ocropy-dewarp        -I O-8     -O O-9
tesserocr-recognize      -I O-9     -O PAGE -P textequiv_level word -P model Fraktur
fileformat-transform     -I PAGE    -O FULLTEXT
```
- run the workflow. For this example I use curl (ocrd network client could be used later): `curl -v -X POST "localhost:8000/workflow/run?mets_path=/data/vd18test/mets.xml&page_wise=True" -H "Content-type: multipart/form-data" -F "workflow=@workflow.txt"`
- optionally query the status: `curl localhost:8000/workflow/job/{your-job-id-from-prev-step} | jq`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -886,6 +886,13 @@ docker: DOCKER_MODULES ?= $(OCRD_MODULES)
 docker: DOCKER_PARALLEL ?= -j1
 docker: docker-latest
 
+OCRD_NETWORK_CONFIG ?= run-network/odem-workflow-config.yaml
+.PHONY: run-network
+run-network:
+	@run-network/creator.py create-docker $(OCRD_NETWORK_CONFIG)
+	@run-network/creator.py create-env $(OCRD_NETWORK_CONFIG)
+	@run-network/creator.py start $(OCRD_NETWORK_CONFIG)
+
 # do not search for implicit rules here:
 Makefile: ;
 local.mk: ;
--- a/run-network/creator.py
+++ b/run-network/creator.py
@@ -1,0 +1,350 @@
+#!/usr/bin/python3
+import re
+import time
+from dataclasses import dataclass, field
+from os import chdir, environ
+from pathlib import Path
+from typing import Any, Dict, ForwardRef, List, Optional, Type
+from collections import Counter
+import subprocess
+import requests
+
+import click
+import yaml
+
+
+@click.group()
+def cli():
+    """A simple CLI program"""
+    pass
+
+
+@cli.command("create-docker")
+@click.argument("config_path")
+def create_docker_cli(config_path: str):
+    """Creates a docker-compose file"""
+    config: Config = Config.from_file(config_path)
+    create_docker_compose(config)
+
+
+@cli.command("create-env")
+@click.argument("config_path")
+def create_env_cli(config_path: str):
+    """Creates .env for docker-compose"""
+    config: Config = Config.from_file(config_path)
+    create_env(config.environment, config.dest_env)
+
+
+@cli.command()
+@click.argument("config_path")
+def start(config_path):
+    """Start docker-compose in base_dir"""
+    config: Config = Config.from_file(config_path)
+    dest = Path(config.dest)
+    chdir(dest.parent)
+    environ["PWD"] = str(dest.parent)
+    command = ["docker-compose", "-f", f"{dest.name}", "up", "-d"]
+    subprocess.run(command)
+    wait_for_startup(f"http://localhost:{config.environment.ocrd_ps_port}")
+
+
+# @cli.command()
+# @click.argument("config_path")
+# def test_config(config_path):
+#     """Validate the configuration file.
+#
+#     This needs external dependency jsonschema"""
+#     from jsonschema import validate
+#     config_path = Path(config_path)
+#     schema_path = Path("creator_schema.yaml")
+#     if not config_path.exists():
+#         print("config file not found")
+#         exit(1)
+#     assert schema_path.exists()
+#
+#     with open(schema_path, "r") as fin:
+#         schema = yaml.safe_load(fin)
+#
+#     with open(config_path, "r") as fin:
+#         instance = yaml.safe_load(fin)
+#     validate(instance, schema)
+
+
+def create_docker_compose(config: Type[ForwardRef("Config")]):
+    """Create docker-compose file from config-object
+
+    The parts of the docker-compose are defined in the config-object. Basically there is a template
+    string for all needed services. These templates are configurable and parts of it are set via
+    info specified in the config file
+    """
+    with open(config.dest, "w") as fout:
+
+        if config.environment.mtu:
+            fout.write(config.network_template)
+            fout.write("\n")
+        fout.write("services:")
+        ps_template = re.sub(
+            r"{{[\s]*image[\s]*}}",
+            config.processing_server_image,
+            config.processing_server_template,
+        )
+        fout.write(ps_template)
+        fout.write(config.mongodb_template)
+        fout.write(config.rabbitmq_template)
+        fout.write(create_workers(config))
+
+
+def create_workers(config: Type[ForwardRef("Config")]):
+    """Create service definition of docker-compose for needed processors
+
+    This function reads the processor-template an replaces placeholders with info from the
+    config-object
+    """
+    res = ""
+    services_counter = Counter()
+    for p in config.processors:
+        service_name = p.name
+        services_counter[service_name] += 1
+        if services_counter[service_name] > 1:
+            service_name = f'{service_name}{services_counter[service_name]}'
+
+        proc_str = re.sub(r"{{[\s]*service_name[\s]*}}", service_name, config.proc_template)
+        proc_str = re.sub(r"{{[\s]*processor_name[\s]*}}", p.name, proc_str)
+        proc_str = re.sub(r"{{[\s]*image[\s]*}}", p.image, proc_str)
+
+        depends_on_str = ""
+        for depends_on in p.depends_on:
+            depends_on_str += "\n"
+            depends_on_str += f"      - {depends_on}"
+        proc_str = re.sub(r"{{[\s]*depends_on[\s]*}}", f"{depends_on_str}", proc_str)
+
+        # add volume mounts for some containers
+        for vol in p.volumes:
+            proc_str = re.sub(
+                r"    volumes:",
+                f'    volumes:\n      - "{vol}"',
+                proc_str,
+            )
+
+        for env in p.environment:
+            proc_str = re.sub(
+                r"    environment:",
+                f"    environment:\n      - {env}",
+                proc_str,
+            )
+
+        res += proc_str
+    return res
+
+
+def create_env(env: Type[ForwardRef("Environment")], dest: str):
+    """Create .env file to configure docker-compose
+
+    Info is read from the config-object and written to the env file
+    """
+    lines = []
+    if env.mtu:
+        lines.append(f"OCRD_PS_MTU={env.mtu}")
+    if env.ocrd_ps_port:
+        lines.append(f"OCRD_PS_PORT={env.ocrd_ps_port}")
+    if env.mongodb_user:
+        lines.append(f"MONGODB_USER={env.mongodb_user}")
+    if env.mongodb_pass:
+        lines.append(f"MONGODB_PASS={env.mongodb_pass}")
+    if env.mongodb_url:
+        lines.append(f"MONGODB_URL={env.mongodb_url}")
+    if env.rabbitmq_user:
+        lines.append(f"RABBITMQ_USER={env.rabbitmq_user}")
+    if env.rabbitmq_pass:
+        lines.append(f"RABBITMQ_PASS={env.rabbitmq_pass}")
+    if env.rabbitmq_url:
+        lines.append(f"RABBITMQ_URL={env.rabbitmq_url}")
+    if env.user_id:
+        lines.append(f"USER_ID={env.user_id}")
+    if env.group_id:
+        lines.append(f"GROUP_ID={env.group_id}")
+    if env.data_dir_host:
+        lines.append(f"DATA_DIR_HOST={env.data_dir_host}")
+    if env.internal_callback_url:
+        lines.append(f"INTERNAL_CALLBACK_URL={env.internal_callback_url}")
+
+    with open(dest, "w+") as fout:
+        fout.write("\n".join(lines))
+
+
+def wait_for_startup(processing_server_url: str):
+    """Wait for completed startup of all docker-compose services
+
+    After the startup the containers need some time to be usable. This function ensures their
+    availability
+    """
+    counter = 0
+    while True:
+        try:
+            response = requests.get(processing_server_url)
+            response.raise_for_status()
+            break
+        except requests.exceptions.ConnectionError:
+            time.sleep(1)
+            counter += 1
+            if counter > 30:
+                raise Exception("processing-server startup failed") from None
+        except requests.HTTPError:
+            # unexpected error
+            exit(1)
+
+
+NETWORK_TEMPLATE = """
+networks:
+  default:
+    driver: bridge
+    driver_opts:
+      com.docker.network.driver.mtu: ${OCRD_PS_MTU}
+"""
+
+PROC_TEMPLATE = """
+  {{ service_name }}:
+    image: {{ image }}
+    container_name: {{ service_name }}
+    command: {{ processor_name}} worker --database $MONGODB_URL --queue $RABBITMQ_URL
+    depends_on: {{ depends_on }}
+    user: "${USER_ID}:${GROUP_ID}"
+    volumes:
+      - "${DATA_DIR_HOST}:/data"
+    environment:
+      - OCRD_NETWORK_LOGS_ROOT_DIR=${LOGS_DIR:-/data/logs}
+"""
+
+PROCESSING_SERVER_TEMPLATE = """
+  ocrd-processing-server:
+    container_name: ocrd-processing-server
+    image: {{ image }}
+    environment:
+      - MONGODB_USER=${MONGODB_USER:-admin}
+      - MONGODB_PASS=${MONGODB_PASS:-admin}
+      - RABBITMQ_USER=${RABBITMQ_USER:-admin}
+      - RABBITMQ_PASS=${RABBITMQ_PASS:-admin}
+      - OCRD_NETWORK_SOCKETS_ROOT_DIR=${SOCKETS_DIR:-/data/sockets}
+      - OCRD_NETWORK_LOGS_ROOT_DIR=${LOGS_DIR:-/data/logs}
+    command: |
+      /bin/bash -c "echo -e \\"
+        internal_callback_url: ${INTERNAL_CALLBACK_URL}
+        use_tcp_mets: true
+        process_queue:
+          address: ocrd-rabbitmq
+          port: 5672
+          skip_deployment: true
+          credentials:
+            username: ${RABBITMQ_USER}
+            password: ${RABBITMQ_PASS}
+        database:
+          address: ocrd-mongodb
+          port: 27017
+          skip_deployment: true
+          credentials:
+            username: ${MONGODB_USER}
+            password: ${MONGODB_PASS}
+        hosts: []\\" > /data/ocrd-processing-server-config.yaml && \\
+        ocrd network processing-server -a 0.0.0.0:8000 /data/ocrd-processing-server-config.yaml"
+    user: "${USER_ID}:${GROUP_ID}"
+    volumes:
+      - "${DATA_DIR_HOST}:/data"
+    ports:
+      - ${OCRD_PS_PORT}:8000
+"""
+
+MONGODB_TEMPLATE = """
+  ocrd-mongodb:
+    container_name: ocrd-mongodb
+    image: mongo:latest
+    environment:
+      - MONGO_INITDB_ROOT_USERNAME=${MONGODB_USER:-admin}
+      - MONGO_INITDB_ROOT_PASSWORD=${MONGODB_PASS:-admin}
+    ports:
+      - "27018:27017"
+"""
+
+RABBITMQ_TEMPLATE = """
+  ocrd-rabbitmq:
+    container_name: ocrd-rabbitmq
+    image: rabbitmq:3-management
+    environment:
+      - RABBITMQ_DEFAULT_USER=${RABBITMQ_USER:-admin}
+      - RABBITMQ_DEFAULT_PASS=${RABBITMQ_PASS:-admin}
+    ports:
+      - "5672:5672"
+      - "15672:15672"
+"""
+
+
+@dataclass
+class Processor:
+    """Configuration of an ocr-d processor"""
+
+    name: str
+    image: str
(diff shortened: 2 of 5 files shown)


run-network/creator.py
```diff
@@ -0,0 +1,350 @@
+#!/usr/bin/python3
```
stweil, line 1: A fixed path for Python won't allow running in a virtual Python environment or with a different Python version.

run-network/creator.py
```diff
@@ -0,0 +1,350 @@
...
+        fout.write(ps_template)
+        fout.write(config.mongodb_template)
+        fout.write(config.rabbitmq_template)
+        fout.write(create_workers(config))
+
+
+def create_workers(config: Type[ForwardRef("Config")]):
+    """Create service definition of docker-compose for needed processors
+
+    This function reads the processor-template an replaces placeholders with info from the
```
stweil, line 100:
```suggestion
    This function reads the processor-template and replaces placeholders with info from the
```

stweil:
Are there already other showcases and documents for `ocrd network` or even complete installations? I recently started my own first experiments with it (based on a native OCR-D installation, no Docker), found only the documentation which is included in the code and in the specification and therefore appreciate this new sample.

run-network/creator.py
```diff
@@ -0,0 +1,350 @@
...
+from dataclasses import dataclass, field
+from os import chdir, environ
+from pathlib import Path
+from typing import Any, Dict, ForwardRef, List, Optional, Type
+from collections import Counter
+import subprocess
+import requests
+
+import click
+import yaml
```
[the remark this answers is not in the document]
joschrew: Thanks for your suggestions. I have used isort in the latest commit. FYI I use `max-line-length` with flake8

joschrew:
Thank you for your help.

I' m afraid I don't know of any other example deployment for ocrd network by now. I have focused on the docker-deployment so far, because no native installation of processors is needed.

MehmedGIT:
> Are there already other showcases and documents for `ocrd network` or even complete installations? I recently started my own first experiments with it (based on a native OCR-D installation, no Docker), found only the documentation which is included in the code and in the specification and therefore appreciate this new sample.

@stweil, there is also this pad with fast instructions that we were not able to put under `ocrd_network` docs yet: https://pad.gwdg.de/Ty6IXzhIRa6AvDdC4kTy_g#. Let me know if you need further assistance. Glad to help.

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+        "-I", input_file_grp,
+        "-O", output_file_grp,
+        "--block",
+        "--print-state",
+    ]
+    for (key, value) in parameter_override:
+        args.append("-P")
+        args.append(key)
+        args.append(value)
+    ocrd_cli(args)
```
bertsky, line 365: I would suggest improving these templates as follows:
1. use a `format`table string, so variables can be filled in directly, without regex matching
2. delegate to the click CLI directly, so its arg parsing and help strings can be reused

```suggestion
DELEGATOR_PROCESSOR_TEMPLATE = """#!/usr/bin/env python

from ocrd_network.cli import client_cli

run_cli = client_cli.commands['processing'].commands['run']
def callback(*args, **kwargs):
    kwargs['address'] = "http://localhost:{ps_port}"
    kwargs['block'] = True
    kwargs['print_state'] = True
    return run_cli.callback({processor_name}, *args, **kwargs)

params = [param for param in run_cli.params
          if param.name not in [
                  'processor_name',
                  'address',
                  'block',
                  'print_state',
          ]]
cli = click.Command(name={processor_name},
                    callback=callback,
                    params=params)

```

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+def cli(mets, workflow):
+    address = f"http://localhost:{PROCESSING_SERVER_PORT}"
+    ocrd_cli([
+        "network", "client", "workflow", "run",
+        "--address", address,
+        "-m", mets,
+        "-w", workflow,
+        "--block",
+        "--print-state",
+    ])
```
bertsky, line 394: similarly here

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+            f'PROCESSOR_NAME = "{proc.name}"',
+            DELEGATOR_PROCESSOR_TEMPLATE.lstrip(),
+            1
+        )
+        content = re.sub(
+            r"PROCESSING_SERVER_PORT\s*=\s*[0-9]+",
+            f"PROCESSING_SERVER_PORT = {port}",
+            content,
+            1
+        )
```
bertsky, line 80: So here it would simply become:
```suggestion
        content = DELEGATOR_PROCESSOR_TEMPLATE.format(processor_name=proc.name, ps_port=port)
```

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+        with open(dest, "w") as fout:
+            fout.write(content)
+        chmod(dest, 0o755)
+
+    content = re.sub(
+        r"PROCESSING_SERVER_PORT\s*=\s*[0-9]+",
+        f"PROCESSING_SERVER_PORT = {port}",
+        DELEGATOR_WORKFLOW_TEMPLATE.lstrip(),
+        1
+    )
```
bertsky, line 91: and analogously

Makefile
```diff
@@ -886,6 +886,18 @@ docker: DOCKER_MODULES ?= $(OCRD_MODULES)
...
+OCRD_NETWORK_CONFIG ?= run-network/odem-workflow-config.yaml
+.PHONY: run-network
+run-network:
+	@$(PYTHON) -m venv $(VIRTUAL_ENV)
+	@$(VIRTUAL_ENV)/bin/pip install click requests pyaml shapely==1.8.5 ocrd
+	@$(VIRTUAL_ENV)/bin/python run-network/creator.py create-docker $(OCRD_NETWORK_CONFIG)
+	@$(VIRTUAL_ENV)/bin/python run-network/creator.py create-env $(OCRD_NETWORK_CONFIG)
+	@$(VIRTUAL_ENV)/bin/python run-network/creator.py start $(OCRD_NETWORK_CONFIG)
+
+	@$(VIRTUAL_ENV)/bin/python run-network/creator.py create-clients $(VIRTUAL_ENV)/bin $(OCRD_NETWORK_CONFIG)
```
bertsky, line 898: I would recommend differentiating between setup phase (just once) and runtime (repeatable) here. Also, we should provide a teardown and a cleanup phase. Also, I would not mix this with our existing `VIRTUAL_ENV` variable (in contrast to which this time we fully control the venv).

How about something like this:
```suggestion
network-setup:
	$(PYTHON) -m venv run-network/.venv
	run-network/.venv/bin/python -m pip install click requests pyaml shapely==1.8.5 ocrd
	run-network/.venv/bin/python run-network/creator.py create-compose $(OCRD_NETWORK_CONFIG)
	run-network/.venv/bin/python run-network/creator.py create-dotenv $(OCRD_NETWORK_CONFIG)
	run-network/.venv/bin/python run-network/creator.py create-clients run-network/.venv/bin $(OCRD_NETWORK_CONFIG)

network-start:
	run-network/.venv/bin/python run-network/creator.py start $(OCRD_NETWORK_CONFIG)

network-stop:
	run-network/.venv/bin/python run-network/creator.py stop $(OCRD_NETWORK_CONFIG)

network-clean:
	$(RM) -r run-network/.venv run-network/.env run-network/docker-compose.yml

```

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+import yaml
+
+
+@click.group()
+def cli():
+    """A simple CLI program"""
+    pass
+
+
+@cli.command("create-docker")
```
bertsky, line 22:
```suggestion
@cli.command("create-compose")
```

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+@click.argument("config_path")
+def create_docker_cli(config_path: str):
+    """Creates a docker-compose file"""
+    config: Config = Config.from_file(config_path)
+    create_docker_compose(config)
+
+
+@cli.command("create-env")
```
bertsky, line 30:
```suggestion
@cli.command("create-dotenv")
```

run-network/creator.py
```diff
@@ -0,0 +1,473 @@
...
+def start(config_path):
+    """Start docker-compose in base_dir"""
+    config: Config = Config.from_file(config_path)
+    dest = Path(config.dest)
+    chdir(dest.parent)
+    environ["PWD"] = str(dest.parent)
+    command = ["docker-compose", "-f", f"{dest.name}", "up", "-d"]
+    subprocess.run(command)
+    wait_for_startup(f"http://localhost:{config.environment.ocrd_ps_port}")
+
```
bertsky, line 49:
```suggestion

@cli.command()
@click.argument("config_path")
def stop(config_path):
    """Stop docker-compose services in base_dir"""
    config: Config = Config.from_file(config_path)
    dest = Path(config.dest)
    chdir(dest.parent)
    command = ["docker-compose", "-f", f"{dest.name}", "down"]
    subprocess.run(command)

```

joschrew:
Thank you for you input, I think I included all parts in the last commit: [d3fa81f](https://github.com/OCR-D/ocrd_all/pull/449/commits/d3fa81ffc2384a0ec49bbe4f53ea4731b3ddd761). As already noted this now depends on https://github.com/OCR-D/core/pull/1303

Makefile
bertsky: My suggestion advised splitting up the setup phase into file-based targets. The reason is simply that this allows selectively repeating/updating these steps. Also, if some file gets edited manually, the new time stamp will prevent it being overwritten. Also, adding more explicit dependencies (like on `OCRD_NETWORK_CONFIG`) could then further help manage this more smoothly.

You have not adopted this. Did my rules not work?

joschrew: I think I missed the edit. Updated with: https://github.com/OCR-D/ocrd_all/pull/449/commits/b9d0f84a50a78325490e3ce136b1ff030b9692d2 and should fit now

Makefile
```diff
@@ -886,7 +886,8 @@ docker: DOCKER_MODULES ?= $(OCRD_MODULES)
 docker: DOCKER_PARALLEL ?= -j1
 docker: docker-latest
 
-OCRD_NETWORK_CONFIG ?= run-network/odem-workflow-config.yaml
+OCRD_NETWORK_CONFIG ?= run-network/ocrd-all-config.yaml
+OCRD_NETWORK_PROFILE ?= minimum
```
bertsky, line 890: Note: strictly, we would not even have to introduce a new variable for this: We could just advice users to utilise Docker's default variable `COMPOSE_PROFILES` for this.
```suggestion
```

run-network/creator.py
```diff
@@ -37,13 +37,17 @@ def create_env_cli(config_path: str):
 
 @cli.command()
 @click.argument("config_path")
-def start(config_path):
+@click.argument("profile", default="minimum", type=click.Choice(["minimum", "medium", "maximum"]))
+def start(config_path, profile):
```
bertsky, line 41: So in here, we would not need an extra argument:
```suggestion
def start(config_path):
```

The only downside I can think of is that the runtime choice cannot be type-checked by Click. (So instead, Compose would fail with a more cryptic error message.)

On the other hand, if the config file does get extended by additional profiles (in the future / by the user), then we would not interfere here (or have to keep the code in sync).

run-network/creator.py
```diff
@@ -37,13 +37,17 @@ def create_env_cli(config_path: str):
...
     config: Config = Config.from_file(config_path)
     dest = Path(config.dest)
     chdir(dest.parent)
     environ["PWD"] = str(dest.parent)
-    command = ["docker-compose", "-f", f"{dest.name}", "up", "-d"]
+    command = ["docker-compose", "-f", f"{dest.name}"]
+    if profile:
+        command += ["--profile", profile]
+    command += ["up", "-d"]
     subprocess.run(command)
```
bertsky, line 51: IINM, here it would simply be:
```suggestion
    command += ["up", "-d"]
    subprocess.run(command)
```

So we don't even need a `env={"COMPOSE_PROFILES": os.getenv("COMPOSE_PROFILES")}` for run, because the shell (and Python) will simply inherit this. Python's `subprocess.Popen` docs say:
>  If `env` is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of inheriting the current process’ environment, which is the default behavior.

Have not tested, though.

run-network/ocrd-all-config.yaml
```diff
@@ -0,0 +1,192 @@
+dest: docker-compose.yml
```
bertsky, line 1: BTW, we could also split up the docker-compose.yml into a fixed (template) part and then [include](https://docs.docker.com/reference/compose-file/include/) the generated configs. (So for example, we could include docker-compose.servers.yml and docker-compose.processors.yml.)

bertsky (commented):
Getting better and better!

Makefile
bertsky: For example:

    make network-start COMPOSE_PROFILES=minimum
    # or, equivalently (in this case)
    COMPOSE_PROFILES=minimum make network-start

Makefile
bertsky: Thanks!

bertsky:
We now have to start thinking about how to generate `ocrd-all-tool.json` (and the newly introduced `ocrd-all-config.yaml`, which could in principle itself be generated from the former, if we would also include the Dockerhub image name into the ocrd-tool.json schema, and introduce a central configuration file mapping modules/subrepos to profiles, which information used to be just part of the `docker` rules).

So far, we required a git checkout of all (enabled) submodules (i.e. `OCRD_MODULES`), and then ran `ocrd-all-tool.py` to concatenate all the individual `*/ocrd-tool.json` files. But for the network setting, this seems too expensive: we already fetch all the module Docker images anyway, so perhaps we should instead `docker cp` each ocrd-tool.json from the images into the CWD. But that in turn requires a list of Dockerhub image names for all (enabled) modules in advance...

Makefile
joschrew: I have now committed these 3 changes regarding the profiles

run-network/creator.py
```diff
@@ -0,0 +1,461 @@
...
+PROC_TEMPLATE = """
+  {service_name}:
+    image: {image}
+    container_name: {service_name}
+    command: {processor_name} worker --database $MONGODB_URL --queue $RABBITMQ_URL
+    depends_on: {depends_on}
+    user: "${{USER_ID}}:${{GROUP_ID}}"
+    profiles: [{profiles}]
+    volumes:
+      - "${{DATA_DIR_HOST}}:/data"
```
bertsky, line 259: @joschrew what about the processor resources, shouldn't that be a volume, too?

Preferably as **named volume** (so module-distributed and user-downloaded files can mix freely), e.g.:
```suggestion
    volumes:
      - "${{DATA_DIR_HOST}}:/data"
      - ocrd-resources:/usr/local/share/ocrd-resources
```

Unfortunately, it seems we missed the opportunity of the latest sweep across modules to define a unique internal resource location. In `ocrd/tesserocr`, we already use `/models` as alias for `/usr/local/share/ocrd-resources`, but in many other images we still use the latter...

bertsky: Perhaps brevity (`/models` instead of `/usr/local/share/ocrd-resources`) is not an issue if we are using scriped mounts and calls, anyway. And `/models` was a _hack_ of sorts: it was not backed by the spec (which still says /usr/local is the `system` location, and `$XDG_DATA_DIR` the `data` location), but only a convention for our fat container images.

Therefore I will switch ocrd/tesserocr back to /usr/local/share/ocrd-resources as the only place to mount (as in all the other _workable_ slim container images).

bertsky: > Therefore I will switch ocrd/tesserocr back to /usr/local/share/ocrd-resources as the only place to mount (as in all the other _workable_ slim container images).

https://github.com/OCR-D/ocrd_tesserocr/commit/70caca02260448cfbc7a76facc5bf6c4bd324691

joschrew: I have added a resource-volume now. What would you suggest to get the resources into the volume? I tried using the resmgr which fails because of missing writing permission on /.config (the setup cannot not use root permissions). Would it be a good idea to to this via the Makefile?

joschrew:
> We now have to start thinking about how to generate `ocrd-all-tool.json` (and the newly introduced `ocrd-all-config.yaml`, which could in principle itself be generated from the former, ~if we would also include the Dockerhub image name into~ via `dockerhub` key in the ocrd-tool.json schema, and introduce a central configuration file mapping modules/subrepos to profiles, which information used to be just part of the `docker` rules).
> 
> So far, we required a git checkout of all (enabled) submodules (i.e. `OCRD_MODULES`), and then ran `ocrd-all-tool.py` to concatenate all the individual `*/ocrd-tool.json` files. But for the network setting, this seems too expensive: we already fetch all the module Docker images anyway, so perhaps we should instead `docker cp` each ocrd-tool.json from the images into the CWD. But that in turn requires a list of Dockerhub image names for all (enabled) modules in advance...

This is still an open issue. Currently simply a copied ocrd-all-tool.json is used. But maybe this is the best way for now

run-network/creator.py
bertsky: > What would you suggest to get the resources into the volume? I tried using the resmgr which fails because of missing writing permission on /.config (the setup cannot not use root permissions). Would it be a good idea to to this via the Makefile?

Ah, that bit us before in the fat containers, so I thought we found a remedy for it: setting `XDG_CONFIG_HOME=/usr/local/share/ocrd-resources` as well, so the volume should cover both the resources themselves and the config file `ocrd/resources.yml`.

Your `/.config` indicates that `XDG_CONFIG_HOME` is not set (as `HOME=/`). What image is this  happening with?

So no, this should work on the user side already.

joschrew: I have now changed the ocrd-resources-volume mount to a lokal directory on the host instead of a named volume. This was to only way for me to use the resources with user's permissions which are necessary for correct permissions of workflow files

joschrew: All images. The setup is started as current user, this is necessary because the created workspace files should be of user's permission. If the volume is of root permission (or a named volume, same permissions) the resource download can only triggered by root. But when the container is started as user and then root user is used to download the resources we have logfile permission-errors again.

bertsky: Oh, right! We do have a problem with named volumes. They are always created as root. Even with `docker volume create`. It seems the only way to get normal ownership is by choosing a container-side parent path which already has the permissions we want (as these get inherited by the named volume). But in cases where the directory itself already exists (from the Docker build), its ownership won't change (i.e. stay root). That _is_ a big problem.

> The value of XDG_CONFIG_HOME has no effect on permissions.

It does have the effect of controlling where `ocrd/resources.yml` gets written.

bertsky: See https://github.com/moby/moby/issues/3124

Okay, so the only way out AFAICS is setting `/usr/local/share/ocrd-resources` (and its subdirectories) to 777 in the Dockerfile already. This means in _every_ Dockerfile.

You can test this: 
- revert https://github.com/OCR-D/ocrd_all/pull/449/commits/a9f2ffd65e36749980282e3d911659e3f58f94d3
- write a Dockerfile.fixup, contents:
```Dockerfile
ARG BASE_IMAGE
FROM $BASE_IMAGE
RUN mkdir -p /usr/local/share/ocrd-resources
RUN find /usr/local/share/ocrd-resources -type d -exec chmod 777 {} ";"
RUN find /usr/local/share/ocrd-resources -type f -exec chmod 666 {} ";"
```
- use it to patch a module of your choice, e.g.:
```sh
cat Dockerfile.fixup | docker build --build-arg BASE_IMAGE=ocrd/tesserocr -t ocrd/tesserocr:permissions -
```
- use the patched image, e.g.:
```sh
docker run --rm -it -v ocrd-resources:/usr/local/share/ocrd-resources ocrd/tesserocr:permissions ocrd resmgr download ocrd-tesserocr-recognize fra.traineddata
```

joschrew: I tried your suggestion, it nearly works for me (setting /usr/local/share/ocrd-resources itself to 777 is also necessary).
Another possibility would be to have more than one folder for the resources (kind of what we have currently). /usr/local/share/ocrd-resoucres could be used for preinstalled resources and the "on-demand"-resources could be installed to another folder (XDG_DATA_HOME/ocrd-resources for example) and for that folder a host-mounted-volume could be used. I'd prefer one folder for all resources but setting global permissions might rise security concerns again like it did when we talked about setting permissions for the logging-folders. Did we have a reason why we don't want to use XDG_DATA_HOME/ocrd-resources?

bertsky: > it nearly works for me (setting /usr/local/share/ocrd-resources itself to 777 is also necessary).

Are you sure you tried the exact command above (which should already cover the top-level directory as well)?

> Another possibility would be to have more than one folder for the resources [...]
> Did we have a reason why we don't want to use XDG_DATA_HOME/ocrd-resources?

Yes, we did have a reason. Not all processors can handle more than 1 location. Tesseract for example can not be brought to look up in more than 1 directory.

It's also not intuitive. You don't want to have to think about resources in terms of where they came from.

PR #454, opened 2024-12-18
switch to core API 3.0 branches

bertsky:
Not to be merged, merely for tracking available v3 branches:

- https://github.com/OCR-D/core/pull/1240 ... https://github.com/OCR-D/core/pull/1312 + https://github.com/OCR-D/core/pull/1313
- https://github.com/qurator-spk/eynollah/pull/130 ... https://github.com/qurator-spk/eynollah/pull/151
- https://github.com/OCR-D/ocrd_calamari/pull/118/
- https://github.com/OCR-D/ocrd_kraken/pull/44 ... https://github.com/OCR-D/ocrd_kraken/pull/46
- https://github.com/OCR-D/ocrd_segment/pull/69
- https://github.com/bertsky/ocrd_cis/pull/5 ... https://github.com/bertsky/ocrd_cis/commit/2cf3c85ced72df0de28573db9f2ea531beaf0a42
- https://github.com/OCR-D/ocrd_tesserocr/pull/216
- https://github.com/bertsky/ocrd_wrap/pull/12
- https://github.com/OCR-D/ocrd_pagetopdf/pull/28
- https://github.com/bertsky/workflow-configuration/pull/34/
- https://github.com/OCR-D/ocrd_olena/pull/99
- https://github.com/bertsky/nmalign/pull/1
- https://github.com/ASVLeipzig/cor-asv-ann/pull/15
- https://github.com/bertsky/ocrd_detectron2/pull/31
- https://github.com/OCR-D/ocrd_anybaseocr/pull/112
- https://github.com/bertsky/docstruct/pull/2
- https://github.com/OCR-D/ocrd_froc/pull/19 ... https://github.com/OCR-D/ocrd_froc/pull/20
- https://github.com/qurator-spk/dinglehopper/pull/128
- https://github.com/OCR-D/ocrd_olahd_client/pull/12
- https://github.com/OCR-D/ocrd_keraslm/pull/23
- https://github.com/bertsky/ocrd_doxa/commit/a6df2f21975c3f4e5c3a8d5bf24618845571819b
- https://github.com/OCR-D/page-to-alto/pull/40

No v3 ports yet:
- ocrd_fileformat
- ocrd_neat
- ~~ocrd_repair_inconsistencies~~ (discontinued, will be covered by ocrd-segment-repair)
- ~~ocrd_im6convert~~ (discontinued, covered by ocrd_wrap's `ocrd-preprocess-image`)
- ~~sbb_binarization~~ (discontinued, now covered by eynollah)

PR #456, opened 2025-04-07
Release 2025 04 07

kba:
With this PR, OCR-D/core is updated to version 3, which brings many improvements in the way processors work. While not all processors have been ported to make use of these improvements, most of the widely used ones are (cf. [this draft PR](https://github.com/OCR-D/ocrd_all/pull/454) by @bertsky tracking adoption) and those that are not can still be run with the older processing API.

ocrd_olena and ocrd_pagetopdf have been rewritten in Python and are now significantly faster and better tested.

ocrd_im6convert has been removed since that functionality is available in a better form in ocrd_wrap's `ocrd-preprocess-image`.

sbb_binarization has also been dropped, as that functionality has been merged into eynollah and remains available as `ocrd-sbb-binarize` from there.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -60,7 +60,7 @@
 	url = https://github.com/OCR-D/ocrd_fileformat.git
 [submodule "ocrd_pagetopdf"]
 	path = ocrd_pagetopdf
-	url = https://github.com/UB-Mannheim/ocrd_pagetopdf.git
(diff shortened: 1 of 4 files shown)


bertsky:
> ocrd_olena and ocrd_pagetopdf have been rewritten in Python and are now significantly faster and better tested.

Same is true for `ocrd-import` and `ocrd-page-transform` from https://github.com/bertsky/workflow-configuration btw.

PR #460, opened 2025-04-10
Disable ocrd_detectron2 and ocrd_segment on macOS

stweil:


kba (approved):
Fine by me if it helps the OSX crowd :)

PR #461, opened 2025-04-15
update modules (missing dockerhub key)

bertsky:
You may also want to include 
- https://github.com/OCR-D/core/pull/1321 (or v3.3.1) 
- https://github.com/qurator-spk/dinglehopper/pull/128
- https://github.com/OCR-D/ocrd_olahd_client/pull/12
- https://github.com/OCR-D/ocrd_keraslm/pull/23

– once they are merged, respectively.

### DIFF
--- a/ocrd-all-dockerhub.py
+++ b/ocrd-all-dockerhub.py
@@ -1,9 +1,0 @@
-import json
-import sys
-all = dict()
-for path in sys.argv[1:]:
-    tool = json.load(open(path))
-    dockerhub = tool.get('dockerhub', '')
-    for executable in tool['tools']:
-        all[executable] = dockerhub
-print(json.dumps(all, indent=2))
--- a/ocrd-all-dockerhub.py
+++ b/ocrd-all-dockerhub.py
@@ -1,0 +1,9 @@
+import json
+import sys
+all = dict()
+for path in sys.argv[1:]:
+    tool = json.load(open(path))
+    dockerhub = tool.get('dockerhub', '')
+    for executable in tool['tools']:
+        all[executable] = dockerhub
+print(json.dumps(all, indent=2))
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,6 +120,14 @@ commands:
             - store_artifacts:
                 path: ocrd-all-module-dir.json
                 destination: ocrd-all-module-dir.json
+            - run:
+                name: store ocrd-all-dockerhub.json
+                command: |
+                  id=`docker create ocrd/all:maximum`
+                  docker cp $id:/build/ocrd-all-dockerhub.json .
+            - store_artifacts:
+                path: ocrd-all-dockerhub.json
+                destination: ocrd-all-dockerhub.json
       - when:
           condition:
             equal: [ maximum-cuda, << parameters.tag >> ]
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ clean: # add more prerequisites for clean below
 	$(RM) -r $(SUB_VENV)
 	$(RM) -r $(CURDIR)/venv # deliberately not using VIRTUAL_ENV here
 	$(RM) -r $(HOME)/.parallel/semaphores/id-ocrd_*
-	$(RM) ocrd-all-tool.json ocrd-all-module-dir.json
+	$(RM) ocrd-all-tool.json ocrd-all-module-dir.json ocrd-all-dockerhub.json
 
 define HELP
 cat <<"EOF"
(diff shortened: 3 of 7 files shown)


ocrd-all-dockerhub.py
```diff
@@ -0,0 +1,9 @@
+import json
+import sys
+all = dict()
+for path in sys.argv[1:]:
+    tool = json.load(open(path))
+    dockerhub = tool.get('dockerhub', '')
+    for executable in tool['tools']:
+        all[executable] = dockerhub
```
MehmedGIT, line 8: Could we also add the version as well? Either as a dictionary `all[executable] = {"dockerhub": dockerhub, "version": version}` or any other way.

MehmedGIT (approved):
Looks great!

ocrd-all-dockerhub.py
bertsky: Sure! We could even put the `git_url` in there. And while we are at it – the commit SHA1.

Should we call the file `ocrd-all-meta.json` then?

MehmedGIT: > Sure! We could even put the git_url in there. And while we are at it – the commit SHA1.

That would be even better!

> Should we call the file `ocrd-all-meta.json` 

Yes.

bertsky: Like this?

(I also added some stderr diagnostics in all `all` scripts.)

mikegerber:
Please don't include the linked PR for dinglehopper and wait a release. The PR currently contains a  change of CER calculation that needs discussion. Otherwise we'd end up with a dinglehopper in ocrd_all that gives different results ...

I also wish if ocrd_all would start using the versioned releases - I specifically started doing those releases for OCR-D, but they don't seem to be used.

stweil:
> I also wish if ocrd_all would start using the versioned releases - I specifically started doing those releases for OCR-D, but they don't seem to be used.

Yes, please, not only for dinglehopper, but for all submodules.

mikegerber:
> Please don't include the linked PR for dinglehopper and wait for a release. The PR currently contains a change of CER calculation that needs discussion. Otherwise we'd end up with a dinglehopper in ocrd_all that gives different results ...

I've released dinglehopper v0.10.0 with the V3 update (but *not* the proposed change to using normalized CER. (For the latter there is a separate PR now.)

bertsky:
@mikegerber The dinglehopper update  was originally (the last remaining submodule) scheduled for this PR as our last release of the ocrd/all fat container images. For that, it **must** still support Python 3.8, which all our base stages are currently built on (because some module still rely on Tensorflow 1.x, which is only available prebuilt via nvidia-tensorflow for Py38). That can only be overcome by the slim containers (where every module can have its own base version of Python etc). So can we please get a version that does *not* rely on uniseg>=0.90? See [here](https://app.circleci.com/pipelines/github/bertsky/ocrd_all/75/workflows/fa04d07d-0eb2-4caa-938e-659671a2bf0a/jobs/109?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) for how this fails our ocrd/all builds. It all hinges on this module now.

I cannot speak to whether or not my `normalized-cer` branch should have been included. We had a frenzy of PRs and updates on our hands for v3 and related updates. So I guess combining this was to save time (as we also depend on usable results without `Infinity` or averages over CER results above 100%).

bertsky:
superseded by #462

mikegerber:
> @mikegerber The dinglehopper update was originally (the last remaining submodule) scheduled for this PR as our last release of the ocrd/all fat container images. For that, it **must** still support Python 3.8, which all our base stages are currently built on (because some module still rely on Tensorflow 1.x, which is only available prebuilt via nvidia-tensorflow for Py38). That can only be overcome by the slim containers (where every module can have its own base version of Python etc). So can we please get a version that does _not_ rely on uniseg>=0.9? See [here](https://app.circleci.com/pipelines/github/bertsky/ocrd_all/75/workflows/fa04d07d-0eb2-4caa-938e-659671a2bf0a/jobs/109?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) for how this fails our ocrd/all builds. It all hinges on this module now.

Ah, that's good context to know, I wasn't aware of this (and only checked EOLness of Python 3.8 after it broke). I will release 0.10.1 shortly with the patch!

mikegerber:
> I cannot speak to whether or not my `normalized-cer` branch should have been included. We had a frenzy of PRs and updates on our hands for v3 and related updates. So I guess combining this was to save time (as we also depend on usable results without `Infinity` or averages over CER results above 100%).

  I can say it definitely didn't save ME time. It's a rather intrusive change because it redefines CER and this needs much more thought w.r.t. proper integration. It gives different results for users! To be me it's clear that it needs more love than sneaking it in with a PR that was just supposed to update to OCR-D's V3 API.

PR #462, opened 2025-04-16
Update 2025 04 17

kba:
Includes #461 but separate PR because otherwise `release.sh changelog` would not produce the output we want.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,6 +120,14 @@ commands:
             - store_artifacts:
                 path: ocrd-all-module-dir.json
                 destination: ocrd-all-module-dir.json
+            - run:
+                name: store ocrd-all-dockerhub.json
+                command: |
+                  id=`docker create ocrd/all:maximum`
+                  docker cp $id:/build/ocrd-all-dockerhub.json .
+            - store_artifacts:
+                path: ocrd-all-dockerhub.json
+                destination: ocrd-all-dockerhub.json
       - when:
           condition:
             equal: [ maximum-cuda, << parameters.tag >> ]
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,7 +26,7 @@
         shallow = true
 [submodule "dinglehopper"]
 	path = dinglehopper
-	url = https://github.com/qurator-spk/dinglehopper.git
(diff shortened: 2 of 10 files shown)


bertsky (approved):
LGTM – thx!

mikegerber:
This contains the - as of yet undiscussed - changes to CER calculation in dinglehopper. Please don't include that in this release. it changes the results for users...

mikegerber:
> This contains the - as of yet undiscussed - changes to CER calculation in dinglehopper. Please don't include that in this release. it changes the results for users...

I've released dinglehopper v0.10.0 with the V3 update (but *not* the proposed change to using normalized CER. (For the latter there is a separate PR now.)

kba:
> > This contains the - as of yet undiscussed - changes to CER calculation in dinglehopper. Please don't include that in this release. it changes the results for users...
> 
> I've released dinglehopper v0.10.0 with the V3 update (but _not_ the proposed change to using normalized CER. (For the latter there is a separate PR now.)

You did, thank you, but you also changed minimum python version which broke the release. And I do need to finish this today because I still need to update our local deployment and start processing over the long weekend with the updated processors. We can switch back to upstream dinglehopper once the discussions are finished after Easter but for now, I really need to finish this, I know the fork works and will keep that version for now.

mikegerber:
> > > This contains the - as of yet undiscussed - changes to CER calculation in dinglehopper. Please don't include that in this release. it changes the results for users...
> > 
> > 
> > I've released dinglehopper v0.10.0 with the V3 update (but _not_ the proposed change to using normalized CER. (For the latter there is a separate PR now.)
> 
> You did, thank you, but you also changed minimum python version which broke the release. And I do need to finish this today because I still need to update our local deployment and start processing over the long weekend with the updated processors. We can switch back to upstream dinglehopper once the discussions are finished after Easter but for now, I really need to finish this, I know the fork works and will keep that version for now.

0.10.1 now adds support back for Python 3.8.

So the latest ocrd_all now contains a dinglehopper that produces different CER results compared to the upstream dinglehopper? :-( I do not understand why this change did not go through a proper process, e.g. PRs/issues!

PR #464, opened 2025-04-17
Drop ocrd_repair_inconsistencies and ocrd_pc_segmentation

kba:
Neither of these are actively maintained and probably not worth porting to v3 or even keeping in ocrd_all.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,9 +42,6 @@
 [submodule "ocrd_anybaseocr"]
 	path = ocrd_anybaseocr
 	url = https://github.com/OCR-D/ocrd_anybaseocr.git
-[submodule "ocrd_pc_segmentation"]
-	path = ocrd_pc_segmentation
-	url = https://github.com/ocr-d-modul-2-segmentierung/ocrd-pixelclassifier-segmentation.git
 [submodule "opencv-python"]
 	path = opencv-python
 	url = https://github.com/skvark/opencv-python.git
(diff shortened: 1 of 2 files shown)


bertsky (approved):
Agreed w.r.t. ocrd_pc_segmentation.

But ocrd_repair_inconsistencies may be worth keeping (and migrating to v3). (It should be more adequately called `ocrd_repair_order` or `ocrd_order_coords` IMO.)

On the other hand, these errors do show up as `ConsistencyError` in the `PageValidator` report. So perhaps it would be best suited for ocrd-segment-repair, which already covers acting on (i.e. repairing) non-textual consistency errors. I guess it would be a rather small change...

bertsky:
> On the other hand, these errors do show up as `ConsistencyError` in the `PageValidator` report. So perhaps it would be best suited for ocrd-segment-repair, which already covers acting on (i.e. repairing) non-textual consistency errors. I guess it would be a rather small change...

tracking in https://github.com/OCR-D/ocrd_segment/issues/70

mikegerber:
> But ocrd_repair_inconsistencies may be worth keeping (and migrating to v3). (It should be more adequately called `ocrd_repair_order` or `ocrd_order_coords` IMO.)
> 
> On the other hand, these errors do show up as `ConsistencyError` in the `PageValidator` report. So perhaps it would be best suited for ocrd-segment-repair, which already covers acting on (i.e. repairing) non-textual consistency errors. I guess it would be a rather small change...

If you want it, fork the project please. I agree that the name could be better and it may be better placed in ocrd-segment-repair.

PR #465, opened 2025-04-29
slim containers...

bertsky:
Since we agreed it is not feasible to continue supporting native installation into a shared venv (along with risk of dependency clashes, necessity to find compromises or sidestep via sub-venv), which will not be needed for the network (WebAPI server-client) installation anyway, this implements the first step: keeping backwards-compatible CLI interfaces, but delegating to Docker images throughout. (The second step will be about replacing or complementing the CLI interfaces with server-client setup, in line with #449.)

- [X] generate executables by delegating to slim-container Docker images
- [X] automagically prepare a shared named volume for models with user-friendly permissions and copying pre-installed models
- [X] create convenient CLIs for `ocrd resmgr` in each slim image
- [X] remove unnecessary native installation rules and definitions
- [X] remove unnecessary fat-container Docker build rules and definitions
- [ ] find a new solution for the `ocrd-all-*.json` targets
- [ ] ...

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -30,9 +30,6 @@
 [submodule "cor-asv-ann"]
 	path = cor-asv-ann
 	url = https://github.com/ASVLeipzig/cor-asv-ann.git
-[submodule "format-converters"]
-	path = format-converters
-	url = https://github.com/OCR-D/format-converters.git
 [submodule "ocrd_cis"]
 	path = ocrd_cis
 	url = https://github.com/cisocrgroup/ocrd_cis.git
@@ -42,10 +39,6 @@
 [submodule "ocrd_anybaseocr"]
 	path = ocrd_anybaseocr
 	url = https://github.com/OCR-D/ocrd_anybaseocr.git
-[submodule "opencv-python"]
-	path = opencv-python
-	url = https://github.com/skvark/opencv-python.git
-        shallow = true
 [submodule "workflow-configuration"]
 	path = workflow-configuration
 	url = https://github.com/bertsky/workflow-configuration.git
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,6 @@
 
 # Python version (python3 required).
 export PYTHON ?= python3
-# PIP_OPTIONS ?= # empty
-# Derived variable to allow filtering -e, or inserting other options
-# (the option --editable must always be last and only applies to src install)
-PIP_OPTIONS_E = $(filter-out -e,$(PIP_OPTIONS))
 # Set to 1 to skip all submodule updates. For development.
 NO_UPDATE ?= 0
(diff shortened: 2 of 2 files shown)


bertsky:
So, to illustrate, if you `make all` (with default options `DOCKER_PULL_POLICY=pull  DOCKER_VOL_MODELS=ocrd-models DOCKER_RUN_OPTS="-v $(DOCKER_VOL_MODELS):/usr/local/share/ocrd-resources -v $$PWD:/data -u $$UID"`), then this will `docker pull` all images and install a delegator shell script under `venv/bin/ocrd-...` for each executable, so for example `ocrd-tesserocr-recognize` will become:
```sh
#!/usr/bin/env bash
docker run --rm "${DOCKER_RUN_OPTS[@]}" -v ocrd-models:/usr/local/share/ocrd-resources -v $PWD:/data -u $UID ocrd/tesserocr ocrd-tesserocr-recognize "$@"
```

It will then proceed to build ocrd-all-tool.json and ocrd-all-meta.json from the checked out `ocrd-tool.json` of every submodule.

If passing `DOCKER_PULL_POLICY=build`, then in each checked out submodule, a respective `make docker` will be run to rebuild the images locally, instead of pulling them from Dockerhub.

To just pull (or build) the images, without (re-)installing the executable shell scripts, do `make images`.

Finally, to initialise the named volume `ocrd-models` from the pre-installed processor resources in the images and fix their permissions, just do `make init-vol-models` once.

To then manage processor resources (list or download), there are now additional delegator shell scripts for every image that just wrap the `ocrd` CLI, respectively. For example, to see what is installed for ocrd/tesserocr, do `ocrd-tesserocr-ocrd resmgr list-installed -e ocrd-tesserocr-recognize`. To install all registered models, do `ocrd-tesserocr-ocrd resmgr download ocrd-tesserocr-recognize "*"`. (That's exactly what `make install-models-tesseract` now does.)

(We cannot just use `ocrd resmgr` for this directly, as that only delegates to the `ocrd/core` image, which has no other processors installed, so it does not know about any resources.)

Using the processor CLIs is as simple as calling them by name, like in the native installation before, but now these will automatically start the respective container, mount the model volume, mount the current working directory into `/data` (the internal CWD) and run the processor in there.

So what is **not possible** anymore is using multi-processor tools like `ocrd process ...` (as `ocrd` just delegates to the `ocrd/core` image, which has no processors besides `ocrd-dummy` and `ocrd-filter`, and you cannot spin up containers from other containers), or `ocrd-make` (as that's just the `ocrd/workflow-configuration` image, which only has `ocrd-page-transform` installed internally). One could install core and workflow-configuration _natively_, so when `ocrd process ...` or `ocrd-make` start other processors, they get to use the delegator scripts...

